PR #2, opened 2022-05-18
feat: base contract setup 

thelostone-mc:
#### Description

This PR sets up a base contract via

```
├── contracts
|   ├── utils                             # useful utils
|   ├── GrantRoundFactory.sol             # factory contract which deploys grant round
|   ├── GrantRoundImplementation.sol      # grant round contract 
|   ├── vote                              # voting mechanism
|   |   ├── IVote.sol                     # voting interface
|   |   ├── BulkVote.sol                  # bulk voting mechanism
├── docs                                  # documentation 
├── scripts                               # deploy scripts
├── tests                                 # contract test
└── README.md
```

### Terminology

- **Round Operators**: wallets that have the permission to deploy & manage the round
- **Voter** : wallets who cast a vote to a grant during the grant round 

### General Setup

These steps would have to be done per chain but there are intended to be deployed only once

1. Deploying all voting mechanism (contracts under `votes/`)
2. Deploy GrantRoundImplementation
3. Deploy GrantRoundFactory with link to GrantRoundImplementation

### Round Setup

1. To create a round, you would not deploy a contract but instead, rely on the create function on the GrantRoundFactory to create a new GrantRoundImplementation
2. The user would have to choose a voting mechanism (already deployed via General Setup)
3. Any interaction in terms of updating parameters etc can be performed against the GrantRoundImplementation contract itself

The GrantRoundFactory enables us to have upgradable contracts on GrantRoundImplementation

### What's Pending :
- [x] GrantRoundFactory
- [x] IVote
- [x] BulkVote
- [x] GrantRoundImplmentation
  - [x] Basic Round implementation 
  - [x] Implement vote function on GrantRoundImplementation
- [ ] Grant Payout // need to figure out if this is it's own contract
- [ ] Add Tests

Note: Opening up this PR cause it compiles and has basic test check.
Will focus on implementing the rest as subsequent PRs


packages/contracts/contracts/utils/MetaPtr.sol
```diff
@@ -0,0 +1,12 @@
...
+
+struct MetaPtr {
+
+  /// @notice Protocol ID corresponding to a specific protocol.
+  /// More info at https://github.com/gitcoinco/grants-round/tree/main/packages/contracts/docs/MetaPtrProtocol.md
+  uint256 protocol;
+  
+  /// @notice Pointer to fetch metadata for the specified protocol
+  string pointer;
+}
```
gravityblast, line 12: @thelostone-mc I was thinking that we could use a URI instead of  MetaPtr, so that we don't need to define all the possible protocols and we can just set it to something like `ipfs://HASH` or `ceramic://did` etc /cc @kevinrolsen

thelostone-mc: Which means at the client-side we'd have to split by prefix and check right ?
It feels like an extra step on the client side. 
We still aren't defining all the protocols, it's just a keeping it as a struct as opposed to a URI which needs to be

gravityblast: @thelostone-mc cool it makes sense! they would be shared in between projects, so from grants-hub we could link to the same page.

PR #33, opened 2022-06-06
feat: switch Program Factory to use proxy 

thelostone-mc:
#### Description

- uses proxy pattern on Program to make an upgradeable contract
- introduces dummy upgradable contract with documentation on how to go about this 

This PR shows how we can rely on Proxy for program factory.
Once we focus on test coverage -> we can consider implementing the proxy pattern for Grant Factory as well 

Further details on what the proxy does can be found in the readme file.
For now **this does not change the behaviour on how program are created / deployed**

### DIFF
--- a/packages/contracts/scripts/program/deployProgramImplementation.ts
+++ b/packages/contracts/scripts/program/deployProgramImplementation.ts
@@ -15,10 +15,10 @@ export async function main() {
   });
 
   // Deploy ProgramImplementation 
-  const contractFactory = await ethers.getContractFactory("ProgramImplementation");
+  const contractFactory = await ethers.getContractFactory("DummyProgramImplementation");
   const contract = await contractFactory.deploy();
 
-  console.log(`Deploying ProgramImplementation to ${contract.address}....`);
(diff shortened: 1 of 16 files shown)


packages/contracts/scripts/program/deployProgramImplementation.ts
```diff
@@ -15,10 +15,10 @@ export async function main() {
   });
 
   // Deploy ProgramImplementation 
-  const contractFactory = await ethers.getContractFactory("ProgramImplementation");
+  const contractFactory = await ethers.getContractFactory("DummyProgramImplementation");
```
chibie, line 18: why was this renamed to `DummyProgramImplementation` ?

chibie (commented):
lgtm... just one question

packages/contracts/scripts/program/deployProgramImplementation.ts
thelostone-mc:
```suggestion
  const contractFactory = await ethers.getContractFactory("ProgramImplementation");
```

thelostone-mc:
Ah nice catch ! Forgot to update that back

PR #165, opened 2022-07-14
feat: use subgraph in program list service

### LINKED ISSUE #149: task: replace get program list and operators using subgraph
**Deployed to** 
https://thegraph.com/explorer/subgraph/thelostone-mc/program-factory-v0

**Subgraph endpoints:**
https://api.thegraph.com/subgraphs/name/thelostone-mc/program-factory-v0

**Queries I've documented here:**
https://github.com/gitcoinco/grants-round/tree/main/packages/graph/docs

note: when you run the query -> make sure sure the addresses are lower case

chibie:
##### Description

This PR uses the subgraph instead of contract interactions to get the list of programs for a given account

##### Refers/Fixes

closes #149

### DIFF
--- a/packages/round-manager/src/features/api/services/program.ts
+++ b/packages/round-manager/src/features/api/services/program.ts
@@ -91,8 +91,12 @@ export const programApi = api.injectEndpoints({
                     protocol
                     pointer
                   }
-                  accounts {
-                    address
+                  roles(where: {
+                    role: "0xaa630204f2780b6f080cc77cc0e9c0a5c21e92eb0c6771e709255dd27d6de132"
+                  }) {
+                    accounts {
+                      address
+                    }
                   }
                 }
               }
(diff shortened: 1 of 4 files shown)


packages/round-manager/src/features/api/services/program.ts
```diff
@@ -78,66 +78,42 @@ export const programApi = api.injectEndpoints({
         try {
           account = ethers.utils.getAddress(account)
 
-          const programFactory = new ethers.Contract(
-            programFactoryContract.address!,
-            programFactoryContract.abi,
-            global.web3Provider
+          const res = await graphql_fetch(
```
thelostone-mc, line 81: would it be worth pushing all the queries into it's own folders for the sake of re-usability? not sure how meaningful it would be

thelostone-mc (approved):
LGMT. Left a comment. Curious on if you see there are any benefits to this

packages/round-manager/src/features/api/services/program.ts
chibie: i thought about this but figured the queries wouldn't be reused elsewhere since components only interact with the services hence i only abstracted the `graphql_fetch` core functionality

Issue #157, opened 2022-07-12
Enable dApp to deploy to Optimism testnet


nfrgosselin:
**User Story:**
As a round operator
I want to deploy a new program and round on Optimism testnet
So that I can operate my round on the Optimism chain

**Acceptance Criteria**
GIVEN that I am in the program setup flow and my wallet is connected to Optimism testnet
WHEN I click to deploy my program
THEN my program contract deploys to the Optimism testnet chain (share link to contract on block explorer)

GIVEN that I am in the round setup flow and my wallet is connected to Optimism testnet
WHEN I click to deploy my round
THEN my round deploys to the Optimism testnet chain (share link to contract on block explorer)

**Product & Design Links:**

**Tech Details:**
- This is for the simple EMV implementation. We will add a spike for Metamask switching implementation separately. 

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

thelostone-mc:
Looking into this , 

- we would need to update https://github.com/gitcoinco/grants-round/blob/main/packages/round-manager/src/features/api/contracts.ts to take network as a parameter to be able to determine what contract to load 

- additionally it might be worth seeing the effort needed to be able to derive the network from metamask and use that to determine what contract to load.

PR #181, opened 2022-07-18
feat: deploy on optimism kovan 

thelostone-mc:
### Description

- update script command to support multiple networks
- add doc to state-supported networks
- update the documentation to reflect updated scripts
- add programContract to program.config.ts
- rename votingMechanism -> votingStrategy

### Refers/Fixes

closes #154 
closes #155 
closes #156 

### Testing

**Program**

- Deploying Upgradable ProgramFactory to [0xb19589C32351EC32652BAb386b61443b741B678a](https://kovan-optimistic.etherscan.io/address/0xb19589C32351EC32652BAb386b61443b741B678a)
- Deploying ProgramImplementation to [0x6038fd0D126CA1D0b2eA8897a06575100f7b16C2](https://kovan-optimistic.etherscan.io/address/0x6038fd0D126CA1D0b2eA8897a06575100f7b16C2)
- Program created:  [0x857e76d3E83f087c7BA18d03A8d7488fbE52DA1B](https://kovan-optimistic.etherscan.io/address/0x857e76d3E83f087c7BA18d03A8d7488fbE52DA1B)

image

image

====

**Round**

- Deploying Upgradable RoundFactory to [0xbB8f276FE1D52a38FbED8845bCefb9A23138Af92](https://kovan-optimistic.etherscan.io/address/0xbB8f276FE1D52a38FbED8845bCefb9A23138Af92)
- Deploying RoundImplementation to [0xc76Ea06e2BC6476178e40E2B40bf5C6Bf3c40EF6](https://kovan-optimistic.etherscan.io/address/0xc76Ea06e2BC6476178e40E2B40bf5C6Bf3c40EF6)
 - Round created:  [0x3Cd6edA7fDF9ab6b6AF6E226Ce184569C5DF8Ae5](https://kovan-optimistic.etherscan.io/address/0x3Cd6edA7fDF9ab6b6AF6E226Ce184569C5DF8Ae5)

Screenshot 2022-07-19 at 2 58 51 AM

image

PR #186, opened 2022-07-20
feat: load factory contract based on web3 network

thelostone-mc:
##### Description

- replace hardcoded factory contract with support for multiple network
- add networks -> 
  - optimism ( contract address to be set after deploy )
  - optimism kovan
  - goerli (default)

This PR is in draft state as it's broken and the API query is not tested/ working.  
This should be gotten in after #181 and rebasing 

##### Refers/Fixes

Closes https://github.com/gitcoinco/grants-round/issues/157

##### Testing

Deploy

### DIFF
--- a/packages/round-manager/src/features/api/utils.ts
+++ b/packages/round-manager/src/features/api/utils.ts
@@ -1,19 +1,44 @@
+import { useGetWeb3Query } from "./services/web3";
 import { IPFSObject } from "./types"
 
+const getGraphQLEndpoint = async () => {
+  // fetch chain id
+  const chainId = (await useGetWeb3Query().data)?.chainId
+
+  let endpoint;
+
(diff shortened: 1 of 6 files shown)


packages/round-manager/src/features/api/utils.ts
```diff
@@ -1,19 +1,44 @@
+import { useGetWeb3Query } from "./services/web3";
 import { IPFSObject } from "./types"
 
+const getGraphQLEndpoint = async () => {
+  // fetch chain id
+  const chainId = (await useGetWeb3Query().data)?.chainId
```
thelostone-mc, line 6: Is this right way to invoke this @chibie ?

chibie (approved):
nice one man :rocket:

PR #207, opened 2022-07-26
chore: update landing page

thelostone-mc:
##### Description

update landing page based on mock 

##### Refers/Fixes

closes #193 

##### Testing

image

PR #215, opened 2022-07-28
feat: add footer

thelostone-mc:
##### Description

Screenshot 2022-07-28 at 11 49 25 AM

##### Refers/Fixes

https://github.com/gitcoinco/grants-round/pull/207#pullrequestreview-1052653462


packages/round-manager/src/assets/landing/banner.svg
```diff
@@ -1,4 +1,4 @@
-<svg width="704" height="800" viewBox="0 0 704 800" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<svg width="704" height="700" viewBox="0 0 704 800" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
shavinac, line 1: why was the height changed here?

packages/round-manager/src/features/common/Footer.tsx
```diff
@@ -0,0 +1,49 @@
+import { ReactComponent as GitcoinCommunityLogo } from "../../assets/gitcoincommunity-logo.svg"
+
+
+const navigation = [
+  {
+    name: 'GitHub',
+    href: 'https://github.com/gitcoinco/grants-round',
+    icon: (props: any) => (
+      <svg width="21" height="20" viewBox="0 0 21 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
chibie, line 9: `props` isn't used here as in L22

chibie (changes requested):

packages/round-manager/src/assets/landing/banner.svg
thelostone-mc: The original image I had exported didn't have the footer while the newer one did! 
That's why it's been updated

packages/round-manager/src/features/common/Footer.tsx
thelostone-mc: Oh good catch

thelostone-mc:
```suggestion
    icon: (props: any) => (
      <svg width="21" height="20" viewBox="0 0 21 20" fill="none" {...props} xmlns="http://www.w3.org/2000/svg">
```

chibie (approved):

PR #238, opened 2022-08-03
feat: integrate lit sdk for decrypt / encrypt

### LINKED ISSUE #220: Integrate LIT SDK
**User Story:**
As a RM engineer
I want to integrate the LIT protocol SDK with Round Manager
So that I can generate public/private key pairs for PII encryption

**Acceptance Criteria**
- Integrate [LIT SDK](https://developer.litprotocol.com/)
- Configure LIT for the storage of public/private key pairs (includes configuring so that only round operators can decrypt for their round). The RM will generate the keyPair using https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle and then save the keyPair generated to Lit using an access control config (ex: address). NOTE: this should allow anyone access to the public key, but only round operators have access to private key.

**Product & Design Links:**

**Tech Details:**
The grant hub team has some examples / tests available here:
- https://gist.github.com/gravityblast/141d294e22a99459c202437f854b9009
- https://github.com/codenamejason/cryptography-lit-example

**Open Questions:**

**Notes/Assumptions:**

thelostone-mc:
##### Description

- creates a working implementation of encrypt / decrypting using lit
- added documenation on how we would encrypt / decrypt data
- locally tested

##### Refers/Fixes

closes #220 

##### Testing

Deploy


packages/round-manager/src/features/api/lit.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    // Encrypting Content and generating symmetric key
+    const { encryptedString, symmetricKey } = await LitJsSdk.encryptString(content);
+    console.log("Encrypted the key pair");
+    console.log("Encrypted String", encryptedString);
+    console.log("Symmetric Key", symmetricKey);
+
+    // Saving the Encrypted Content to the Lit Nodes
+    const encryptedSymmetricKey = await this.litNodeClient.saveEncryptionKey({
+      accessControlConditions: this.isRoundOperatorAccessControl(),
```
thelostone-mc, line 99: Shouldn't this evmContractConditions as well ?

packages/round-manager/docs/EncryptingPII.md
```diff
@@ -0,0 +1,53 @@
...
+- `encryptedSymmetricKey`
+
+Here is an example of the entire flow in action.
+
+```javascript
+
+// Init LIT config
+const lit = new Lit({
+  chain: "goerli",
+  contract: "0x22c0e3EDc90f6A890A259130B416Cd5F3Ee4Aca0"
```
shavinac, line 37: just to make sure i have this right - this address is a round contract, and LIT only decrypts for addresses who are round operators for that particular round?

shavinac (approved):

packages/round-manager/src/features/api/lit.ts
```diff
@@ -0,0 +1,149 @@
...
+    // Obtain Auth Signature to verify signer is wallet owner
+    const chain = this.chain;
+    const authSig = await LitJsSdk.checkAndSignAuthMessage({ chain });
+
+    // Encrypting Content and generating symmetric key
+    const { encryptedString, symmetricKey } = await LitJsSdk.encryptString(content);
+    // console.log("Encrypted the key pair");
+    // console.log("Encrypted String", encryptedString);
+    // console.log("Symmetric Key", symmetricKey);
+
```
chibie, line 94: can we remove console log comments like these?

packages/round-manager/src/features/api/lit.ts
```diff
@@ -0,0 +1,149 @@
+import { Buffer } from 'buffer';
+const LitJsSdk = require('lit-js-sdk');
```
chibie, line 2: why `require` and not `import`?

chibie (changes requested):
few comments + merge conflicts that need to be resolved

packages/round-manager/src/features/api/lit.ts
chibie: i know why now..

chibie (approved):

PR #239, opened 2022-08-03
fix: suppress warnings in tests and assert in scripts

### LINKED ISSUE #185: chore: should not emit warnings where unnecessary
**User Story:**
As a grant round engineer
I want to not see warnings in CI
So that I don't conflate them for actual issues

**Acceptance Criteria**
* we shouldn't throw warnings unless they need to be

**Tech Details:**
* we can resolve this by removing warnings from `hardhat.config.ts` and using them appropriately through modules in deploy scripts

**Notes/Assumptions:**
* for context, see https://github.com/gitcoinco/grants-round/discussions/169#discussioncomment-3156471

QEDK:
##### Description

Suppresses warnings on tests and only asserts variable existence during script runs.

##### Refers/Fixes

Fixes #185


thelostone-mc (approved):
Ooo i like this :D

PR #266, opened 2022-08-10
Adds encrypted flag to round metadata, only email is set to true for now

### LINKED ISSUE #221: Add encryption boolean to round metadata
**User Story:**
As a RM engineer
I want to pass encryption information in my application schema
So that Grant Hub knows which fields to encrypt

**Acceptance Criteria**
- ~Public key can be deployed with round schema~
- ~Encryption details are added to round application schema~
- Make it possible to add encrypt as a boolean to every question. default it to false
- Ensure when the application schema is updated , the documentation is updated as well 
- For MVP, email is the only field which should be encrypted 

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

LukeDowell:
Resolves #221 

Round contract: 
0x001b73d0c92f50e2a5385b1328d615d2addac438

Application Metadata:
https://gitcoin.mypinata.cloud/ipfs/bafkreifkt4sond4c5amadjffnoy6cebxbmflcid5trmxazf5os3zlpwi2a

### DIFF
--- a/packages/round-manager/src/features/api/utils.ts
+++ b/packages/round-manager/src/features/api/utils.ts
@@ -1,4 +1,4 @@
-import { IPFSObject, MetadataPointer } from "./types"
+import {ApplicationMetadata, IPFSObject, MetadataPointer} from "./types"
 
 export enum ChainId {
   GOERLI_CHAIN_ID= 5,
@@ -8,7 +8,7 @@ export enum ChainId {
 
 /**
  * Fetch subgraph network for provided web3 network
- * 
+ *
  * @param chainId - The chain ID of the blockchain2
  * @returns the subgraph endpoint
  */
@@ -85,7 +85,7 @@ export const fetchFromIPFS = (cid: string) => {
 
 /**
  * Check status of a grant application
- * 
+ *
  * @param id - the application id
  * @param projectsMetaPtr - the pointer to a decentralized storage
  */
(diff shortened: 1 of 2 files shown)


packages/round-manager/src/features/api/utils.ts
```diff
@@ -176,37 +176,35 @@ export const abbreviateAddress = (address: string) => `${address.slice(0, 8)}...
...
-      schema.push({
-        id: schema.length,
-        question: camelToTitle(key),
-        type: "TEXT",
-        required: true,
-        info: metadata[key],
-        choices: []
-      })
-    }
+export const generateApplicationSchema = (metadata: ApplicationMetadata) => {
```
thelostone-mc, line 179: Could we add the return type as an array ? Easier for doc generation

packages/round-manager/src/features/api/utils.ts
```diff
@@ -176,37 +176,35 @@ export const abbreviateAddress = (address: string) => `${address.slice(0, 8)}...
...
+  if (!metadata.customQuestions) return []
+
+  const maybe_questions = metadata.customQuestions
```
thelostone-mc, line 182:
```suggestion
  const customQuestions = metadata.customQuestions
```

let's keep the coding style consistent and keep it camelCase
Additionally let's avoid renaming variables. might cause confusion down the line

packages/round-manager/src/features/api/utils.ts
```diff
@@ -176,37 +176,35 @@ export const abbreviateAddress = (address: string) => `${address.slice(0, 8)}...
...
+
+  return questions.map(([name , answer], i) => {
+    return {
+      id: i,
+      question: camelToTitle(name),
+      type: "TEXT",
+      required: true,
+      info: answer,
+      choices: [],
+      encrypted: name === 'email'
```
thelostone-mc, line 207: worth adding a TODO comment that we would need to make remove this hardcoding and make it dynamic ?

thelostone-mc (dismissed):
@LukeDowell pointing to this https://discord.com/channels/562828676480237578/973967228577935400/1006868185116585984

Could we mark 
- which issue this fixes
- share the round contract where this applcation form was created (with the encrypted email)
we would need this info to share it to the grant-hub team as well

^ Ideally throw that in the PR description adding a bit more context

packages/round-manager/src/features/api/utils.ts
LukeDowell: Ha you are totally right, I have been doing python recently and it slipped in

packages/round-manager/src/features/api/utils.ts
LukeDowell: Personally, I don't much mind the static nature of the type checking here. I guess in this case it's all strings and wild-westy, but my preference would be to take the hit of having to update this when the input changes as a way of keeping ourselves honest about what we add to the metadata. 

Either way I'll add the TODO

chibie (changes requested):
@LukeDowell the round application schema doc will have to be updated to include the `encrypt` field

`packages/contracts/docs/ROUND_APPLICATION.md#L26`

shavinac (approved):

Issue #282, opened 2022-08-16
Round access control


nfrgosselin:
**User Story:**
As a
I want to
So that

**Acceptance Criteria**
GIVEN that I am on Round Manager AND my wallet is connected
WHEN I try to view the round details of a program to which I don't have access
THEN I see the No Access page

GIVEN that I am on Round Manager AND my wallet is connected
WHEN I try to view the application details of an application in a round to which I don't have access
THEN I see the No Access page

**Product & Design Links:**
[No Access page design](https://www.figma.com/file/K31Tf7tSn10BuN65VC6YLc/Round-Manager---Hi-Fidelity-Designs?node-id=939%3A5674)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

thelostone-mc:
Building it as part of changes in [NotFound](https://github.com/gitcoinco/grants-round/issues/243) rendering as this had to be taken into account while adding the conditional rendering

nfrgosselin:
This works for me on the round details page but doesn't do anything on the application details page. Trying to return to the round details page also gives me the 404 page (should be Access Denied)

https://share.vidyard.com/watch/y51pxT7WpmtToUVcvBzWpn?

nfrgosselin:
LGTM

Issue #281, opened 2022-08-16
Program access control


nfrgosselin:
**User Story:**
As a
I want to
So that

**Acceptance Criteria**
GIVEN that I am on Round Manager AND my wallet is connected
WHEN I try to view the program details of a program to which I don't have access
THEN I see the No Access page

**Product & Design Links:**
[No Access page design](https://www.figma.com/file/K31Tf7tSn10BuN65VC6YLc/Round-Manager---Hi-Fidelity-Designs?node-id=939%3A5674)

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

thelostone-mc:
Building it as part of changes in [NotFound](https://github.com/gitcoinco/grants-round/issues/243) rendering as this had to be taken into account while adding the conditional rendering

nfrgosselin:
This isn't working for me — I see the access denied screen but then it quickly flips to the 404 page:

https://share.vidyard.com/watch/NFdVQT1AMgGEP4xoaQa1GP?

thelostone-mc:
Hmm that's cause we fetch the programs by the users logged in wallet address. We should look to change that but this might mean updating the ProgramContext API we just rewrote. Will bring this up on standup

shavinac:
See this https://github.com/gitcoinco/grants-round/pull/297#discussion_r954334536 on the original 404/no-access pr

it should be straightforward to make the same change to ProgramContext `listProgram` api (`packages/round-manager/src/features/api/program.ts`) to allow program lookup by id without address, same as the query in `packages/round-manager/src/features/api/services/program.ts`

shavinac:
have some notes in [281--fix-program-access-control-context ](https://github.com/gitcoinco/grants-round/tree/281--fix-program-access-control-context) branch ([diff](https://github.com/gitcoinco/grants-round/compare/281--fix-program-access-control-context)) on how we could add this into Program Context -- it's a very rough draft so please reach out to me for info :)

nfrgosselin:
LGTM — thank you!

Issue #243, opened 2022-08-04
Handle invalid URL parameter errors


nfrgosselin:
**User Story:**
As a round operator
I want to be redirected to appropriate error pages
So that I know how to troubleshoot issues

**Acceptance Criteria**
- If program not found on program detail page -> redirect to 404
- If round not found on round detail page -> redirect to 404
- If application not found on application detail page -> redirect to 404

~~GIVEN that I am on Round Manager AND my wallet is connected~~
~~WHEN I try to access a page with an invalid parameter~~
~~THEN I see the 404 page~~

~~GIVEN that I am on Round Manager AND my wallet is not connected~~
~~WHEN I try to access a page~~
~~THEN I see the main RM page with connect wallet button~~

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

thelostone-mc:
Could you explain with an example here ? 
Me trying to understand what qualifies as a 404 vs access denied page in our app

thelostone-mc:
I believe this app already works this way:


https://rmgitcoin.on.fleek.co/#/program/0x51184b429edbcdc5ca5f60c69467a45d50e2c482 with connected wallet shows page.
https://rmgitcoin.on.fleek.co/#/program/0x51184b429edbcdc5ca5f60c69467a45d50e2c482 without a wallet connected redirects user to home page with connect wallet button

[Invalid routes --->  404 :](https://rmgitcoin.on.fleek.co/#/program/0x51184b429edbcdc5ca5f60c69467a45d50e2c482)
- https://rmgitcoin.on.fleek.co/#/beep
- https://rmgitcoin.on.fleek.co/#/random
- https://rmgitcoin.on.fleek.co/#/pop


![Image](https://user-images.githubusercontent.com/5358146/185544863-0a0254f0-6dc3-4182-b5eb-b7ade2faa711.gif)

thelostone-mc:
synced with @chibie ! 

@nfrgosselin this can be updated to say:

- If program not found on program detail page -> redirect to 404
- If round not found on round detail page -> redirect to 404
- If application not found on application detail page -> redirect to 404

nfrgosselin:
thanks @thelostone-mc / @chibie  — updated!

nfrgosselin:
LGTM

PR #297, opened 2022-08-22
chore: redirect to 404 / access denied for program/round

thelostone-mc:
##### Description

- add AccessDeniedPage
- set state to check if program / round exists
- set state to check if user has access to program
- set state to check if user has access to round

##### Refers/Fixes

fixes https://github.com/gitcoinco/grants-round/issues/243
fixes https://github.com/gitcoinco/grants-round/issues/281
fixes https://github.com/gitcoinco/grants-round/issues/282

##### Testing

Deploy


packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -29,6 +33,18 @@ export default function ViewProgram() {
     isSuccess: isRoundsFetched
   } = useListRoundsQuery({ address, signerOrProvider: provider, programId: id })
 
+  let [programExists, setProgramExists] = useState(true)
+  let [hasAccess, setHasAccess] = useState(true)
+
+  useEffect(() => {
+    if (isRoundsFetched) {
+      program ? setProgramExists(true) : setProgramExists(false)
```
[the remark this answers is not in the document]
thelostone-mc: Oooo this i didn't know :O

packages/round-manager/src/features/program/__tests__/ViewProgramPage.test.tsx
```diff
@@ -48,75 +67,44 @@ describe("<ViewProgram />", () => {
 
   describe("when there are no rounds in the program", () => {
     beforeEach(() => {
-      (useListRoundsQuery as jest.Mock).mockReturnValue({ data: [], isLoading: false, isSuccess: true });
+      (useListRoundsQuery as jest.Mock).mockReturnValue({ rounds: [], isLoading: false, isSuccess: true });
+      renderWrapped(<ViewProgram />);
```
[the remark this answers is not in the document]
thelostone-mc: Ooo didn't know this

packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -29,6 +33,18 @@ export default function ViewProgram() {
...
+
+      program?.operatorWallets.includes(address?.toLowerCase()) ? setHasAccess(true) : setHasAccess(false)
```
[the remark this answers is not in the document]
thelostone-mc: Yup ! Running the query against the graph returns lowercase instead the IBAN address 🙌🏼

packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -109,47 +125,53 @@ export default function ViewProgram() {
...
-                    <div className="flex flex-row justify-between">
-                      <p className="font-bold">My Rounds</p>
-                      <Link to={`/round/create?programId=${program?.id}`} className="text-violet-400 font-thin"
-                            data-testid="create-round-small-link">
-                        <PlusSmIcon className="h-5 w-5 inline -translate-y-0.5" aria-hidden="true" /> Create round
-                      </Link>
-                    </div>
-                    {roundItems}
+      {!programExists && <NotFoundPage />}
+      {!hasAccess && <AccessDenied />}
```
shavinac, line 129: I noticed that both "Not found" and "No access" pages will both appear on tall enough screens o.0 

i think this is because `hasAccess` is checking `program?.operatorWallets.includes(address?.toLowerCase())`, but this evaluates to a falsy value because of the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator (`program?.` == `undefined` if `program` is nullish)

Screen Shot 2022-08-24 at 4 29 34 PM

packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -29,6 +33,18 @@ export default function ViewProgram() {
...
   } = useListRoundsQuery({ address, signerOrProvider: provider, programId: id })
 
+  let [programExists, setProgramExists] = useState(true)
+  let [hasAccess, setHasAccess] = useState(true)
+
+  useEffect(() => {
+    if (isRoundsFetched) {
+      setProgramExists(!!program)
+
+      program?.operatorWallets.includes(address?.toLowerCase()) ? setHasAccess(true) : setHasAccess(false)
```
shavinac, line 43: i wasn't able to get an "Access Denied" to appear for the program page - I always got the 404 even though the program existed.

Steps to reproduce:
1. View a program detail page using Wallet 1
2. Switch to Wallet 2, where Wallet 2 is not one of the operators
3. See the 404 page, instead of just the Access Denied page as expected

Since the `GetPrograms`/`GetRounds` queries includes the address, would the subgraph ever return programs/rounds that _don't_ belong to the given address?

shavinac: actually, i am able to get the Access Denied page when I attempt to look at a round where I am not an operator - I noticed that the `ViewRoundPage` doesn't pass in the `address` into the `GetRounds` query, so it is able to retrieve any round by id

and the `ViewProgramPage` _does_ pass in `address` into the query, so any attempt to retrieve a program not matching the address acts the same as non-existing programs

thelostone-mc:
@shavinac I've tweaked the code to cover all the scenarios.

packages/round-manager/src/features/program/ViewProgramPage.tsx
thelostone-mc: Ah okie got what you are pointing out !
Fixed it up 🙌🏼  thanks for pointing this out

packages/round-manager/src/features/program/ViewProgramPage.tsx
thelostone-mc: Ugh that's ugly. Realized what's happening. Tweaked the condition to ensure only one of those flags are set to false

packages/round-manager/src/features/api/services/program.ts
```diff
@@ -109,11 +113,14 @@ export const programApi = api.injectEndpoints({
               }
             `,
             chainId,
-            { address: address.toLowerCase() }
+            { address: address?.toLowerCase(), programId }
           )
 
           const programs: Program[] = []
 
+          console.log(res);
```
shavinac, line 121: nitpick: remove console.log

shavinac: resolved in [5d0721b](https://github.com/gitcoinco/grants-round/pull/297/commits/5d0721b02753d0ce9f240ec00b496ce21d4fc56c)

shavinac (approved):

Issue #247, opened 2022-08-05
Approval/Reject decision not going through


nfrgosselin:
**Overview**

Fetch the application status from contract instead of relying on the graph
For more context refer this:
https://discord.com/channels/562828676480237578/1011646812127502427/1011955061401845780

This task would involve a few things:
- ensuring we log errors where IPFS call fails in graph
- ensure projectsMetaPtr is saved so that the frontend can use that to make a call to the contract and fetch the latest approval/rejection state

thelostone-mc:
This is not a bug and instead an expected situation of the subgraph indexing !
Ideally it shouldn't be too long once the txn succeeds. 
@nfrgosselin could you give this another try and record how long it took after the txn succeeded to it being reflected on the UI

shavinac:
This is happening, with >30 minute delay on the subgraph indexing 😬 

Example: I created a round [0xbb7240d08e2ebbf54bda6e7fa8e14b4ccb6665e2](https://rmgitcoin.on.fleek.co/#/round/0xbb7240d08e2ebbf54bda6e7fa8e14b4ccb6665e2), applied to the round, and rejected the application. ([transaction on goerli etherscan here](https://goerli.etherscan.io/tx/0x2b46fc5a050d45d27e99535e2ce25f79438a34fce0845766551d9a186cf4f4d6))

The subgraph `GetGrantApplications` query returns:
```
{
  "data": {
    "roundProjects": [
      {
        "id": "0x3339000000000000000000000000000000000000000000000000000000000000-0xbb7240d08e2ebbf54bda6e7fa8e14b4ccb6665e2",
        "metaPtr": {
          "protocol": 1,
          "pointer": "bafkreigu7i6gt35kfcnso22x2xbh5skjni5wacikpgz3ne6u6hcxckizxu"
        },
        "status": "PENDING",
        "round": {
          "projectsMetaPtr": {
            "protocol": 1,
            "pointer": "bafkreigfvyxtsk5wuxpduxgsgonfuc25tqvtn77thvbhgzmn5woxlkwt6i"
          }
        }
      }
    ]
  }
}
```
Note the `status` is still `PENDING`, but the `projectsMetaPtr` is up-to-date as `bafkreigfvyxtsk5wuxpduxgsgonfuc25tqvtn77thvbhgzmn5woxlkwt6i`.

[...]

PR #300, opened 2022-08-24
chore: improve handling of null metaPtr

thelostone-mc:
- add warning log when metaPtr is empty
- save round once metaPtr is updated

fixes https://github.com/gitcoinco/grants-round/issues/247


chibie (approved):
this looks good to me!

Issue #319, opened 2022-09-01
bug: fix CI logs


thelostone-mc:
> https://github.com/gitcoinco/grants-round/pull/301#issuecomment-1226886917
Looks like this PR throwing a lot of errors in console but the test pass. Could we fix it so that the tests don't reach those error state / explore a way of handling it  (not sure what would work here)

**Error 1**  (Fixed)
```
TypeError: Cannot read properties of undefined (reading 'operatorWallets')
```
This can adding operatorWallets to our mock data in the test, so should be simpler to fix

**Error 2**  (Fixed)
```
There are no focusable elements inside the <FocusTrap />
```

This log also turns up on the browser 

![Image](https://user-images.githubusercontent.com/5358146/187853266-fd1c27e0-7fd2-476e-ad59-a6879a04a096.png)

**Error 3** 

```
 FAIL  src/features/api/__tests__/program.test.ts
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'base16')

      2 | import { isJestRunning } from './utils';
      3 |
    > 4 | const LitJsSdk = isJestRunning() ? null : require('lit-js-sdk');
```

shavinac:
Error 1 and Error 3 fixed by PR #317

Error 2 fixed by https://github.com/gitcoinco/grants-round/pull/321

PR #321, opened 2022-09-01
chore: fix There are no focusable elements inside the FocusTrap

thelostone-mc:
##### Description

- fix There are no focusable elements inside the FocusTrap. Picked up from https://github.com/tailwindlabs/headlessui/issues/407#issuecomment-822753975

##### Refers/Fixes

refs: https://github.com/gitcoinco/grants-round/issues/319

##### Testing

Locally Tested

### DIFF
--- a/packages/round-manager/src/features/common/ProgressModal.tsx
+++ b/packages/round-manager/src/features/common/ProgressModal.tsx
@@ -163,6 +163,8 @@ export default function ProgressModal(
             </Transition.Child>
           </div>
         </div>
+        {/* Adding invisible button as modal needs to be displayed with a button */}
+        <button className="h-0 w-0 overflow-hidden"/>
       </Dialog>
     </Transition.Root>
   )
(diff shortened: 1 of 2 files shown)


packages/round-manager/src/features/common/ProgressModal.tsx
```diff
@@ -163,6 +163,8 @@ export default function ProgressModal(
             </Transition.Child>
           </div>
         </div>
+        {/* Adding invisible button as modal needs to be displayed without a button */}
```
chibie, line 166: shouldn't this read "needs to be displayed **with** a button"?

packages/round-manager/src/features/api/utils.ts
```diff
@@ -224,4 +224,4 @@ export const generateApplicationSchema = (metadata: any): Array<SchemaQuestion>
 }
 
 // Checks if tests are being run jest
-export const isJestRunning = () =>  process.env.JEST_WORKER_ID !== undefined;
```
shavinac, line 227: we found the lit-sdk issue! fixed in https://github.com/gitcoinco/grants-round/pull/317/commits/c45759aed555d007fb26bb949b895f54fde37f1d

PR #413, opened 2022-09-20
Merge bharghav/hmrtn/qv into quadratic-voting branch 

hmrtn:
##### Description

Initial PR for the quadratic-voting strategy module. 

##### Refers/Fixes

Adds quadratic voting strategy, a voter register, and a voting strategy interface.

##### Testing
Some tests are included, with more to be added in the future.


hmrtn:
This will be forced pushed as it is in a new quadratic-voting branch.

PR #442, opened 2022-09-30
feat(monorepo): use lerna for running tasks in parallel

### LINKED ISSUE #441: actually use lerna for tasks
as a developer
i want to run tasks using lerna
so that my tests and builds run in parallel, saving me time and making CI pipeline faster

dod: 
- all tasks are run using lerna, with correct scoping and parallelization

vacekj:
- use lerna for running top-level scripts
- update husky to use lerna
- update lerna version to v5
- use nx in lerna for faster task running

##### Refers/Fixes

fixes #441 

##### Testing

no change in tests

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-cd packages/round-manager
-yarn lint-staged
+yarn lerna run --scope=round-manager lint-staged
--- a/lerna.json
+++ b/lerna.json
@@ -2,8 +2,10 @@
   "packages": [
     "packages/contracts",
     "packages/graph",
-    "packages/round-manager"
+    "packages/round-manager",
+    "packages/grant-explorer"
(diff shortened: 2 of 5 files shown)


packages/round-manager/package.json
```diff
@@ -61,7 +61,8 @@
     "lint": "eslint ./src --cache",
     "lint:fix": "eslint ./src --fix --cache",
     "prepare": "cd ../.. && husky install",
-    "lint-staged": "lint-staged"
+    "lint-staged": "lint-staged",
+    "slither": "slither ."
```
shavinac, line 65: do we have any solidity files in `round-manager` package? if not, then we don't need the `slither` script here

thelostone-mc: That's a good callout

thelostone-mc:
```suggestion
```

thelostone-mc:
resolved the conflicts + addressed review comment

thelostone-mc (approved):

Issue #258, opened 2022-08-08
Create initial monitoring logs in Datadog


nfrgosselin:
**Acceptance Criteria**
- review error states in app 
- create monitoring settings for those states in Datadog

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

nfrgosselin:
deferring to Eng acceptance

PR #446, opened 2022-10-03
chore: improve datadog logging

thelostone-mc:
##### Description

Improve datadog logging 

##### Refers/Fixes

fixes: https://github.com/gitcoinco/grants-round/issues/258


shavinac (approved):

shavinac (commented):
should add datadog logs in 

* `packages/round-manager/src/features/api/program` catch blocks in each api
* `packages/round-manager/src/features/api/lit.ts` in decryptString

otherwise the original error gets swallowed in the catch blocks of those^, which returns / throws a different error

PR #452, opened 2022-10-03
feat(applicationmetadata): change question labels to be more human re…

dschinkel:
update application question labels to human-friendly copy

fixes: #430

Issue #430, opened 2022-09-27
update application question labels to human-friendly copy


shavinac:
Update the application schema generated during round creation to have human-friendly labels
Update round application form labels to stay consistent with the labels in the schema

* `Email` -> `Email Address`
* `Funding Source` -> `Funding Sources`
* `Profit2022` -> `2022 Profit`
* etc

see discord thread - https://discord.com/channels/562828676480237578/1022908453389344789

nfrgosselin:
LGTM

PR #453, opened 2022-10-03
fix(rm): use new application friendly labels when displaying answers

shavinac:
##### Description
 
use new application friendly labels when displaying answers on application detail page

#430 updated the application schema to use human-friendly labels for the questions. However, the application detail page was hard-coded to the old labels. This change updates the application detail page to read application answers using the new labels.

NOTE -- this is currently not backward-compatible -- applications submitted with the "old" labels will not show correct answers on the application page anymore

##### Refers/Fixes

fix bug introduced in PR #452 (original story: #430)


packages/round-manager/src/features/round/ViewApplicationPage.tsx
```diff
@@ -41,13 +42,6 @@ enum VerifiedCredentialState {
   PENDING,
 }
 
-enum ApplicationQuestions {
```
shavinac, line 44: question: we don't need to make RM "backward-compatible" with the old application question labels, right?

vacekj: afaik no we dont't

vacekj (approved):

PR #456, opened 2022-10-04
chore(githooks): make eslint fail on warnings to unify ci and local behaviour

### LINKED ISSUE #455: Make the lint fail on warnings in prepush hook to unify behaviour between CI and local
AS A developer
I want the linting results to be the same between CI and local dev env
so that I don't push code that passes checks locally but fails in CI

vacekj:
ignore the eslint unused variable pattern if the variable name is exactly `_`

##### Refers/Fixes

Fixes #455

### DIFF
--- a/packages/round-manager/.eslintrc.js
+++ b/packages/round-manager/.eslintrc.js
@@ -20,6 +20,14 @@ module.exports = {
       rules: {
         "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/no-non-null-assertion": "off",
+        "@typescript-eslint/no-unused-vars": [
+          "warn",
+          {
+            varsIgnorePattern: "_",
+            destructuredArrayIgnorePattern: "_",
+            argsIgnorePattern: "_",
+          },
+        ],
       },
     },
   ],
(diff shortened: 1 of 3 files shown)


packages/round-manager/.eslintrc.js
```diff
@@ -20,6 +20,14 @@ module.exports = {
       rules: {
         "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/no-non-null-assertion": "off",
+        "@typescript-eslint/no-unused-vars": [
+          "warn",
+          {
+            varsIgnorePattern: "_",
```
shavinac, line 26: `"_"` matches _any_ string containing `_` , so the rule will ignore all unused variable names containing `_` like `_thing`, `some_thing`, etc

`"^_$"` will match exactly `_` only

shavinac: resolved in [8237eaa](https://github.com/gitcoinco/grants-round/pull/456/commits/8237eaae9760d2d6da5f60c9b748f2c91fa1ab36)

shavinac (approved):

PR #464, opened 2022-10-06
fix: nest error modal inside progress modal to prevent scroll locking

### LINKED ISSUE #449: Page freeze during round creation
**Overview**
Occasionally, the second page of the round creation flow won't allow the user to scroll. They end up having to refresh and start over. 

**Observed  behavior**
Page doesn't let the user scroll

**Expected behavior**
User should be able to scroll

**Screenshots**
[Loom video here](https://www.loom.com/share/3665c5a7996f4d52a539075b5bdca14c)

**Additional context**

Steps to reproduce:
1. Go to a program page and click "Create Round"
2. Enter in past dates in the first form page (but still following chronological rules of the application/round start/end dates). This will cause an error during submission because the contract will revert.
3. Click "next" to go to the next form page
4. Click "submit" and see that the "Deployment" step fails.
5. See the error modal. Click "Done"
6. Click "Previous" to return to the previous form page
7. Click "Next" to go to the next page
8. See that the page prohibits scrolling. Also see that the error modal pops up after ~2 seconds

Alternative steps to reproduce:
1. Go to a program page and click "Create Round"
2. Enter valid dates in the form. Click "next" to go to the next form page
3. Click "submit".
4. Reject the wallet transaction when it appears. This will cause an error during submission
5. Follow steps 5-8 as above

Dev notes:
* following the steps above - there is an added `overflow: hidden` style in the `<html>` tag when the scrolling is prevented. this style is not present during normal usage except when the modals are open. This seems to be a side effect of the error modal - `overflow: hidden` is set after clicking "Done" on the error modal

shavinac:
##### Description

the dialog component from headlessui currently doesn't support simultaneous dialogs.
  even though the progress modal should close at the same time the error modal opens,
  they will exist simultaneously due to the transition animation time.
  the recommended fix is to either nest the dialogs,
  or delay the opening of the second dialog until the end of the first dialog's animation.

##### Refers/Fixes

fixes #449

### DIFF
--- a/packages/round-manager/src/features/common/ErrorModal.tsx
+++ b/packages/round-manager/src/features/common/ErrorModal.tsx
@@ -9,6 +9,7 @@ interface ErrorModalProps {
   heading?: string;
   subheading?: string;
   tryAgainFn?: () => void;
+  doneFn?: () => void;
 }
 
 export default function ErrorModal({
@@ -19,6 +20,7 @@ export default function ErrorModal({
   tryAgainFn = () => {
     /**/
   },
+  doneFn = () => setIsOpen(false),
 }: ErrorModalProps) {
   return (
     <Transition.Root show={isOpen} as={Fragment}>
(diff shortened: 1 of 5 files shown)


thelostone-mc:
Ah that's messy bug ! That said nesting it makes the most sense
Neat job on catching this <3

thelostone-mc (approved):

PR #494, opened 2022-10-11
remove rule overrides in .eslintrc.js

### LINKED ISSUE #411: Fix eslint warnings
# User Story:
As a developer
I want to disable ignoring of eslint warnings and fix the errors
So that our code is clean and functional

# Acceptance Criteria
- the following rule overrides are removed in `.eslintrc.js` root configuration:
```
```
- the command `CI=true yarn lint:fix` passes without warnings or errors
- specific overrides for certain file patterns (e.g. turn off `no-empty-function`, `no-non-null-assertion` rules for all test files) should be used if reasonable

# Notes
- empty functions that violate the `@typescript-eslint/no-empty-function` rule should have comments inserted in them explaining why they are empty / why they are necessary

bhargavaparoksham:
##### Description

1. The following rule overrides are removed in .eslintrc.js:
    - [x]  "@typescript-eslint/no-non-null-assertion": "off",
    - [x]  "@typescript-eslint/no-empty-function": "off", 
2. The command CI=true yarn lint:fix passes without warnings or errors

##### Refers/Fixes

fixes #411

### DIFF
--- a/packages/round-manager/.eslintrc.js
+++ b/packages/round-manager/.eslintrc.js
@@ -12,13 +12,11 @@ module.exports = {
   rules: {
     "react/no-unescaped-entities": ["error", { forbid: [">", "}"] }],
     "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/no-non-null-assertion": "off",
   },
   overrides: [
     {
       files: ["**/*.test.{ts,tsx}"],
       rules: {
-        "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/no-non-null-assertion": "off",
         "@typescript-eslint/no-unused-vars": [
           "warn",
--- a/packages/round-manager/.eslintrc.js
+++ b/packages/round-manager/.eslintrc.js
@@ -17,6 +17,8 @@ module.exports = {
     {
       files: ["**/*.test.{ts,tsx}"],
       rules: {
+        "@typescript-eslint/no-non-null-assertion": "off",
+        "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/no-unused-vars": [
           "warn",
           {
--- a/packages/round-manager/.eslintrc.js
+++ b/packages/round-manager/.eslintrc.js
@@ -17,7 +17,6 @@ module.exports = {
     {
       files: ["**/*.test.{ts,tsx}"],
       rules: {
-        "@typescript-eslint/no-non-null-assertion": "off",
         "@typescript-eslint/no-unused-vars": [
           "warn",
           {
--- a/packages/round-manager/src/features/common/__tests__/CopyToClipboardButton.test.tsx
+++ b/packages/round-manager/src/features/common/__tests__/CopyToClipboardButton.test.tsx
@@ -6,7 +6,7 @@ const textToCopy = "CLICK ME!";
 
 Object.assign(navigator, {
   clipboard: {
-    writeText: () => {},
+    writeText: () => true,
     readText: () => textToCopy,
   },
 });
--- a/packages/round-manager/src/features/common/__tests__/CopyToClipboardButton.test.tsx
+++ b/packages/round-manager/src/features/common/__tests__/CopyToClipboardButton.test.tsx
@@ -6,7 +6,9 @@ const textToCopy = "CLICK ME!";
 
 Object.assign(navigator, {
   clipboard: {
-    writeText: () => true,
+    writeText: () => {
+      /* do nothing.*/
+    },
     readText: () => textToCopy,
   },
 });
--- a/packages/round-manager/src/features/program/ViewProgramPage.tsx
+++ b/packages/round-manager/src/features/program/ViewProgramPage.tsx
@@ -63,7 +63,7 @@ export default function ViewProgram() {
           >
             <div className="flex-1 min-w-0">
               <p className="text-sm pb-3 mb-1 font-medium text-gray-900">
-                {round.roundMetadata.name}
+                {round.roundMetadata?.name ?? ''}
               </p>
 
               <div className="grid sm:grid-cols-3">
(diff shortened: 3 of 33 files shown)


packages/round-manager/.eslintrc.js
```diff
@@ -12,14 +12,11 @@ module.exports = {
   rules: {
     "react/no-unescaped-entities": ["error", { forbid: [">", "}"] }],
     "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/no-non-null-assertion": "off",
   },
   overrides: [
     {
       files: ["**/*.test.{ts,tsx}"],
       rules: {
-        "@typescript-eslint/no-empty-function": "off",
```
shavinac, line 21: i think we should leave the `no-empty-function: off` override for test files

bhargavaparoksham: Added back the overrides for test files.

thelostone-mc: Ah okie ! Makes sense to do that 🙌🏼

packages/round-manager/src/features/common/__tests__/CopyToClipboardButton.test.tsx
```diff
@@ -6,7 +6,9 @@ const textToCopy = "CLICK ME!";
 
 Object.assign(navigator, {
   clipboard: {
-    writeText: () => {},
+    writeText: () => {
+      /* do nothing.*/
```
thelostone-mc, line 10: Haha idk why but is much cleaner. <3

packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -63,7 +63,7 @@ export default function ViewProgram() {
           >
             <div className="flex-1 min-w-0">
               <p className="text-sm pb-3 mb-1 font-medium text-gray-900">
-                {round.roundMetadata!.name}
+                {round.roundMetadata?.name ?? ''}
```
thelostone-mc, line 66: Hmm i'm conflicted on this.
Ideally the metadata will always have a name so in that sense this shouldn't be optional right ? 

Does this error out if we we `{round.roundMetadata.name''}` @bhargavaparoksham  ?

cc @shavinac

packages/round-manager/src/features/round/ApplicationsApproved.tsx
```diff
@@ -34,7 +34,7 @@ export default function ApplicationsApproved() {
   const { id } = useParams();
   const { chain } = useWallet();
 
-  const { applications, isLoading } = useApplicationByRoundId(id!);
+  const { applications, isLoading } = useApplicationByRoundId(id ?? '');
```
thelostone-mc, line 37: Hmm curious if we it's smarter to update the useApplicationByRoundId to accept a nullable field / we update this to make sure this is not invoked if id is missing

packages/round-manager/src/features/program/ViewProgramPage.tsx
bhargavaparoksham: Yes, it is throwing an error, if we {round.roundMetadata.name''}

packages/round-manager/src/features/round/ApplicationsApproved.tsx
bhargavaparoksham: Updating useApplicationByRoundId to accept nullable field is causing other issues, as the id, is being passed as a parameter to another function within useApplicationByRoundId & the function accepts only a valid string ID.

I feel updating it to make sure it is not invoked isn't needed, as in the next step, only valid approved applications are being filtered out. 

Code link: https://github.com/gitcoinco/grants-round/blob/0e5bb58941789bdc04bb87e0346a5ea4f8ead9e2/packages/round-manager/src/features/round/ApplicationsApproved.tsx#L38

packages/round-manager/src/features/program/ViewProgramPage.tsx
shavinac: i think the `Round` type has `roundMetadata` set as an optional field, since the round contract could have a null/empty metadata pointer. But `name` is a required field on `roundMetadata`

in that case, we could add an initial filtering step to remove any rounds without metadata before generating the round list UI, like `rounds.filter(round => !!round.roundMetadata).map(...)`. then the ui should be able to show `<p>{round.roundMetadata.name}</p>` without using non-null assertion or causing TS compiler errors

packages/round-manager/src/features/round/ApplicationsApproved.tsx
shavinac: Generally you are not supposed to call React hooks conditionally -- you will get an ESLint warning
`React Hook "useApplicationByRoundId" is called conditionally. React Hooks must be called in the exact same order in every component render. (react-hooks/rules-of-hooks)`

Either we can change `useApplicationByRoundId` to accept optional `id`, and return some default values if `id` is not defined. Or disable the non-null assertion rule for this line, with the explanation that the `id` URL parameter should never be null

thelostone-mc: > we disable the non-null assertion rule for this line, with the explanation that the id URL parameter should never be null

I would lean towards this

packages/round-manager/src/features/program/ViewProgramPage.tsx
thelostone-mc: > rounds.filter(round => !!round.roundMetadata).map(...)

Yup i like this 🙌🏼

thelostone-mc:
Let's avoid using `?? : ""` and instead either add a skip assertion rule / run a filter.

Reason for this is : if the type states a parameter is required -> then passing an empty string can cause a bug to trickle down and break things.

thelostone-mc (approved):

shavinac (approved):

PR #532, opened 2022-10-17
feat: graph implementation for voting

### LINKED ISSUE #366: Define subgraph entity schema for QFVoting
**User Story:**

Given that a voting contract is now linked the round. 
The subgraph should be updated to fetch information about the votes made against a contract 

The goal here would be to 
- define an Entity Schema to capture the contributions recorded in the voting contract
- handlers which would be able index the votes recorded against the subgraph 
- deploy the subgraph on test network successfully  and fetch the votes

**Acceptance Criteria**
- User can query subgraph and receive votes

**Dev notes:**
WIP draft PRs for reference: 
- https://github.com/gitcoinco/grants-round/pull/532 
- https://github.com/gitcoinco/grants-round/pull/556

Related spike
- https://github.com/gitcoinco/grants-round/issues/367

thelostone-mc:
##### Description

- add VotingStrategy entity to schema
- add VotingStrategy abi
- update config to include QFVotingStrategFactory address
- update subgraph template to include QFVotingStrategy
- add handleContractCreated to index QF contract
- add handleVote to index QF vote
- add tests for handlers
- update handleRoundCreated to link Round entity to votingStrategy entity

##### Refers/Fixes

closes #366
 
##### Testing

https://thegraph.com/hosted-service/subgraph/thelostone-mc/round-labs

```
{
  votingStrategies(where:{
    id:"0x1a497d28890efb320d04f534fa6318b6a0657619"
  }) {
    id
    votes {
      id
      amount
      token
      from
      to
    }
  }
}
```


packages/graph/tests/votingStrategy/quadraticFunding/implementation.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  newVoteEvent.parameters.push(tokenParam);
+  newVoteEvent.parameters.push(amountParam);
+  newVoteEvent.parameters.push(voterParam);
+  newVoteEvent.parameters.push(grantAddressParam);
+  newVoteEvent.parameters.push(roundAddressParam);
+
+
+  // TODO: how to set event contract address as votingStrategy contract
+  // We need to be able to set event.receipt.logs[0].address as the VotingStrategy
+  // Only then can the handler set the QFVote.votingStrategy value
```
thelostone-mc, line 42: This gets tricky cause ideally  we wanna be able to set the contract address deploying the event manually

thelostone-mc:
Pending: 
- Emit vote events and check if the subgraph indexes them right

thelostone-mc:
@vacekj 3 things which we need to do :

- Update our actual subgraph 
- See if we can link QFVote to Round (this doesn't happen right now which is why i removed it on the last commit)
- Wire in a github workflow to run tests on this package

packages/graph/package.json
```diff
@@ -26,7 +26,6 @@
     "test": "graph test"
   },
   "dependencies": {
-    "@graphprotocol/graph-cli": "0.30.4",
```
thelostone-mc, line 29: removing this cause it breaks on m1

More info here: https://discord.com/channels/438038660412342282/903628101819461632/1009174614192181342

packages/graph/src/votingStrategy/quadraticFunding/implementation.ts
```diff
@@ -1,42 +1,52 @@
...
 
-  // load Round contract
-  let round = Round.load(votingStrategy.round!);
-  if (!round) {
-    log.warning("--> handleVotingContractCreated {} : round is null", [
-      "QF",
-    ]);
-    return;
-  }
+  // if (!votingStrategy.round) {
```
thelostone-mc, line 24: This is something we have to figure out how to implement

packages/graph/schema.graphql
```diff
@@ -85,4 +86,33 @@ type RoundProject @entity {
...
+
+# id = keccak256(event.receipt.transactionHash.toHex(), QFVote.to)
+type QFVote @entity {
+  id: ID!
+  votingStrategy: VotingStrategy!
+  token: String!
+  amount: BigInt!
+  from: String!
+  to: String!
+  # round: Round!
```
thelostone-mc, line 117: I would love to integrate this but it looks like since VotingStrategy.Round is not set when the data is indexed , this fails when you try setting it via votingStrategy.round

Here is code which was removed https://github.com/gitcoinco/grants-round/pull/532/commits/a87a2b147e5cf85f1a7e6b09791a07fc77aaf5f3#r1015412397

nfrgosselin (commented):
Defer to engineering approval

melissa-neira:
Defer to engineering approval 🤠

packages/graph/schema.graphql
hmrtn: This may be due to the ordering 

> The following process is used to order the triggers for a data source within a block:
> 
> - The transaction index within the block orders event and call triggers
> - Should there be event and call triggers within the same transaction, the following convention is used for ordering: event triggers first then call triggers. Each type respects the particular order they are defined in the manifest.
> - After event and call triggers, block triggers are run in the order as defined in the manifest.

packages/graph/src/votingStrategy/quadraticFunding/implementation.ts
hmrtn: See comment above. I think we can get this merged and write out a separate story for the issue here.

hmrtn (approved):
This looks fine to merge and passes tests. There are a couple of outstanding issues noted here that are not necessarily immediate blockers: 
-  See https://github.com/gitcoinco/grants-round/pull/532/files#r1018354431

nfrgosselin:
Thanks for the review @hmrtn! I just created a shell follow-on story #653 

@thelostone-mc / @vacekj can you guys populate that shell story?

I'm going to merge this in the meantime

PR #544, opened 2022-10-19
feat: cleaner date picker UX

### LINKED ISSUE #450: Cleaner date picker UX
**User Story:**
As a round operator
I want the UI to provide clearer feedback around valid round dates
So that I don't have unnecessary / unclear errors

**Acceptance Criteria**
GIVEN that I am on the first step of the round creation page and have clicked on a date field
WHEN the date picker opens
THEN past dates are grayed out
AND I cannot select past dates

GIVEN that I am on the first step of the round creation page
WHEN I select today's date
THEN the time defaults to the top of the next hour (e.g. if it is 10:45, then the time will set at 11:00)

GIVEN that I am on the first step of the round creation page and have entered a date+time prior to the current time
WHEN I click the Next button
THEN I see a validation error under the field with the incorrect date that says "You must enter a date and time in the future."

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

thelostone-mc:
##### Description

- prevent users from picking past dates
- add validation error if application start date is in the past

##### Refers/Fixes

fixes #450

### DIFF
--- a/packages/round-manager/src/features/common/styles.ts
+++ b/packages/round-manager/src/features/common/styles.ts
@@ -1,3 +1,4 @@
+import { FieldError } from "react-hook-form";
 import tw from "tailwind-styled-components";
 
 type ButtonProps = {
@@ -22,7 +23,7 @@ export const Button = tw.button`
 `;
 
 type InputProps = {
-  $hasError?: boolean;
+  $hasError?: boolean|FieldError;
   $disabled?: boolean;
 };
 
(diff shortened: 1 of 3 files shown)


shavinac (approved):

melissa-neira (approved):

nfrgosselin (changes requested):
Overall it looks good, but we're missing the requirement to set the time to a point later than now if you select today as a date

shavinac:
> Overall it looks good, but we're missing the requirement to set the time to a point later than now if you select today as a date

Addressed in latest push to this PR 👍

nfrgosselin (approved):

Issue #534, opened 2022-10-17
Editable application toggle states


nfrgosselin:
**User Story:**
As a round operator
I want to configure my application form
So that I can customize the information that I receive from grantees

**Acceptance Criteria**
GIVEN that I am on the Review Information step of the round creation flow
WHEN I click a toggle to ON
THEN the toggle state (e.g. encrypted or required) is set to ON for that question

GIVEN that I am on the Review Information step of the round creation flow
WHEN I click a toggle to OFF
THEN the toggle state (e.g. encrypted or required) is set to OFF for that question

GIVEN that I have Application Questions for my round
WHEN I click to Launch the round
THEN the toggle states are recorded in the question metadata for that round

**Product & Design Links:**
[Figma](https://www.figma.com/file/K31Tf7tSn10BuN65VC6YLc/Round-Manager---Hi-Fidelity-Designs?node-id=1916%3A15042)

**Tech Details:**
- depends on #501

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

Issue #501, opened 2022-10-11
Add toggles to application questions


nfrgosselin:
**SYNC WITH NATE + MELISSA BEFORE STARTING THIS STORY**

**User Story:**
As a round operator
I want to configure my application form
So that I can customize the information that I receive from grantees

**Acceptance Criteria**
GIVEN that I am on the Review Information step of the round creation flow
WHEN I look at the Application Information section
THEN I see an "Encrypted" toggle and a "Required" toggle next to every question

GIVEN that I am on the Review Information step of the round creation flow
WHEN I hover over the tool tip for "Encrypted"
THEN I see the tool tip helper copy

GIVEN that I am on the Review Information step of the round creation flow
WHEN I look at the Application Information section
THEN both toggles next to "Payout Wallet Address" default to grayed out / uneditable
AND both toggles next to "Email Address" default to "ON"
AND all other toggles default to "OFF"

**Product & Design Links:**
[Figma](https://www.figma.com/file/K31Tf7tSn10BuN65VC6YLc/Round-Manager---Hi-Fidelity-Designs?node-id=1916%3A15042)

**Tech Details:**
- depends on https://github.com/gitcoinco/grants-round/issues/451

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

PR #545, opened 2022-10-19
feat: (rm)  toggle application question

thelostone-mc:
##### Description

- add encrypted toggle to editable questions
- add required toggle to editable questions
- set payout wallet address in disable state
- add empty cross icon which does nothing
- update .gitignore to exclude .vscode

##### Refers/Fixes

closes https://github.com/gitcoinco/grants-round/issues/501
closes https://github.com/gitcoinco/grants-round/issues/534

### DIFF
--- a/packages/round-manager/src/features/round/RoundApplicationForm.tsx
+++ b/packages/round-manager/src/features/round/RoundApplicationForm.tsx
@@ -1,5 +1,7 @@
 import { useContext, useEffect, useState } from "react";
 import {
+  Control,
+  Controller,
   FieldArrayWithId,
   SubmitHandler,
   useFieldArray,
@@ -24,8 +26,21 @@ import ErrorModal from "../common/ErrorModal";
 import { errorModalDelayMs } from "../../constants";
 import { useCreateRound } from "../../context/round/CreateRoundContext";
 import { datadogLogs } from "@datadog/browser-logs";
-import { CheckIcon, PencilIcon } from "@heroicons/react/solid";
+import {
+  CheckIcon,
+  PencilIcon,
+  XIcon,
+  InformationCircleIcon,
+} from "@heroicons/react/solid";
+import { Switch } from "@headlessui/react";
+import ReactTooltip from "react-tooltip";
 
+const payoutQuestion: QuestionOptions = {
+  title: "Payout Wallet Address",
+  required: true,
+  encrypted: false,
+  inputType: "text",
+};
 export const initialQuestions: QuestionOptions[] = [
   {
     title: "Email Address",
@@ -223,6 +238,7 @@ export function RoundApplicationForm(props: {
               editStates={isInEditState}
               setEditStates={setIsInEditState}
               getValues={getValues}
+              control={control}
             />
 
             <div className="px-6 align-middle py-3.5 shadow-md">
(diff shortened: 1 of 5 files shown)


packages/round-manager/src/features/round/RoundApplicationForm.tsx
```diff
@@ -342,6 +362,42 @@ function ApplicationInformation(props: {
...
+          {...register(`applicationMetadata.questions.${index}.encrypted`)}
+          checked={encryptedStates[index]}
+          onChange={() => {
+
+            const newEncryptedState = [...encryptedStates];
+            newEncryptedState[index] = !encryptedStates[index];
+            setEncryptedStates(newEncryptedState);
+
+            // THIS IS BROKEN
+            // WHERE TO REGISTER
```
thelostone-mc, line 383: @shavinac this seems to fail and my hunch is cause the registering doesn't happen right

shavinac: i think this is fixed by using a [form controller](https://react-hook-form.com/api/usecontroller/controller) instead of local state to control the switch value. and also manually setting `checked` [prop on the Switch](https://headlessui.com/react/switch#switch)
```
<Switch
    {...field}
    checked={field.value}
    value={field.value.toString()}
    className={classNames(
      field.value ? "bg-indigo-600" : "bg-gray-200",
      "relative inline-flex h-6 w-11 flex-shrink-0 cursor-pointer rounded-full border-2 border-transparent transition-colors duration-200 ease-in-out focus:outline-none focus:ring-2 focus:ring-indigo-500 focus:ring-offset-2"
    )}
    data-testid="encrypted-toggle"
>
```

had to set the `value` prop to override the `value` set by the `{...field}`. otherwise the TS compiler complains that the `field.value` (boolean) and Switch's `value` (string/number/undefined) don't match 😒 😩

packages/round-manager/src/features/round/RoundApplicationForm.tsx
```diff
@@ -22,9 +25,17 @@ import ErrorModal from "../common/ErrorModal";
 import { errorModalDelayMs } from "../../constants";
 import { useCreateRound } from "../../context/round/CreateRoundContext";
 import { datadogLogs } from "@datadog/browser-logs";
-import { CheckIcon, PencilIcon } from "@heroicons/react/solid";
+import { CheckIcon, PencilIcon, XIcon, InformationCircleIcon } from "@heroicons/react/solid";
+import { Switch } from "@headlessui/react";
+import ReactTooltip from 'react-tooltip';
 
 export const initialQuestions: QuestionOptions[] = [
+  {
```
shavinac, line 33: wondering if we should leave out the Payout Wallet Address question here?

We can create a "hardcoded" component to display on the UI, since it doesn't need to be interactive
Like this: https://github.com/gitcoinco/grants-round/compare/478--toggle-application-question...shavinac/501-toggles 

Also I think we should sync with Grant Hub team about whether or not we should include the Payout Address question in the application schema. Currently they are hard-coding the question in the Grant Hub form -- so if we add the payout question on the round application schema, then there will be duplicate Payout Address questions on Grant Hub

melissa-neira (commented):
Small update - Encrypted toggles: should say "unencrypted" when inactive.
Example:
Screen Shot 2022-10-20 at 16 00 15
Also, the required toggle should say "optional" when inactive. see example above^
unless there's another story addressing this already?

shavinac:
> Small update - Encrypted toggles: should say "unencrypted" when inactive. Example: Screen Shot 2022-10-20 at 16 00 15 Also, the required toggle should say "optional" when inactive. see example above^ unless there's another story addressing this already?

We can make those changes in this PR!

shavinac (approved):

melissa-neira (approved):

nfrgosselin (approved):

PR #574, opened 2022-10-27
feat(graph): create a local hardhat + graph-node docker compose

### LINKED ISSUE #234: chore: add in ability to run subgraph locally
## Description 

The goal here is to setup subgraph locally thus making it easier to run a local node and test your queries without having to actual deploy to a hosted service maintained by the graph

vacekj:
@thelostone-mc @nfrgosselin I need some additional clarification on #234. What I added in this PR is a docker compose, than when run, will spin up a local hardhat node, which you can deploy contracts against, then spin up a graph-node against anvil. Then you should be able to deploy contracts as usual, interact with them and the subgraph should pick everything up. I haven't yet set-up the Anvil node as a hardhat network, since I first want to get confirmation that this is the right way to go about this.

EDIT: replace anvil node with hardhat node

##### Refers/Fixes

fixes #234

### DIFF
--- a/packages/graph/README.md
+++ b/packages/graph/README.md
@@ -50,7 +50,19 @@ To know more about the queries which can be run on the playground, check out the
 To know the relationship between the different entities and the type of queries. Refer [schema.graphql](./schema.graphql)
 
 
-## Deploy Subgraph
+## Deploy subgraph locally
+
+The `docker-compose.yml` contains everything you need to run your own local graph-node, including a hardhat node.
+
+1. Install Docker with Docker Compose
+2. `docker compose up` - start all the services
+3. `yarn deploy-local` -  deploy your subgraph to your local graph-node instance
+3. Deploy your contracts as described in `contracts/docs/DEPLOY_STEPS.md`
+4. Interact with the contracts to create some data
+5. Interact with the subgraph on localhost as outlined in the `graph-node` container logs
+
+
+## Deploy Subgraph on a hosted service
 Generate your hosted-service API key on the graph
 
 - Remove redundant files
@@ -72,6 +84,7 @@ yarn prepare:<NETWORK_TO_DEPLOY_SUBGRAPH>
 | optimism       |
 | fantom         |
 | fantom-testnet |
+| localhost      |
 
 
 - Run codegen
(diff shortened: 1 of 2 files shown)


thelostone-mc:
Questions: 

- can we not use the hardhat node as opposed to relying on anvil ?
- additionally to deploy the contracts on local chain -> I'm assuming we'd need to write up a script for that as well right ?

melissa-neira:
Defer to engineering 🕺

thelostone-mc:
Could we add in documentation on the graph/README on how to run this ?

nfrgosselin (commented):
Defer to engineering review

hmrtn:
Docker compose looks good and runs just fine with frequent block updates. 

It appears to be using hardhat rather than anvil - is this correct?

hmrtn (approved):

nfrgosselin (approved):
defer to engineering approval

packages/graph/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+        "postgres",
+        "-cshared_preload_libraries=pg_stat_statements"
+      ]
+    environment:
+      POSTGRES_USER: graph-node
+      POSTGRES_PASSWORD: let-me-in
+      POSTGRES_DB: graph-node
+      # FIXME: remove this env. var. which we shouldn't need. Introduced by
+      # <https://github.com/graphprotocol/graph-node/pull/3511>, maybe as a
+      # workaround for https://github.com/docker/for-mac/issues/6270?
```
thelostone-mc, line 54: ^ Is this needed ?

thelostone-mc (changes requested):
could we update the documentation to state how to run this  + update the PR message to state we are using hardhat node and not anvil

packages/graph/docker-compose.yml
vacekj: Removed

vacekj:
> Docker compose looks good and runs just fine with frequent block updates.
> 
> It appears to be using hardhat rather than anvil - is this correct?

That is correct, updated PR text to reflect this change.

thelostone-mc (approved):

PR #588, opened 2022-10-28
chore: add nate and melissa to CODEOWNERS. remove shavina and dave

shavinac:
As per Oct 27 Retro action item: add Melissa + Nate to default PR reviewer list, so they will automatically be tagged as reviewers for all PRs

Also removing myself and Dave since today is our last day with RM/GE team 😭


thelostone-mc (approved):
You guys will be missed <3

Issue #586, opened 2022-10-28
Add projects to final donation


nfrgosselin:
**User Story:**
As a donor
I want to add shortlist projects to my final donation
So that I can submit my donations

**Acceptance Criteria**
GIVEN that I am in the selecting state and have >=1 projects selected
WHEN I click the "Add Selected..." button
THEN all selected projects are moved to my final ballot
AND those selected projects are removed from my shortlist
AND the "Add Selected..." button reverts to the "Select" button state and the gray unselected state

**Product & Design Links:**
[Figma](https://www.figma.com/file/lZwHrKYiWJna3AePzyTAiG/Grant-Explorer?node-id=1583%3A63811)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

Issue #488, opened 2022-10-08
Remove from final donation


nfrgosselin:
**User Story:**
As a donor
I want to be able to remove projects from my final donation
So that I can narrow down my final choices

**Acceptance Criteria**
GIVEN that I have projects on my final ballot
WHEN I click the back arrow icon
THEN the project is removed from my final donation
AND the project is added back to the shortlist

**Product & Design Links:**
[Figma](https://www.figma.com/file/lZwHrKYiWJna3AePzyTAiG/Grant-Explorer?node-id=1583%3A63811)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

vacekj:
@thelostone-mc just to confirm, this is already fixed by #645 right?

thelostone-mc:
Yup that is right! We just needed to wire it to the UI

PR #645, opened 2022-11-09
feat: update BallotContext with finalBallot implementation

thelostone-mc:
##### Description

- implement addToFinalBallot to remove from shortlist before adding to finalBallot
- implement removeFromFinalBallot to add to shortlist after removing from shortlist
- update localstorage to include loadFinalBallot and saveFinalBallot

##### Refers/Fixes

refs #586
refs #488

### DIFF
--- a/packages/grant-explorer/src/context/BallotContext.tsx
+++ b/packages/grant-explorer/src/context/BallotContext.tsx
@@ -7,7 +7,7 @@ import React, {
   useEffect,
   useState,
 } from "react";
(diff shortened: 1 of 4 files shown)


chibie (approved):
this looks good to me!

thelostone-mc:
> Am curious why it is entirely necessary to use a TestingUseBallotComponent component to proceed with the tests. Is this due to the context limitations?

I assume so! Let's bring this up on tech retro and see if we can avoid doing it -> benefits of that

PR #724, opened 2022-11-29
feat: CRUD operation for api package

### LINKED ISSUE #383: votingsStrategy: return matching of all projects in a round
**User Story:**

Once these steps are implemented, there should be a URL for us to fetch the latest data

- Add a new endpoint for fetch which takes in input of roundID and chain ID 
- When endpoint is invoked -> we query the DB to fetch the latest information and return the JSON
- If DB has no entry -> we invoke calculate to persist DB and return results

**Product & Design Links:**
- [Flow in Miro](https://miro.com/app/board/uXjVPY3Csls=/?moveToWidget=3458764532978275196&cot=14)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

### LINKED ISSUE #715: feat votingStrategy : persist result of calculate onto DB
Now that we have a calculate function within linear QF.
The goal here would be to add a util function called `saveToDB` which ideally just saves the data for the round.

AKA each time the calculate function is invoked -> we run the numbers and then update the DB

### LINKED ISSUE #717: votingStrategy: return matching of a project in a round
**User Story:**

Once these steps are implemented, there should be a URL for us to fetch the latest data

- Add a new endpoint for fetch which takes in input of roundID and chain ID and project ID
- When endpoint is invoked -> we query the DB to fetch the latest information and return the matching amount of the project ID
- If DB has no entry -> we invoke calculate to persist DB and return results

thelostone-mc:
##### Description

- wire in prisma ORM
- define DB schema 
- `calculateHandler` does upsert into Round and PayoutTable
- update documentation
-  add new endpoint `fetch-matching` to fetch all payouts of a given round and a project's payout of a given round 

##### Refers/Fixes

closes #715
closes #383 
closes #717

##### Testing

- http://localhost:8000/fetch-matching?roundId=0xcef1772dd6764c95f14c26b25e8f012c072c5f77&projectId=0x5cdb35fadb8262a3f88863254c870c2e6a848cca

- http://localhost:8000/fetch-matching?roundId=0xcef1772dd6764c95f14c26b25e8f012c072c5f77


packages/api/src/handlers.ts
```diff
@@ -0,0 +1,107 @@
...
+        results = await linearQFCalculate(metadata, votes);
+        break;
+    }
+
+    console.log(results);
+
+    if (results && results.distribution.length > 0) {
+      // save the distribution results to the db
+      // TODO: figure out a way to do bulk insert
+      results.distribution.forEach(async (match) => {
```
hmrtn, line 81: This should be a standard for loop instead of forEach, as forEach expects a synchronous function.

thelostone-mc: doneee

hmrtn:
Running `docker-compose up` will now start the API and database containers instead of having to run the server separately.

PR #725, opened 2022-11-29
chore: use grantId instead of applicationId

thelostone-mc:
- update RoundImplementation applyToRound to use uint256 grantId
- update vote method in QuadraticFundingVotingStrategyImplementation to emit grantId


thelostone-mc:
Close as this it not applicable ATM

PR #727, opened 2022-11-30
Docker node server

thelostone-mc:
##### Description

- [x] local + prod docker setup
- [x] prisma schema
- [x] prisma docs
- [x] restructure folders
- [x] insert payouts in db
- [ ] setup env variables
- [ ] update prod dockerfile with prisma config
- [ ] Pass express PORT env to dockerfile


packages/api/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+version: '3.8'
+services:
+  # web:
```
hmrtn, line 3: This is commented out but should build and start the express web API. 
Currently fails on M1 mac because of a binary target error.

hmrtn:
To get started with testing the api and database locally: 

Build and run the docker image. It will create a database.
```bash
docker-compose up
```

Then generate the prisma client, migrate the schema, and run the express app locally.
```bash
npm run generate
npm run migrate
npm run dev
```

A post request to the `/calculate` like:
```json
{
  "chainId": "3",
  "roundId": "0xcef1772dd6764c95f14c26b25e8f012c072c5f77"
}
```
will upload the rounds project voting data to the database, and can be viewed easily with the `all/` endpoint. 

Notice - you will need to configure a new `env` var, the database URL. Set it to this for now:
```
PORT=8000
DATABASE_URL="postgresql://root:root@localhost:5432/grants-round?schema=public"
```

What needs to be done: 
- [ ] Fix the docker compose to build correctly, that way we can simply run `docker-compose up` and have the API image start as well.
- [ ] Resolve the relationship and round ID issue in the schema (see todo)
- [ ] Implement updating the database instead of simply creating a new record, when possible. 
- [ ] Fix prisma document generation and serving

melissa-neira:
Defer to engineers 🫡

hmrtn (approved):

PR #746, opened 2022-12-06
feat: denominate matching contribution as the projects token

### LINKED ISSUE #738: feat: (api) add convert function which convert tokens contributions to round token
Add new the `calculate` handler function to include a util function 

- accepts the output of linearQFFetchVotes and round token address as the input
- the util function would then loop through all the contributions are convert the contribution amount in same token address as the round (this is done using the function written in https://github.com/gitcoinco/grants-round/pull/734)
- the output from that is passed to the linearQFCalculate function and that ensure all calculations are done with the contributions in the same denomination

hmrtn:
fixes #738 

Todo: 
- [ ] Figure out testnet cases, as coingecko does not support testnets


hmrtn:
The token from the response of running calculateHandler needs to match the token it is converted to. If it fails - it needs to match the input token. This is so we can find the discrepancy later, if necessary.

thelostone-mc (approved):

melissa-neira:
defer to engineering 😎

PR #743, opened 2022-12-05
chore(api): wip docker compose improvements

vacekj:
add basic prisma seed
adjust env variables

PR #761, opened 2022-12-07
feat: docker fixes

chibie:
##### Description

This PR introduces the following:

* sync machine folders with container for reload on file changes
* fix out of memory issue


thelostone-mc (approved):
locally this seems to work for me ! 
Defer to @vacekj to see if we need to make more changes w.r.t to https://github.com/gitcoinco/grants-round/pull/743

PR #785, opened 2022-12-09
update calculation service

### LINKED ISSUE #760: api: changes to the calculate endpoint
### Description 

- [ ] ~Update API response to return conversion rate~ 
- [x] Update output of fetchVotes to include both the original and coverted token amount
- [ ] Ensure when calculations are run -> if conversion rate isn't applicable -> the api returns an error instead of returning incorrect calculation
- [ ] Figure out testnet cases, as coingecko does not support testnets

chibie:
##### Description

* update output of fetchVotes to include both the original and converted token amount
* properly handle errors + push them up to be return in response json
* fix backward compatibility issue with voting strategy name

##### Refers/Fixes

closes #760


packages/api/src/handlers.ts
```diff
@@ -48,7 +48,13 @@ export const calculateHandler = async (req: Request, res: Response) => {
...
     const metadata = await fetchRoundMetadata(req.body.chainId, roundId);
 
-    const { id: votingStrategyId, strategyName } = metadata.votingStrategy;
+    let { id: votingStrategyId, strategyName } = metadata.votingStrategy;
+
+    // for backward compatibility with older subgraph version
+    strategyName =
+      strategyName === "quadraticFunding"
+        ? VotingStrategy.LINEAR_QUADRATIC_FUNDING
+        : strategyName;
```
thelostone-mc, line 57: ^ we can remove this ! We've redeployed the subgraph on all networks

packages/api/src/votingStrategies/linearQuadraticFunding.ts
```diff
@@ -32,35 +31,66 @@ export const fetchVotesHandler = async (
...
+  }
+
+  contributions = await Promise.all(
+    votes.map(async (vote: any) => {
+      // denominate the contribution in the round token
+      // TODO: look into this logic. When should conversion happen?
+      // TODO: Determine how we want to pass any errors. If the conversion is unavailable, then the converted amount will
+      //       be the same as the original amount. Is this an error and should it be reported to the user?
+      //       The DenominationResponse type is defined in types.ts and includes a message/success field for future use.
+
```
thelostone-mc, line 64: ^ could we create follw up story on this as well so that we can get them added into our backlog

melissa-neira:
Defer to engineering review 🤠

packages/api/src/handlers.ts
thelostone-mc: Good point ! Nice catch here

thelostone-mc (approved):

PR #806, opened 2022-12-09
fix: mobile view on round page

thelostone-mc:
##### Description

Just did a quick overview of the round page https://grant-explorer.gitcoin.co/#/round/5/0x0739c6bb86297d847db22cc56388d3abfda551ff on mobile and threw in a few tweaks


melissa-neira (changes requested):
Everything seems good except the footer is now a bit wonky. let's align the 2 icons to the right and centered with the gitcoin community logo.
This is is what it looks like on my end:
Screen Shot 2022-12-09 at 20 17 35

thelostone-mc:
@melissa-neira fixed that up as well

melissa-neira (approved):

PR #856, opened 2022-12-17
feat: re-implement match calculate handler

hmrtn:
This reimplements the calculate handler and adds a new endpoint
`/match/:chainId/:roundId`
ex `GET http://localhost:8000/match/250/0xebdb4156203c8b35b7a7c6f320786b98e5ac67c3`

```
{
    "success": true,
    "message": "match calculations",
    "data": {
        "distribution": [
            {
                "projectId": "0xd27e1a1a60ebc1b70f4cae5265092c0f6edc7f9d",
                "matchAmountInUSD": 6632.864972367244,
                "totalContributionsInUSD": 19081.678575961618,
                "matchPoolPercentage": 0.10876045276567153,
                "matchAmountInToken": 5.438022638283576
            },
            {
                "projectId": "0x39d77e51c485f1ff65b1b3b42e9f67cda221f597",
                "matchAmountInUSD": 5763.189422331367,
                "totalContributionsInUSD": 11665.960964348564,
                "matchPoolPercentage": 0.09450020369152538,
                "matchAmountInToken": 4.725010184576269
            },
            {
                "projectId": "0x8ddfa84bbf59c0e6dbd05a56e9b9931b64255f08",
                "matchAmountInUSD": 4142.798495581125,
                "totalContributionsInUSD": 18486.442540658776,
                "matchPoolPercentage": 0.06793032000100228,
                "matchAmountInToken": 3.3965160000501142
            },
                  ...
        ],
        "isSaturated": true,
        "updatedAt": "2022-12-20T00:22:32.844Z"
    }
}
```

PR #883, opened 2022-12-21
Calc cache n handlers

hmrtn:
read the readme for info on the endpoints

we need tests for the handlers

should be good place to work off of


hmrtn:
This will be merged without approval as it extends pr #856

PR #884, opened 2022-12-21
adding ftmTestnet apiKey

drumfeet:


thelostone-mc:
Could you add more context on why this is needed and what about the opera entry which is already uses the same env file.
Also an example of the contract verified to show this works as intended using the changes made would be needed

PR #920, opened 2022-12-29
fix: docker setup

hmrtn:
this fixes the issue with docker directly firing up and connecting to the database. 

it adds an ephemeral migration container to handle the prisma migration

running `docker-compose up` should work now - initializing a database and starting the API.


packages/api/package.json
```diff
@@ -44,6 +43,7 @@
     "express": "^4.17.1",
     "node-cache": "^5.1.2",
     "node-fetch": "^2.6.6",
+    "prisma-docs-generator": "^0.5.0",
```
thelostone-mc, line 46: ^ this should be a dev dependency

hmrtn: Strangely enough - leaving the doc generator in the dev dependency breaks the initialization of the docker containers. This is because the generator is used at compile time from the schema.prisma file. 
Not sure if there is a way to get around this.

thelostone-mc:
Looks like this works as expected ! So closing this out

PR #937, opened 2023-01-04
chore: summary endpoint fix

thelostone-mc:
##### Description

It looks like the endpoint are wired in wrong.
I'm not clear if all of them should be POST but sticking to the whatever is written and just updating the URL path for now


hmrtn:
The endpoint was set to `update` because if the data does not exist, the response will be empty. This was a temporary solution until we start regularly updating the database with a corn job, etc.

hmrtn (approved):

PR #959, opened 2023-01-09
927 add open passport buttons

thelostone-mc:
##### Description

- add open passport / create passport
- added dummy button for recalculate score

##### Refers/Fixes

closes #927

PR #958, opened 2023-01-08
925 add passport verification page

thelostone-mc:
##### Description

- add Passport connect page
- wire in connect buttons on PassportBanner

##### Refers/Fixes

closes #925

PR #962, opened 2023-01-09
feat: add PassportConnect

thelostone-mc:
##### Description

- chore: add open passport buttons
- chore: add fetch score logic
- chore: increase test coverage
- removes undefined behavior in passport state
- chore: score to fixed 2 decimal
- chore: add recalculate score trigger
- add zero score check

##### Refers/Fixes

closes https://github.com/gitcoinco/grants-round/issues/962
closes https://github.com/gitcoinco/grants-round/issues/958
closes https://github.com/gitcoinco/grants-round/issues/959

### DIFF
--- a/packages/grant-explorer/package.json
+++ b/packages/grant-explorer/package.json
@@ -61,6 +61,7 @@
   "scripts": {
     "start": "craco start",
     "build": "craco build",
+    "test-log": "craco test --watchAll=false",
     "test": "craco test --watchAll=false --silent",
     "test:coverage": "craco test --watchAll=false --silent --coverage",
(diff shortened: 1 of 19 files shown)


hmrtn:
We need a way of fetching a users community ID before continuing with the recalculate button. Currently, and incorrectly, this is achieved via a env var.

thelostone-mc:
> We need a way of fetching a user or round community ID before continuing with the recalculate button. Currently, and incorrectly, this is achieved via a env var.

Yup we'll have to fix this when passport gives us the means to create an community via the API. Then we'd be able to create a community and use that

PR #963, opened 2023-01-09
Tagmanager url fragment fix

giliomeejg:
##### Description

react-gtm4 isn't going to work, as it bypasses tagmanager and works directly with Google Analytics.  We're trying to avoid this so that the tagmanager admin has more flexibility to tweak tags and triggers without programmatic intervention.


packages/grant-explorer/src/tagmanager.ts
```diff
@@ -1,7 +1,13 @@
-import ReactGA from "react-ga4";
+import TagManager from "react-gtm-module";
```
thelostone-mc, line 2: Don't we need to reinstall package

vacekj (approved):

Issue #977, opened 2023-01-12
Passport verification page updates


melissa-neira:
**Acceptance Criteria**
The page looks pretty much the same as before, but without tabs. Copy has been updated as well.  [Figma](https://www.figma.com/file/lZwHrKYiWJna3AePzyTAiG/Grant-Explorer?node-id=2951%3A79342&t=hbYikfSTcCn84J6z-4)
- No more tabs
- Updated copy
- Added link for support ("need help" button on the bottom) - I don't have a link for this yet, I should be receiving something soon from the support team.

thelostone-mc:
Review on https://github.com/gitcoinco/grants-round/pull/982

nfrgosselin:
Accepting this on behalf of @melissa-neira so we can get this in for the production release

Issue #974, opened 2023-01-12
feat: update the recalculate score flow for new API 


thelostone-mc:
### Description

Read this section on how to make these changes:
https://discord.com/channels/562828676480237578/1042785636333146204/1063022368840232981

Passport has updated how their submit score endpoint works and it's now an async process!
We would now have to ensure that we keeping fetching the score until the status is DONE. 
Additionally the threshold is also provided by the get score endpoint

thelostone-mc:
Review on https://github.com/gitcoinco/grants-round/pull/982

nfrgosselin:
This looks good to me — moving to staging

PR #980, opened 2023-01-13
chore: update PassportConnect

thelostone-mc:
##### Description

This PR is being merged to passport-integration-branch

- update design of passport connect page
- update calculate button 

refs #977 
refs #974

### DIFF
--- a/packages/grant-explorer/src/features/common/PassportBanner.tsx
+++ b/packages/grant-explorer/src/features/common/PassportBanner.tsx
@@ -43,6 +43,12 @@ export default function PassportBanner(props: {chainId?: string, roundId?: strin
         const res = await fetchPassport(address, PASSPORT_COMMUNITY_ID);
         if (res.ok) {
           const json = await res.json();
+
+          if (json.status == "PROCESSING") {
+            await callFetchPassport();
+            return;
+          }
+
           setPassport(json);
           setPassportState(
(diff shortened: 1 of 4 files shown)


melissa-neira (changes requested):
Hey, this doesn't include the latest copy changes made. Please update :)

melissa-neira:
Also, something weird is happening to my Passport score. I'm not sure if this is related to this PR or not.
https://www.loom.com/share/70656ca576b242388a6ca5f5d6381ef3

thelostone-mc:
@melissa-neira could you hit recalculate score

packages/grant-explorer/src/features/common/PassportBanner.tsx
```diff
@@ -43,6 +43,12 @@ export default function PassportBanner(props: {chainId?: string, roundId?: strin
         const res = await fetchPassport(address, PASSPORT_COMMUNITY_ID);
         if (res.ok) {
           const json = await res.json();
+
+          if (json.status == "PROCESSING") {
+            await callFetchPassport();
```
vacekj, line 48: won't this result in a possibly infinite loop? If it never changes from processing. Also, we should probably have exponential backoff here. I'll merge anyway, but I'll also add a TODO here to address this.

PR #994, opened 2023-01-17
fix: query first 1000

KurtMerbeth:
due to graphs limit of 100 results, not all projects are shown in round manager. to fix this, the limit is set to 1000

### DIFF
--- a/packages/round-manager/src/features/api/application.ts
+++ b/packages/round-manager/src/features/api/application.ts
@@ -119,7 +119,8 @@ export const getApplicationsByRoundId = async (
         // (status ? `status: $status` : ``)
         // +
         `
-          }) {
+          }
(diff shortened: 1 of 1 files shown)


DanieleSalatti:
Note that this is a hotfix to try and unblock the round, the ideal solution is to use pagination :)

chibie (approved):

hmrtn (approved):

hmrtn:
NOTE: Add pagination similar to [this](https://github.com/gitcoinco/grants-round/blob/main/packages/api/src/votingStrategies/linearQuadraticFunding.ts#L173) (in the future). 

Let's get a ticket out for this. 1000 results should be sufficient - for now. 

Thanks @KurtMerbeth

PR #1010, opened 2023-01-18
bug: ensure first contributions usd value is populated

thelostone-mc:
##### Description

It looks like in the current implementation the usdValue is not set for the very first contribution made by the contributor.
This causes that value to be ignored and results in negative matching for projects with 2 contributions 

Additionally my hunch is that when a project has 1 contribution , it would be ignored and the percentage would be 0

### DIFF
--- a/packages/api/src/votingStrategies/linearQuadraticFunding.ts
+++ b/packages/api/src/votingStrategies/linearQuadraticFunding.ts
@@ -418,24 +418,37 @@ export const matchQFContributions = async (
   for (const contribution of contributions) {
     const { projectId, amount, token, contributor } = contribution;
 
-    if (!contributionTokens.includes(token)) {
-      contributionTokens.push(token);
-    }
+    const usdAmount = Number(formatUnits(amount)) * prices[token];
 
+    // check if projectID is already in the mapping
     if (!contributionsByProject[projectId]) {
+
+      // add projectID to mapping along with the contribution
       contributionsByProject[projectId] = {
-        contributions: contribution ? { [contributor]: contribution } : {},
+        contributions: {        // all contributions made to the projectId
+          [contributor]: {      // all contributions made by contributor to the projectId
+            ...contribution,    // list of all contributions made by contributor to the projectId
+            usdValue: usdAmount // total USD amount for all contributions made by contributor to the projectId
+          }
+        },
       };
     }
 
+    // check if contributor has already made contributions to the project
     if (!contributionsByProject[projectId].contributions[contributor]) {
-      contributionsByProject[projectId].contributions[contributor] = {
-        ...contribution,
-        usdValue: Number(formatUnits(amount)) * prices[token],
-      };
+
+      // append contributor to the projectId mapping
+      contributionsByProject[projectId]
+        .contributions[contributor] = {
+          ...contribution,
+          usdValue: usdAmount,
+        };
     } else {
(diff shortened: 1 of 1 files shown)


packages/api/src/votingStrategies/linearQuadraticFunding.ts
```diff
@@ -418,24 +418,37 @@ export const matchQFContributions = async (
   for (const contribution of contributions) {
     const { projectId, amount, token, contributor } = contribution;
 
-    if (!contributionTokens.includes(token)) {
```
thelostone-mc, line 421: This seems redundant as this is already done in L404. 
This is unreachable code from what I can tell

packages/api/src/votingStrategies/linearQuadraticFunding.ts
```diff
@@ -418,24 +418,37 @@ export const matchQFContributions = async (
...
-      contributionTokens.push(token);
-    }
+    const usdAmount = Number(formatUnits(amount)) * prices[token];
```
thelostone-mc, line 421: Extracting usdAmount into a variable making the code cleaner

packages/api/src/votingStrategies/linearQuadraticFunding.ts
```diff
@@ -418,24 +418,37 @@ export const matchQFContributions = async (
...
+    // check if projectID is already in the mapping
     if (!contributionsByProject[projectId]) {
+
+      // add projectID to mapping along with the contribution
       contributionsByProject[projectId] = {
-        contributions: contribution ? { [contributor]: contribution } : {},
+        contributions: {        // all contributions made to the projectId
+          [contributor]: {      // all contributions made by contributor to the projectId
+            ...contribution,    // list of all contributions made by contributor to the projectId
+            usdValue: usdAmount // total USD amount for all contributions made by contributor to the projectId
```
thelostone-mc, line 431: this was not initially set which is why I log the results -> the usdValue was NaN

packages/api/src/votingStrategies/linearQuadraticFunding.ts
```diff
@@ -418,24 +418,37 @@ export const matchQFContributions = async (
...
-        ...contribution,
-        usdValue: Number(formatUnits(amount)) * prices[token],
-      };
+
+      // append contributor to the projectId mapping
+      contributionsByProject[projectId]
+        .contributions[contributor] = {
+          ...contribution,
+          usdValue: usdAmount,
+        };
```
thelostone-mc, line 445: ^ no change here, just made the code easier to read

packages/api/src/votingStrategies/linearQuadraticFunding.ts
```diff
@@ -418,24 +418,37 @@ export const matchQFContributions = async (
...
     } else {
-      contributionsByProject[projectId].contributions[contributor].usdValue +=
-        Number(formatUnits(amount)) * prices[token];
+
+      // update total USD amount as this contributor has already made contributions to the project
+      contributionsByProject[projectId]
+        .contributions[contributor] // all contributions made by contributor to the projectId
+        .usdValue += usdAmount; // total USD amount for all contributions made by contributor to the projectId
```
thelostone-mc, line 451: ^ no change here, just made the code easier to read

hmrtn (approved):

PR #1018, opened 2023-01-19
feat: wire in passport

### LINKED ISSUE #990: Pull Passport scores into matching calculation
**User Story:**
As a round operator
I want donor's Passport scores to be included in their matching calculation
So that my round is protected from Sybil

**Acceptance Criteria**

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

thelostone-mc:
##### Description

Looks like passport bulk API is quite slow. 
Fetching all passport score for a given community requires a GET call which limits results to 1000.
and each call takes about 3 seconds and currently there are 8000 stamps (and this will increase over time)

We will need to figure out a more scalable solution

fixes #990

### DIFF
--- a/packages/api/.env.sample
+++ b/packages/api/.env.sample
@@ -21,3 +21,6 @@ SENTRY_DSN=####
 
 # Etherscan API key
 OPTIMISM_ETHERSCAN_API=####
+
+# Passport API key
+PASSPORT_API_KEY=####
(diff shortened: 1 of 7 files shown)


thelostone-mc:
Tests need to be fixed ! Feature works otherwise

PR #1029, opened 2023-01-20
fix: ensure api endpoints are not case sensitive

### LINKED ISSUE #975: bug: ensure api endpoint are not case sensitive
#### Description

Currently the api expects all lowercase for round Id and project ID in the request
We need to update the API package to ensure all parameters are made lowercase to ensure endpoint works when address is sent in right format 

https://discord.com/channels/562828676480237578/973967228577935400/1062384932120436807

bhargavaparoksham:
##### Description

Currently the api expects all lowercase for round Id and project ID in the request
We need to update the API package to ensure all parameters are made lowercase to ensure endpoint works when address is sent in right format

##### Refers/Fixes

fixes #975

### DIFF
--- a/packages/api/src/handlers/getProjectMatchDataHandler.ts
+++ b/packages/api/src/handlers/getProjectMatchDataHandler.ts
@@ -8,7 +8,7 @@ export const getProjectMatchDataHandler = async (
   req: Request,
   res: Response
 ) => {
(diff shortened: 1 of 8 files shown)


melissa-neira:
defer to engineers 🕺

nfrgosselin:
Defer to engineers

thelostone-mc (approved):

PR #1032, opened 2023-01-22
fix(navbar, 404): broken links to homepage

llllvvuu:
##### Description

Neither `"#"` nor `"/"` point to the grants round homepage. You can see this bug by going to https://grant-explorer.gitcoin.co/#/round/1/404 and observing the infinite loop, or trying to click on the nav bar to go home. It has been changed to ``/round/${chainId}/${roundId}`` which is used elsewhere in the app and is correct.

##### Refers/Fixes

N/A

##### Testing

N/A - change too minor

### DIFF
--- a/packages/grant-explorer/src/features/common/Navbar.tsx
+++ b/packages/grant-explorer/src/features/common/Navbar.tsx
@@ -1,4 +1,4 @@
-import { Link } from "react-router-dom";
+import { Link, useParams } from "react-router-dom";
 import { ReactComponent as GitcoinLogo } from "../../assets/gitcoinlogo-black.svg";
 import { ReactComponent as GrantsExplorerLogo } from "../../assets/topbar-logos-black.svg";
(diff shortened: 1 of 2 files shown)


llllvvuu:
Actually, don't merge this: when `baseUrl` is different, `"/"` is not correct either. That means the other instance where we already have `"/"` (`NotFoundPage.tsx:29`) is also incorrect and you can see this by going to https://grant-explorer.gitcoin.co/#/round/1/404 which is an infinite loop currently.

thelostone-mc (changes requested):
This might need further discussion as the navbar and NotFoundPage would link to the explorer page which is yet to be built.
There are instances where you might land on a page which has no round and that point we won't have a round / chain to link it to  (like the NotFound page)

@nfrgosselin this is valid point which folks have been facing -> Do we have a default landing page ?

PR #1053, opened 2023-01-26
fix: upsert issue

### LINKED ISSUE #1015: bug: Upsert issue
### Description

Currently in api package, the Match table has projectid set to be unique 
This is incorrect has projects can apply to different rounds and hence there will be duplicate entries.

Ideally the fix here is we have a joint unique key of 
- projectId
- roundId

Cause for every round , there can be only 1 entry for a given project in the match table

A/C: 
- update schema to make roundId + projectID unique 
- update schema to remove only projectId as unique
- ensure all endpoints work
An easy way to test this is to ensure matching endpoints work for all alpha rounds as you will have projects applying to all those rounds. Ensure data for each round is uploaded into the DB after each run

hmrtn:
simplifies db upserts
adds unique identifier (of project and round) to all relevant entries

fixes #1015


melissa-neira:
defer to engineering 🫡

thelostone-mc (approved):

thelostone-mc (changes requested):
Seems to break on my local when I try running the endpoints
Can you validate and see if the API work as expected ?

`/Users/thelostone-mc/Labs/grants-round/packages/api/bin/src/database.js:97:41`

image

PR #1058, opened 2023-01-27
feat: wire in passport

thelostone-mc:
**Description**

Looks like passport bulk API is quite slow.
Fetching all passport score for a given community requires a GET call which limits results to 1000.
and each call takes about 3 seconds and currently there are 8000 stamps (and this will increase over time)

We will need to figure out a more scalable solution

fixes https://github.com/gitcoinco/grants-round/issues/990


melissa-neira (approved):

hmrtn (approved):

nfrgosselin (commented):
Defer to engineering review

PR #1064, opened 2023-01-31
fix: api summary bug 

hmrtn:
this fixes an unlisted bug on the summary update handler. The issue was summary data was returning 0 on the USD, contributors, etc. 

It also has a quality-of-life refactor for the graph fetches.

### DIFF
--- a/packages/api/src/utils.ts
+++ b/packages/api/src/utils.ts
@@ -631,7 +631,7 @@ export const fetchProjectIdToPayoutAddressMapping = async (
   for (const project of projects) {
     // project.id format ->  applicationId-roundId
     const projectId = project.id.split("-")[0];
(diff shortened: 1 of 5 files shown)


packages/api/src/utils.ts
```diff
@@ -631,7 +631,7 @@ export const fetchProjectIdToPayoutAddressMapping = async (
   for (const project of projects) {
     // project.id format ->  applicationId-roundId
     const projectId = project.id.split("-")[0];
-    const payoutAddress = getAddress(project.payoutAddress);
+    const payoutAddress = getAddress(project.payoutAddress).toLowerCase();
```
thelostone-mc, line 634: Oh nice catch

thelostone-mc (approved):

PR #1071, opened 2023-02-01
fix: resolve round summary upsert issue

### LINKED ISSUE #1070: bug: round summary upsert issue
**Overview**

This is an issue that happens when using the update summary for a round endpoint. If an item already existed in a table, it updating it would violate our key indexes. 

Resolve it by investigating the prisma and database implementation for update round summary handler.

hmrtn:
fixes #1070

### DIFF
--- a/packages/api/src/database.ts
+++ b/packages/api/src/database.ts
@@ -169,15 +169,17 @@ export class DatabaseInstance {
       const roundSummaryData = {
         contributionCount: summary.contributionCount,
         uniqueContributors: summary.uniqueContributors,
-        totalContributionsInUSD: Number(summary.totalContributionsInUSD),
-        averageUSDContribution: Number(summary.averageUSDContribution),
+        totalContributionsInUSD: Number(summary.totalContributionsInUSD) ? Number(summary.totalContributionsInUSD) : 0,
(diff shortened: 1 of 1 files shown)


packages/api/src/database.ts
```diff
@@ -169,15 +169,17 @@ export class DatabaseInstance {
       const roundSummaryData = {
         contributionCount: summary.contributionCount,
         uniqueContributors: summary.uniqueContributors,
-        totalContributionsInUSD: Number(summary.totalContributionsInUSD),
-        averageUSDContribution: Number(summary.averageUSDContribution),
+        totalContributionsInUSD: Number(summary.totalContributionsInUSD) ? Number(summary.totalContributionsInUSD) : 0,
+        averageUSDContribution: Number(summary.averageUSDContribution) ? Number(summary.averageUSDContribution) : 0,
```
thelostone-mc, line 173: ^ Why is this needed ? 
The default value for summary.totalContributionsInUSD and summary.averageUSDContribution is 0 
Why would we have anything else here ?

melissa-neira:
defer to engineering 🦋

packages/api/src/database.ts
hmrtn: I added this and forgot to remove it as a sanity check.

vacekj (approved):

PR #1085, opened 2023-02-04
feat: add vote table, database seeding (WIP)

hmrtn:
This PR modifies the schema of our database to include a Votes table. 

The table looks like : 
```
  id               Int     
  createdAt        DateTime 
  updatedAt        DateTime 
  roundId          String
  projectId        String
  graphId          String
  voteAmount       String
  voteTimestamp    String
  voterAddress     String
  voteToAddress    String
  voteToken        String
  voteVersion      String
```

The PR also includes a seeding script - which fills the votes (and upserts the project/rounds) table with _all_ votes pulled from the graph. 

Having this data will be very useful. 

**Todo**
- [ ] Figure out if the graph is doing wonky business (the responses from large queries are inconsistent 💀 )
- [x] Add Optimism
- [x] Add Fantom
- [x] Testnets
- [ ] Tests
- [ ] Project Ids and future proofing
- [ ] TBD

After this is complete, we can start refactoring our endpoints to utilize the cached data. 

We will also need a service to watch for new votes and subsequently append votes to the new table as they are emitted.

### DIFF
--- a/packages/api/prisma/schema.prisma
+++ b/packages/api/prisma/schema.prisma
@@ -13,47 +13,109 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
-model Round {
-  id                 Int            @id @default(autoincrement())
-  createdAt          DateTime       @default(now()) @db.Timestamptz(3)
-  updatedAt          DateTime       @updatedAt @db.Timestamptz(3)
-  chainId            ChainId
-  roundId            String         @unique
-  votingStrategyName VotingStrategy
-  matches            Match[]
-  isSaturated        Boolean        @default(false)
-  roundSummary       RoundSummary?
-  projects           Project[]
-  votes              Vote[]
+model Program {
+  id               Int      @id @default(autoincrement())
+  createdAt        DateTime @default(now()) @db.Timestamptz(3)
+  updatedAt        DateTime @updatedAt @db.Timestamptz(3)
+  chainId          String
+  programId        String
+  programCreatedAt String
+  programUpdatedAt String
+  rounds           Round[]
 
+  @@unique([chainId, programId], name: "programIdentifier")
+  @@map("program")
+}
+
+model Round {
+  id                    Int       @id @default(autoincrement())
+  createdAt             DateTime  @default(now()) @db.Timestamptz(3)
+  updatedAt             DateTime  @updatedAt @db.Timestamptz(3)
+  chainId               String
+  program               Program   @relation(fields: [chainId, programId], references: [chainId, programId])
+  programId             String
+  roundId               String
+  roundCreatedAt        String
+  roundUpdatedAt        String
+  applicationsStartTime String
+  applicationsEndTime   String
+  roundStartTime        String
+  roundEndTime          String
+  payoutStrategy        String
+  roundToken            String
+  projects              Project[]
+  matches               Match[]
+  roundSummary          RoundSummary[]
+
+  @@unique([chainId, roundId], name: "roundIdentifier")
   @@map("round")
 }
 
-model Project {
-  id               Int              @id @default(autoincrement())
-  createdAt        DateTime         @default(now()) @db.Timestamptz(3)
-  updatedAt        DateTime         @updatedAt @db.Timestamptz(3)
-  chainId          ChainId
-  round            Round            @relation(fields: [roundId], references: [roundId])
-  roundId          String
-  projectId        String
-  projectSummaries ProjectSummary[]
-  votes            Vote[]
+model VotingStrategy {
+  id              Int      @id @default(autoincrement())
+  createdAt       DateTime @default(now()) @db.Timestamptz(3)
+  updatedAt       DateTime @updatedAt @db.Timestamptz(3)
+  roundId         String
+  chainId         String
+  strategyAddress String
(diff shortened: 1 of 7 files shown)


packages/api/src/utils.ts
```diff
@@ -674,3 +674,83 @@ export const getValidCoinGeckoTokenAddress = (
   }
   return validAddress;
 };
+
+// TODO: Docs
+export const fetchGraphQFContributionsForRound = async (
```
thelostone-mc, line 679: Could we add JSDocs comment for this ?

packages/api/src/database.ts
```diff
@@ -14,6 +16,75 @@ export class DatabaseInstance {
...
 
+  // TODO: createVoteRecord for singular vote, getVotes, methods
+  async createVoteRecords(
+    chainId: string,
+    strategyName: VotingStrategy,
+    data: GraphResponse<GraphQFVotes>,
+  ) {
+    try {
+
+      const voteData : any = [];
```
thelostone-mc, line 27: let's avoid using any

packages/api/prisma/schema.prisma
```diff
@@ -91,6 +93,26 @@ model ProjectSummary {
   @@map("projectSummary")
 }
 
+model Vote {
+  id               Int      @id @default(autoincrement())
```
thelostone-mc, line 97: It would be useful to have a table with voting strategies and an entry for QFVote
And a vote would have a link to that VotingStrategy entry

packages/api/prisma/schema.prisma
```diff
@@ -91,6 +93,26 @@ model ProjectSummary {
...
+  createdAt        DateTime @default(now()) @db.Timestamptz(3)
+  updatedAt        DateTime @updatedAt @db.Timestamptz(3)
+  round            Round    @relation(fields: [roundId], references: [roundId])
+  roundId          String
+  project          Project  @relation(fields: [projectId, roundId], references: [projectId, roundId])
+  projectId        String
```
thelostone-mc, line 103: Can't projectId be derived from project ? Is this mainly for generating the unique entry ?

packages/api/prisma/schema.prisma
```diff
@@ -91,6 +93,26 @@ model ProjectSummary {
...
+  graphId          String
+  voteAmount       String
+  voteTimestamp    String
+  voterAddress     String
+  voteToAddress    String
+  voteToken        String
+  voteVersion      String
+
+  @@unique([roundId, projectId, graphId], name: "voteIdentifier")
```
thelostone-mc, line 112: Can't just the graphId be unique ?

packages/api/prisma/schema.prisma
```diff
@@ -37,6 +38,7 @@ model Project {
   roundId          String
   projectId        String
   projectSummaries ProjectSummary[]
+  votes            Vote[]
```
thelostone-mc, line 41: Shouldn't this be optional ? Not every project would have a vote right ?

packages/api/prisma/schema.prisma
```diff
@@ -24,6 +24,7 @@ model Round {
   isSaturated        Boolean        @default(false)
   roundSummary       RoundSummary?
   projects           Project[]
+  votes              Vote[]
```
thelostone-mc, line 27: Shouldn't this be optional ? Not every project would have a vote right ?

packages/api/prisma/schema.prisma
```diff
@@ -91,6 +93,26 @@ model ProjectSummary {
   @@map("projectSummary")
 }
 
+model Vote {
+  id               Int      @id @default(autoincrement())
+  createdAt        DateTime @default(now()) @db.Timestamptz(3)
+  updatedAt        DateTime @updatedAt @db.Timestamptz(3)
+  round            Round    @relation(fields: [roundId], references: [roundId])
+  roundId          String
```
thelostone-mc, line 101: Can't roundId be derived from round ? Is this mainly for generating the unique entry ?

thelostone-mc (commented):
Left a few comments

packages/api/prisma/schema.prisma
hmrtn: The array is optional! There can be no elements!

packages/api/prisma/schema.prisma
hmrtn: Unique ids - now along with the chain, too

hmrtn:
Do not merge - there are no tests.

However, I would like to encourage you to run the seed process. All tables should fill out for graph replication by the block number. 

If we include this soon - it will be much more consistent with our calculations.

Why? Because the graph is inconsistent (even with a set blocknumber!?) so effectively, every time we run the calcs in our current implementation - they will differ a lot. 

This would allow a reproducible snapshot for our calculations - but it still needs to be guaranteed accurate. We _really_ need to figure out how the graph works. 

For example - not setting a block number will use different indexers and thus return different results..
I will speak more on this during standup.

thelostone-mc:
overall looks good ! i'll pull this in and run it locally and give the seeding a shot

thelostone-mc:
Closing this out as we are relying on the indexer to index this

PR #1115, opened 2023-02-09
Enforce socials when applying to a round

boudra:
##### Description

Screenshot 2023-02-09 at 17 35 49

### DIFF
--- a/packages/builder/src/components/application/Form.tsx
+++ b/packages/builder/src/components/application/Form.tsx
@@ -1,18 +1,25 @@
 import { Stack } from "@chakra-ui/react";
 import { datadogRum } from "@datadog/browser-rum";
-import { ExclamationTriangleIcon } from "@heroicons/react/24/solid";
+import {
+  ExclamationTriangleIcon,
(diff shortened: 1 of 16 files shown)


packages/builder/src/components/application/Form.tsx
```diff
@@ -200,6 +220,15 @@ export default function Form({
     handleSubmitApplication();
   };
 
+  const editProjectPath = (id: string) => {
```
gravityblast, line 223: @boudra we can maybe move it to `routes` like `projectPathByID`

packages/builder/src/components/application/Form.tsx
```diff
@@ -220,10 +249,55 @@ export default function Form({
     setProjectOptions(currentOptions);
   }, [props.allProjectMetadata]);
 
+  const projectRequirementsResult = [];
+
+  console.log(roundApplication.applicationSchema.requirements);
```
gravityblast, line 254: forgot log maybe?

gravityblast (approved):

PR #1117, opened 2023-02-09
[DevRel] Smart Contract documentation + readme

ZakkMan:
##### Description

First pass at a readme and documentation for the smart contracts

##### Testing

Just adding documentation - not touching any of the code


packages/contracts/README.md
```diff
@@ -1,10 +1,50 @@
-# grants-round/contracts
+# Allo Contracts
 
-This project is built using [hardhat](https://hardhat.org).
-All the contracts required to run / manage a round can be found here.
-If you come across any vulnerabilties, do create an issue / raise a PR to help improve the contracts. ^_^
+Allo Protocol is a set of smart contracts that enable the democratic allocation
+and distribution of funds through the power of [Quadratic
+Funding](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3243656). It is the
```
thelostone-mc, line 5: This isn't entirely true 
- centralised Gitcoin used pairwise QF while we use a light weight linear QF
- Allo doesn't restrict itself to just QF as it's build to allow folks to add their own voting strategies like QV / Maci. Is it worth rephrasing that to state it supports add-ons but we chose to lead with QF

packages/contracts/SETUP.md
```diff
@@ -0,0 +1,62 @@
+# Setup
+
+## Terminology
+
+- **Program Operators**: wallets that have the permission to create & manage the program
+- **Program**: maintained by Program Operators which together form a team
+- **Round Operators**: wallets that have the permission to create & manage the round
+- **Round**: created by a Program and deal with everything relating to running a round
+- **Voter** : wallet who cast a vote to a grant during the round
```
thelostone-mc, line 9: ^ Would this belong in the setup ? It was originally intended to be more product documentation but we never got to that

thelostone-mc (changes requested):
Could we avoid the contract changes as we have a big change coming in for payout piece. This would result in a lot of conflicts.

Sidenote: what's the reason for moving away from multiline style comments in the solidity

packages/contracts/SETUP.md
ZakkMan: I'm going to remove it and have product documentation live in a Gitbook

packages/contracts/SETUP.md
```diff
@@ -0,0 +1,62 @@
...
+4. Link `RoundImplementation` to `RoundFactory` contract
+
+
+## Program Setup
+
+1. To create a program, you would not deploy a contract but instead, rely on the create function on the `ProgramFactory` to create a clone of the already deployed `ProgramImplementation` contract
+2. Any interaction in terms of updating parameters etc can be performed against the `ProgramImplementation` contract itself
+
+
+The ProgramFactory enables us to have upgradable contracts on ProgramImplementation
```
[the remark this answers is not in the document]
ZakkMan: the multiline style is technically accepted but everyone uses the three dashes (`///`) and most text editors will highlight it a different color from comments (which you don't get with multiline comments)

thelostone-mc:
Closing this out as contacts are hosted in Allo org

PR #1127, opened 2023-02-10
fix: tweaked error message for passport failures, added Datadog logging

### LINKED ISSUE #1113: Error fetching Passport score
**Overview**
Passport is erroring out on [Explorer](https://grant-explorer.gitcoin.co/#/round/1/0x1b165fe4da6bc58ab8370ddc763d367d29f50ef0/passport/connect)

**Screenshots**
image

DanieleSalatti:
Fixes #1113 

This makes the error message a bit more clear to the user (and us) when the passport API returns an error.

I also added Datadog logging to our passport connection logic to make it easier to troubleshoot in case we don't have a way to repro (like in this case).

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -7,6 +7,7 @@ export enum PassportState {
   MATCH_INELIGIBLE,
   LOADING,
   ERROR,
+  INVALID_RESPONSE,
 }
 
 type PassportEvidence = {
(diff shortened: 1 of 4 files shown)


thelostone-mc (approved):

melissa-neira:
I agree with @nfrgosselin 🤝

PR #1134, opened 2023-02-10
Fix/providers

bitbeckers:
* Added alchemy provider to config because Polygon Mumbai is not available via Infura (without CC) and the providers started 429-ing with the default keys


bitbeckers:
Wrong repo

PR #1138, opened 2023-02-10
fix: missing peer deps

DanieleSalatti:
Missing peer dependencies in the `pnpm-lock.yaml` seems to be the culprit for the current outage.

Had to change the type of our icons on the support page after that.


thelostone-mc:
Closing this out as it's been resolved

PR #1168, opened 2023-02-14
fix: fantom upload hotfix

hmrtn:


hmrtn:
Do not merge. It is to provide a preview link for fantom with unsafe changes.

PR #1174, opened 2023-02-16
fix: transaction cancelled error handling

### LINKED ISSUE #1069: Explorer: Donation Progress stops on step 2 but the transaction still goes through
**Overview**
I'm seeing this TRANSACTION_REPLACED error in the logs. Points to the transaction being repriced due to an increase in gas fee.

**Steps To Reproduce**
No known reproduction yet. We have a sample TX: https://etherscan.io/tx/0x1b7dace8eaf33f2e14fe57458f90db835cccf727944a66d339dcf415a7dc492d in which  the user got the error but the transaction went through

**Observed  behavior**
 While donating, the user gets an error message on the frontend and the donation is stuck on the 2nd step however the transaction is successful
image

**Expected behavior**
We should not error out the transaction during a repricing

**Screenshots**
https://www.loom.com/share/9ca37644afed4ae2934e3ff71b683b60
image

hmrtn:
note: we need to start thinking about proper error handling with application specific error codes and a new error class.

fixes #1069

### DIFF
--- a/packages/grant-explorer/src/features/round/ViewBallotPage.tsx
+++ b/packages/grant-explorer/src/features/round/ViewBallotPage.tsx
@@ -1175,7 +1175,6 @@ export default function ViewBallot() {
       });
     } catch (error) {
       if (error === Logger.errors.TRANSACTION_REPLACED) {
-        console.error("Transaction replaced:", error);
         setTransactionReplaced(true);
       } else {
         datadogLogs.logger.error(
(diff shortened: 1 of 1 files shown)


packages/grant-explorer/src/features/round/ViewBallotPage.tsx
```diff
@@ -1165,10 +1174,15 @@ export default function ViewBallot() {
         votingStrategy: round.votingStrategy,
       });
     } catch (error) {
-      datadogLogs.logger.error(
-        `error: handleSubmitDonation - ${error}, id: ${roundId}`
-      );
-      console.error("handleSubmitDonation - roundId", roundId, error);
+      if (error === Logger.errors.TRANSACTION_REPLACED) {
+        console.error("Transaction replaced:", error);
```
thelostone-mc, line 1178: Should we remove this ? This would get logged on sentry

melissa-neira:
defer to engineers 🦕

vacekj (approved):

PR #1198, opened 2023-02-17
fix: add alchemy/infura key usage and priority to wagmi

### LINKED ISSUE #1173: Use Grants Stack keys for infura/alchemy
**Overview**
The round manager (and maybe the explorer as well) is using infura/alchemy with the test default keys taken from ethers js.
When we have many users on the same domain using the same key some users have a `429 Too Many Requests` error.

**Steps To Reproduce**
1. Go to the round manager page
2. Login with your account
3. In the chrome console, filter network requests with the string `infura`
4. Each request URL ends with the default test project ID: `84842078b09946638c03157f83405213 `

**Expected behavior**

We should have our own key/projectID for each chain. 
Since we load from different chains, we can't have the same applicationID or key, since each infura/alchemy key/application is for a different chain.

We can implement it like we already do in `builder` 

https://github.com/gitcoinco/grants-round/blob/f8922096719a0ed516609a39da352031e1719ebe/packages/builder/src/utils/wagmi.ts#L72-L74

**Additional context**

hmrtn:
note: I added the Infura env var and left the Alechmy one as a dummy var, until we sort it out, to our fleek round manager deployments. 
Be sure to check the new .env.example. 

It will still fallback to the public key if not on mainnet because we have not setup infura/alchemy for each project and chain (which incurs additional costs). 

But at least on mainnet - it should use our production infura key. 

fixes #1173

### DIFF
--- a/packages/round-manager/.env.sample
+++ b/packages/round-manager/.env.sample
@@ -11,6 +11,7 @@ REACT_APP_SUBGRAPH_FANTOM_MAINNET_API=################
 
 # Ethereum Providers
 REACT_APP_INFURA_ID=######################
+REACT_APP_ALCHEMY_ID=######################
 
 # Datadog Monitoring
 REACT_APP_DATADOG_APPLICATION_ID=######################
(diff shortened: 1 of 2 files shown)


melissa-neira:
defer to engineers 💅

vacekj (approved):

vacekj:
@hmrtn updated the description to mention fixing the issue so it's closed automatically. also, could you please create the branch using the following button in the issue? keeps the branch names nice and standard. Thanks!
image

PR #1204, opened 2023-02-21
chore: refactor fee

thelostone-mc:
##### Description

This has 2 commits

- [chore: refactor updateStartAndEndTime](https://github.com/gitcoinco/grants-round/commit/632c4eb9ae7ea6cb40bafcd5024103d7816ee024) which cleans up the logic and ensure round times and application times can overlap 

- [chore: add fee on top of match amount](https://github.com/gitcoinco/grants-round/commit/a29dd6f674c8c40dfd6fe71a10ae57ee7ad26dd9) which adds the fee on top of the matchAmount. 
This commit is isolated for the sole reason for being able to rollback if the team decides we want to have the fee amount to be taken from the match Amount 

Both commits add additional tests to ensure functionality works as expected.
This is to be reviewed and only then merged


melissa-neira:
defer to engineers 🐣

PR #1228, opened 2023-03-13
chore: add merkle tree util function

### LINKED ISSUE #1195: chore: Update data scheme when uploading data to payout contract
### Description

When the operator wants to upload distribution onto the payout contract  with the final distribution.
The flow needs to be updated to ensure

- We generate the Merkle root and Merkle proofs using OZ library 
- Ensure the Merkle root is stored on the payout contract (current we store it as an empty string)
- Ensure the MetaPtr includes
  - the actual distribution 
  - the Merkle proof
  - the schema version

```javascript
{ 
  version: "1.0.0",
  type: "MerklePayoutStrategy",
  distribution:[],
  merkleProof: []
}

```

thelostone-mc:
##### Description

- Add util function to generate merkle tree
- Update data distribution to ensure merkle root is generated
- Add new parameters on the distribution
  - index
  - matchingAmountInToken
  - projectPayoutAddress

##### Refers/Fixes

fixes #1195

### DIFF
--- a/packages/round-manager/src/context/round/FinalizeRoundContext.tsx
+++ b/packages/round-manager/src/context/round/FinalizeRoundContext.tsx
@@ -19,6 +19,8 @@ import {
 } from "../../features/api/round";
 import { datadogLogs } from "@datadog/browser-logs";
 import { ethers } from "ethers";
+import { generateMerkleTree } from "../../features/api/utils";
+import { formatBytes32String } from "ethers/lib/utils";
 
 export interface FinalizeRoundState {
   IPFSCurrentStatus: ProgressStatus;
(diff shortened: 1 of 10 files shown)


DanieleSalatti (approved):
LGTM

nfrgosselin:
Defer to engineering review

melissa-neira:
defer to the devs 🫡

packages/round-manager/src/context/round/FinalizeRoundContext.tsx
```diff
@@ -225,10 +230,8 @@ async function finalizeToContract(
       payload: { finalizeRoundToContractStatus: ProgressStatus.IN_PROGRESS },
     });
 
-    const merkleRoootInBytes = ethers.utils.formatBytes32String(merkleRoot);
-
     const encodedDistribution = encodeDistributionParameters(
-      merkleRoootInBytes,
+      merkleRoot,
       distributionMetaPtr
     );
```
bhargavaparoksham, line 236: @thelostone-mc  finalizeRoundToContract function which is called after encoding parameters, calls updateDistribution function via roundImplementation. 
updateDistribution has been removed in roundImp v2.  So this needs to be updated as well for this to work

bhargavaparoksham (changes requested):

PR #1253, opened 2023-03-20
feat: add static stats page

### LINKED ISSUE #810: Add round stats tab + layout
**User Story:**
As a round operator
I want to see my round stats
So that I can monitor progress of my round

**Acceptance Criteria**
GIVEN that I am on the round dashboard
WHEN I click on the Round Stats link in the nav column
THEN I am brought to the Round Stats tab
AND the Round Stats link is set to selected state

GIVEN that I am on the Round Stats tab
WHEN I look at the page
THEN I see the Overview and Matching Funds section with all boxes and their titles

* Note — you don't need to enter any data into the boxes in this step. Just create the layout

**Product & Design Links:**
QF: [Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=208-21959&t=OmGjFYjhb8zvYEhF-4)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

vacekj:
fixes #810 as describe in the issue

also adds some miscellaneous fixes


melissa-neira (changes requested):
Looks good, but I would change a couple of things.
First, the header isn't very responsive-friendly. The "view on grants explorer" button is supposed to align-right with the rainbow kit component as well as the right-most stat boxes.
Here's what I see on my end:
screencapture-bafybeicstzbfffmq5umuuqlolve3ylml7e3qyflx4ckgptodekcj5yxggy-on-fleek-co-2023-03-21-11_09_30
^on 1440px width (grant explorer button slightly more left-aligned)

screencapture-bafybeicstzbfffmq5umuuqlolve3ylml7e3qyflx4ckgptodekcj5yxggy-on-fleek-co-2023-03-21-11_09_45
^on my 32" monitor (left-alignment more noticeable)

The other thing is that there should be more spacing between the matching stats and the footer. Currently, there seems like there's no spacing between these 2 items. See [Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=208-21959&t=RqGLo5DZ4JUf0ASb-4) for reference.

Let me know if you have any questions!! 🕺

packages/round-manager/src/features/round/ViewRoundPage.tsx
```diff
@@ -63,6 +63,11 @@ export default function ViewRoundPage() {
       setRoundExists(!!round);
 
       if (round) {
+        /* During development, give frontend access to all rounds */
+        if (process.env.NODE_ENV === "development") {
```
boudra, line 67: We might want to test that round access works in development, so using a debug mode environment flag or a similar alternative instead of the Node env is a better approach here. What do you think?

vacekj: I dont' see your point. You can set the NODE_ENV variable yourself if you want to change it during testing or on CI.

boudra: Yes, you can change it. But my point is that they mean completely different things, the `NODE_ENV` specifies the environment you're currently running on and you're trying to disable some feature to help test. These two things are not related, because there will be situations where you might want to test this specific feature you're disabling while keeping the `NODE_ENV` as `development`. You're also breaking feature parity with production for the whole development team, this will be confusing for team members that don't know this exists, it is safer to have the same features as production by default and disable them manually by changing your env.

vacekj: Ah now I get it - fixed by introducing a new IGNORE_FRONTEND_CHECKS environment variable.

gravityblast (approved):

boudra (approved):

packages/round-manager/src/features/round/ViewRoundPage.tsx
```diff
@@ -65,6 +65,11 @@ export default function ViewRoundPage() {
       setRoundExists(!!round);
 
       if (round) {
+        /* During development, give frontend access to all rounds */
+        if (process.env.REACT_APP_IGNORE_FRONTEND_CHECKS) {
```
hmrtn, line 69: Should we note this new env var and use it elsewhere,too?

hmrtn (approved):

packages/round-manager/src/features/round/ViewRoundPage.tsx
vacekj: Noted in .env.example. Use elsewhere would be great but oos for this pr.

PR #1270, opened 2023-03-22
Populate stats page with real data

### LINKED ISSUE #811: Populate Overview stats
**User Story:**
As a round operator
I want to see my round stats
So that I can monitor progress of my round

**Acceptance Criteria**
GIVEN that I am on the round stats tab
WHEN I look at Est. Donations Made
THEN I see the total USD value of donations made in the round

GIVEN that I am on the round stats tab
WHEN I look at Matching Funds Available
THEN I see the total USD value of the matching pool

GIVEN that I am on the round stats tab
WHEN I look at Unique Contributors
THEN I see the total number of unique wallets that have donated

GIVEN that I am on the round stats tab
WHEN I look at No. of Contributions
THEN I see the total number of contributions (contributions = unique wallet donating to one project)

**Product & Design Links:**
[Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=208-21959&t=OmGjFYjhb8zvYEhF-4)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

vacekj:
fixes #811

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -1,4 +1,6 @@
 import useSWR from "swr";
+import { useParams } from "react-router";
+import { isAddress } from "viem";
 import {
   ChainId,
   graphql_fetch,
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -130,7 +130,7 @@ export type Payout = {
   token: string;
   version: string;
   createdAt: string;
-}
+};
 
 /**
  * Fetches the payouts that happened for a given round from TheGraph
@@ -138,12 +138,15 @@ export type Payout = {
  * @param chainId Chain ID
  * @returns
  */
-export function fetchProjectPaidInARound(roundId: string, chainId: ChainId): Promise<Payout[]> {
+export function fetchProjectPaidInARound(
+  roundId: string,
+  chainId: ChainId
+): Promise<Payout[]> {
   const { data, error, mutate } = useSWR(
     [roundId, chainId],
     ([roundId, chainId]: [roundId: string, chainId: ChainId]) => {
       return graphql_fetch(
-      `
+        `
         query GetPayouts($roundId: String) {
           payoutStrategies(
             where:{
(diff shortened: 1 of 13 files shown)


vacekj:
Needs cleanup and error handling, otherwise feature complete.

packages/common/src/index.ts
```diff
@@ -1,4 +1,6 @@
 import useSWR from "swr";
+import { useParams } from "react-router";
+import { isAddress } from "viem";
```
gravityblast, line 3: can't we use `ethers` for this since it's already a dependency of the 3 apps using this `common` package?

gravityblast: I actually see that the viem implementation it's the following, so it would be better to copy it inside instead of including another big package like this:

```
const addressRegex = /^0x[a-fA-F0-9]{40}$/
export function isAddress(address: string): address is Address {
  return addressRegex.test(address)
}

```

vacekj: viem is treeshakable, so in this case, we are only importing this exact function, and the impact on the budle size would therefore be exactly the same.

vacekj: ethers is fine but it's not threeshakeable as far as I am aware, so it's better to include viem since it's tiny rather than potentially duplicating another huge ethers version into our bundle.

hmrtn: We should just use ethers.utils.isAddress

melissa-neira (approved):
looks good to me! I can't really test it though because I don't have any rounds with donations 😅

vacekj:
@melissa-neira 0xD95A1969c41112cEE9A2c931E849bCef36a16F4C here's the web3 round from alpha grants.

vacekj:
@nfrgosselin @melissa-neira @gitcoinco/grantsround-engineering requesting one last review after rebase and fixes. The table is not populated, but this PR is already huge so let's create a new ticket for that, since it's also blocked by the matching calcs which won't be done quickly. Let's ship this and keep momentum, also avoiding stalled 90% tickets as per retro 🙌

melissa-neira (approved):

hmrtn (approved):
Thanks for fixing the conflicts. One thing to note is the update to node 18 with the rebase from main. Perhaps we should be utilizing the nvmrc.

vacekj:
@hmrtn I think adding the "engine" field into package.jsons would be enough.

Issue #1331, opened 2023-04-03
Manager branding updates


nfrgosselin:
**Acceptance Criteria**
- [x] update landing page app name to "Manager"
- [x] update header branding / assets 

**Product & Design Links:**
[Figma with assets](https://www.figma.com/file/78IRtNC6ZRZP8wg6Y2v01U/GPC-x-MMM?node-id=272-3769&t=3a2gk9smsfknDtnd-4)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

hmrtn:
We don't have a light or dark toggle on round manager. So I implemented the light version only. 

If we want that, there should be a separate ticket.

PR #1339, opened 2023-04-03
add: rebrand to new design

hmrtn:
#1331


codenamejason (approved):
Looks good 

One of the gitcoin logos looks a bit smaller when I run it:

Screenshot 2023-04-03 at 9 29 16 PM
Screenshot 2023-04-03 at 9 29 27 PM

melissa-neira (commented):
Looks good to me! We should keep the dark version of the nav bar though - not sure why MMM threw a light version in there but let's stick to the original color scheme for now. 

Also, we should probably update the favicon for Manager. Let me know if that needs a separate ticket 🕺

PR #1358, opened 2023-04-05
feat: prevent project from reapplying to round

thelostone-mc:
##### Description

- add warning alert 
- fetch data from subgraph to see if application exists

fixes https://github.com/gitcoinco/grants-stack/issues/1327


packages/builder/src/actions/projects.ts
```diff
@@ -469,6 +537,7 @@ export const fetchProjectApplications =
             (app: Application) => app.roundID
           );
 
+          // TODO: REMOVE AS THIS IS NOT NECESSARY
           dispatch<any>(
             fetchApplicationStatusesFromContract(
```
thelostone-mc, line 542: THIS WILL BREAK cause applications are on-chain

packages/builder/src/actions/projects.ts
```diff
@@ -1,26 +1,25 @@
...
 import { ethers, utils } from "ethers";
 import { Dispatch } from "redux";
+import { convertStatusToText } from "common";
 // import ProjectRegistryABI from "../contracts/abis/ProjectRegistry.json";
 import {
   Client as AlloClient,
   Application as GrantApplication,
 } from "allo-indexer-client";
-import RoundImplementationABI from "../contracts/abis/RoundImplementation.json";
+// import RoundImplementationABI from "../contracts/abis/RoundImplementation.json";
```
hmrtn, line 10: same here

packages/builder/src/actions/projects.ts
```diff
@@ -469,16 +475,16 @@ export const fetchProjectApplications =
...
-
-          dispatch<any>(
-            fetchApplicationStatusesFromContract(
-              roundAddresses,
-              projectID,
-              projectApplicationID,
-              chain.id
-            )
-          );
+          // const roundAddresses = applications.map((app: Application) => app.roundID);
```
hmrtn, line 478: remove deprecated code 🧑‍🚒

packages/builder/src/actions/projects.ts
```diff
@@ -445,15 +449,17 @@ export const fetchProjectApplications =
           );
 
           if (response.errors) {
-            throw response.errors;
```
hmrtn, line 448: there should be some logging if the application fetching fails

packages/common/src/index.ts
```diff
@@ -294,3 +294,17 @@ export enum VerifiedCredentialState {
   INVALID,
   PENDING,
 }
+
+export const convertStatusToText = (applicationStatus: string | number) => {
```
hmrtn, line 298: `convertStatusToText` does not coerce a string into a number for the switch statement. If applicationStatus is a string, the switch statement will not match any cases, and the function will return "PENDING" due to the default case.

Perhaps we should use an enum here for application status. Maybe something like:
```
enum ApplicationStatus {
  PENDING = "0",
  APPROVED = "1",
  REJECTED = "2",
  CANCELLED = "3",
}

export const convertStatusToText = (applicationStatus: string | number): string => {
  // Ensure the applicationStatus is a string
  applicationStatus = applicationStatus.toString();

  switch (applicationStatus) {
    case ApplicationStatus.CANCELLED:
      return "CANCELLED";
    case ApplicationStatus.REJECTED:
      return "REJECTED";
    case ApplicationStatus.APPROVED:
      return "APPROVED";
    case ApplicationStatus.PENDING:
    default:
      return "PENDING";
  }
};
```

packages/builder/src/actions/projects.ts
```diff
@@ -1,26 +1,25 @@
-import { datadogLogs } from "@datadog/browser-logs";
 import { datadogRum } from "@datadog/browser-rum";
 import { ethers, utils } from "ethers";
 import { Dispatch } from "redux";
+import { convertStatusToText } from "common";
 // import ProjectRegistryABI from "../contracts/abis/ProjectRegistry.json";
```
hmrtn, line 5: remove this commented import

hmrtn (changes requested):
just a couple of small things!

PR #1468, opened 2023-04-13
Fix Builder Image Sizes

boudra:
##### Description

- Simplify the image cropping code
- Apply a resize to avoid uploading huge images
- Use JPEG instead of PNG for a x5 image size reduction

This PR + changing our IPFS gateway to our CDN (https://ipfs-grants-stack.gitcoin.co) should reduce our costs to almost nothing.

In the future we can also optimize and convert images to other formats (like webp) on the fly based on the user agent, we can do this with Cloudfront.

More discussion here:

https://github.com/gitcoinco/grants-stack/issues/1395
https://github.com/gitcoinco/grants-stack/discussions/1129

### DIFF
--- a/packages/builder/src/components/base/ImageInput.tsx
+++ b/packages/builder/src/components/base/ImageInput.tsx
@@ -201,7 +201,7 @@ export default function ImageInput({
         dimensions={dimensions}
         onClose={() => setShowCrop(false)}
         saveCrop={(imgUrl) => {
-          imgUrl.toBlob((blob) => blob && imgHandler(blob));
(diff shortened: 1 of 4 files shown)


melissa-neira (approved):

gravityblast (approved):

vacekj:
should we enable webp support now? all modern browsers support it: https://caniuse.com/webp

boudra:
> should we enable webp support now? all modern browsers support it: https://caniuse.com/webp

I suggest we deploy and test this first then we can look into this since it's a bit more work, I don't think browser support will be an issue but we deffo need to respect the request headers sent by the browsers and serve the right image format.

PR #1476, opened 2023-04-14
Twitter and Github Verification Fixes

boudra:
##### Description

This PR simplifies and adds better error handling amongst other refactorings.

- The BroadcastChannel callback happens inside React, this is not ideal because React could re-render the component or even not render at all, there could be a redirect etc. Since this is not related to the UI at all it now happens in the index script before anything else, this will ensure things are being called.
- Separate unexpected errors from Verification errors and surface them to the user instead of silencing, also callback with errors from OAuth which at the moment they're silenced as well.
- Unify all the logic to create credentials from OAuth instead of having relying on `useEffect`, this will ensure errors happening with this procedure will always be caught instead of just doing nothing
- Give feedback to the user if the pop up doesn't open, if it's blocked for example
- Handle and surface errors from Passport when verifying the credentials

WIP: at the moment this is only for twitter

### DIFF
--- a/packages/builder/src/components/grants/Landing.tsx
+++ b/packages/builder/src/components/grants/Landing.tsx
@@ -80,7 +80,7 @@ function Landing() {
     if (queryCode) {
       channel.postMessage({
         target: "twitter",
-        data: { code: queryCode, state: queryState },
+        data: { error: queryError, code: queryCode, state: queryState },
       });
     }
     // always close the redirected window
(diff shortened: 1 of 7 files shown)


codenamejason (approved):
Awesome job @boudra

jesuloluwao:
@codenamejason @boudra Hey I still have a customer unable to verify github. I asked them to make a video and they sent https://drive.google.com/file/d/1X0O9la1wlt7Tq8MNd-yNjUJlf27h30uq/view?usp=share_link. I was also able to verify that they have a public profile but they get a 403 error with a bad request

jesuloluwao:
In the video above, I've seen the issue is due to the case sensitivity in the organization name. Can we also include this in the error message in the UI?

PR #1477, opened 2023-04-14
1320 prevent user from over funding contract

codenamejason:
This prevents a user from depositing more funds then necessary into the contract for payouts in the UI only.

- Disables Button
- Disables Input
- Updates placeholder with message

Screenshot 2023-04-14 at 9 40 22 AM


melissa-neira:
@nfrgosselin did we want to go this route or let the user keep funding the contract if they want to?

nfrgosselin:
@melissa-neira I like having this in place for now. The calculator works from the amount listed as the total pot size, so they'll need to update that value anyway to make sure the extra funds are distributed

nfrgosselin (approved):

thelostone-mc (approved):

PR #1485, opened 2023-04-17
fix: remove usd if NaN or 0

codenamejason:
Screenshot 2023-04-17 at 10 52 08 AM

### DIFF
--- a/packages/round-manager/src/features/round/FundContract.tsx
+++ b/packages/round-manager/src/features/round/FundContract.tsx
@@ -304,7 +304,7 @@ export default function FundContract(props: {
               </ReactTooltip>
             </span>
           </p>
-          <p className="text-sm">{props.round.protocolFeePercentage}%</p>
+          <p className="text-sm">{props.round.protocolFeePercentage ?? 0}%</p>
         </div>
         <div className="flex flex-row justify-start mt-6">
           <p className="flex flex-row text-sm w-1/3">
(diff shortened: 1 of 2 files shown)


thelostone-mc (approved):

nfrgosselin (changes requested):
Now I'm not seeing any price data. Is that correct?

image

packages/round-manager/src/features/round/FundContract.tsx
```diff
@@ -349,9 +353,11 @@ export default function FundContract(props: {
           <p className="text-sm">
             {" "}
             {totalAmountLeftToFund} {matchingFundPayoutToken?.name}{" "}
-            <span className="text-sm text-slate-400 ml-2">
-              ${amountLeftToFundInUSD} USD
-            </span>
+            {amountLeftToFundInUSD > 0 ? (
```
DanieleSalatti, line 356: You should be able to use `&&` here and in other places in this PR I think?

DanieleSalatti (approved):
LGTM, left a minor comment

packages/round-manager/src/features/round/FundContract.tsx
codenamejason: I didn't want to show zero value

thelostone-mc:
Just tested it locally and it worked 🙌🏼

PR #1498, opened 2023-04-18
fix: check for token undefined as condition to refetch round

DanieleSalatti:
##### Description

Fixes the same issue we had on the round manager


Ameralameri:
Go

thelostone-mc (approved):

Issue #1342, opened 2023-04-04
Hook up stats page to calculator


nfrgosselin:
**User Story:**
As a program manager
I want to monitor the progress of my round
So that I can keep tabs on how my community's funds are being distributed

**Acceptance Criteria**
GIVEN that I am on the Round Stats tab of my round dashboard
WHEN I look at the Current Matching Funds tab
THEN I see the current distribution as determined by the calculator

**Product & Design Links:**
[Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=208-21959&t=WM3Rxzk31pmLitZG-4)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

PR #1520, opened 2023-04-20
Hook up round stats page to calculator

hmrtn:
wip branch for #1342 

depends on: https://github.com/gitcoinco/allo-indexer-client/pull/8 https://github.com/gitcoinco/allo-indexer/pull/31

### DIFF
--- a/packages/round-manager/src/features/round/ViewRoundStats.tsx
+++ b/packages/round-manager/src/features/round/ViewRoundStats.tsx
@@ -52,8 +52,8 @@ export default function ViewRoundStats() {
 
   const { data: matches } = useRoundMatchingFunds(roundId);
 
-  const matchAmount = round?.matchAmount;
   const matchAmountUSD = round?.matchAmountUSD;
+  console.log(round);
 
   return (
     <div className="flex flex-center flex-col mx-auto mt-3 mb-[212px]">
@@ -73,10 +73,7 @@ export default function ViewRoundStats() {
           title={"Est. Donations Made"}
         />
         <StatsCard
-          text={
-            matchAmount &&
-            Number(utils.formatEther(matchAmount.toString())).toFixed(2)
-          }
+          text={matchAmountUSD && matchAmountUSD.toFixed(2)}
           title={"Matching Funds Available"}
         />
         <StatsCard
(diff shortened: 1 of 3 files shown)


packages/round-manager/src/features/round/ViewRoundStats.tsx
```diff
@@ -57,29 +64,27 @@ export default function ViewRoundStats() {
...
+              style: "currency",
+              currency: "USD",
+            }).format(round.amountUSD)
           }
           title={"Est. Donations Made"}
         />
         <StatsCard
-          text={matchAmount ? matchAmount[0] : "-"}
+          text={
+            matchAmount &&
```
hmrtn, line 77: one thing to note here that may be confusing to the user: is this token amount, or USD amount displayed in the UI? We should consider having this displayed.

packages/round-manager/src/features/round/ViewRoundStats.tsx
```diff
@@ -161,7 +163,7 @@ export default function ViewRoundStats() {
 }
 
 type StatsCardProps = {
-  text: string | number;
+  text?: string | number;
   title: string;
   grayBorder?: boolean;
```
hmrtn, line 168: I think instead of using a boolean value, we could use a string prop named borderColor to make it more flexible for future use:
```
type StatsCardProps = {
  text?: string | number;
  title: string;
  borderColor?: "gray" | "violet";
};
```

packages/round-manager/src/features/round/ViewRoundStats.tsx
```diff
@@ -114,19 +119,19 @@ export default function ViewRoundStats() {
               </tr>
             </thead>
             <tbody>
-              {Array(10)
-                .fill(null)
-                .map(() => {
+              {matches &&
+                matches.map((match: Match) => {
                   return (
                     <tr>
```
gravityblast, line 125:
```suggestion
                    <tr key={match.applicationId}>
```

packages/round-manager/src/features/round/ViewRoundStats.tsx
```diff
@@ -114,19 +119,19 @@ export default function ViewRoundStats() {
               </tr>
             </thead>
             <tbody>
-              {Array(10)
-                .fill(null)
-                .map(() => {
+              {matches &&
```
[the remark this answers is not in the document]
vacekj: added

packages/round-manager/src/features/round/ViewRoundStats.tsx
vacekj: if we need to style the border colors further, we can pass-through classNames later, rn no requirement to do this so will keep as-is

vacekj (approved):

thelostone-mc (approved):

PR #1523, opened 2023-04-20
1389 application filter

### LINKED ISSUE #1398: hide invalid projects
### Description 

The goal is to ensure we only show projects applications made by the owners on the builder and round-manager

- [x] Update the handleNewProjectApplication https://github.com/allo-protocol/graph/blob/main/round/src/round/implementation.ts#L96 to capture the sender - [added in this commit to the grant repo](https://github.com/allo-protocol/graph/commit/72214cd4d0e35d3d44717127314e85635b0b0547)
- [x] Write a util function on common package in grants-stack which given chainId, registry, projectId and msg.sender -> it returns if msg.sender is owner or not 
- [ ] update our query filter on  round-manager to show real applications on review page
- [ ] update our query filter on  builder to show real applications 

More context: https://discord.com/channels/562828676480237578/1095277042674716772/1095277046114029649

KurtMerbeth:
fixes #1398


packages/common/src/registry.ts
```diff
@@ -0,0 +1,115 @@
...
+    (result: any) => (result.returnValues ? result.returnValues : []),
+  ) : [[]];
+};
+
+export const addressesByChainID = (chainID: number) => {
+  const chainName: string = chains[chainID];
+  return addresses[chainName];
+};
+
+export const chains: { [key: number]: string } = {
```
hmrtn, line 80: is there an enum available for this?

packages/common/src/registry.ts
```diff
@@ -0,0 +1,115 @@
...
+    ProjectRegistryABI,
+    provider,
+  );
+
+  try {
+    const projectOwners = await projectRegistry.getProjectOwners(projectID);
+    return projectOwners as string[];
+  } catch (err) {
+    // console.log(err);
+    console.log(
```
boudra, line 28: shouldn't we just return the error? or at least log it with `console.error`

packages/round-manager/.env.sample
```diff
@@ -12,6 +12,7 @@ REACT_APP_SUBGRAPH_FANTOM_MAINNET_API=################
 # Ethereum Providers
 REACT_APP_INFURA_ID=######################
 REACT_APP_ALCHEMY_ID=######################
+REACT_APP_ALCHEMY_OPTIMISM=######################
```
boudra, line 15:
```suggestion
REACT_APP_ALCHEMY_OPTIMISM_ID=######################
```

packages/common/src/registry.ts
```diff
@@ -0,0 +1,115 @@
...
+    {
+      reference: "projectRegistry",
+      contractAddress: registryAddress,
+      abi: ProjectRegistryABI,
+      calls: [],
+    },
+  ]
+
+  for (const projectID of projectIDs) {
+    contractCallContext[0].calls.push({
```
boudra, line 59: If we want to batch multiple calls in one, JSON-RPC already supports this natively, it allows you to make multiple calls and get their results in one single call, instead of introducing a third party library and contract.

packages/common/src/registry.ts
```diff
@@ -0,0 +1,115 @@
+import ProjectRegistryABI from "./abis/ProjectRegistry.json";
```
boudra, line 1: if this file contains things related to the Project Registry, wouldn't `projectRegistry.ts` be a better name?

packages/common/src/registry.ts
```diff
@@ -0,0 +1,115 @@
...
+    projectRegistry: "0x984749e408FF0446d8ADaf20E293F2F299396631",
+  },
+  fantom: {
+    projectRegistry: "0x8e1bd5da87c14dd8e08f7ecc2abf9d1d558ea174",
+  },
+};
+
+export type DeploymentAddress = {
+  [key: string]: {
+    projectRegistry: string | undefined;
```
boudra, line 113: looks like it might not be needed?
```suggestion
    projectRegistry: string;
```

packages/builder/src/actions/projects.ts
```diff
@@ -439,52 +447,93 @@ export const fetchProjectApplications =
...
+      }
+    })
+  );
+
+  return apps.flat();
+};
+
+export const fetchProjectApplicationInRound = async (
+  applicationId: string,
+  roundID: string,
```
boudra, line 488:
```suggestion
  roundId: string,
```

packages/builder/src/actions/projects.ts
```diff
@@ -439,52 +447,93 @@ export const fetchProjectApplications =
...
+  reactEnv?: any
+): Promise<any> => {
+  const splitApplicationId = applicationId.split(":");
+  const projectChainId = Number(splitApplicationId[0]);
+  const projectRegistryAddress = splitApplicationId[1];
+  const projectNumber = splitApplicationId[2];
+
+  const projectApplicationID = generateUniqueRoundApplicationID(
```
boudra, line 496:
```suggestion
  const projectApplicationId = generateUniqueRoundApplicationId(
```

boudra:
This is some great work @KurtMerbeth, I added some comments for small things, I will continue reviewing & testing later, it's a big one!

packages/common/src/registry.ts
codenamejason: Not yet, we could put this on the list for our refactor/tech-debt sprint and put some enums in the common package.

packages/common/src/registry.ts
codenamejason: I would like to update this but should we do in another PR so this is not held up any longer @boudra ?

packages/builder/src/actions/projects.ts
KurtMerbeth: The problem: we use sometimes Id and sometimes ID across the apps. Changing it at one place results in a change in other files. would keep it for now and maybe add it to our refactor list

packages/common/src/registry.ts
boudra: Which also begs the question that if we're not in a rush to deploy this anymore we should think about using the indexer to get the applications, it should simplify this PR a lot.

PR #1558, opened 2023-04-24
Fix Set Application Statuses

boudra:
##### Description

resolves https://github.com/gitcoinco/grants-stack/issues/1557

Issues:

- Row index is calculated wrong  from application index
- Mutating applications received in `_bulkUpdateGrantApplication`
- Using application index as a column index
- Cancelled status is wrong
- Rows and applications don't need to be sorted

This PR also refactors some small things and renames the variables a bit to make things more readable an easy to follow. I didn't want to change too much to avoid introducing other issues.


gravityblast (approved):

codenamejason (approved):
This looks good, how can we test the whole flow?

hmrtn (commented):
lgtm

PR #1578, opened 2023-04-25
fix: setState inside a component render function in passport connect

bhargavaparoksham:
##### Description

fix: setState inside a component render function in passport connect


thelostone-mc (approved):
NOICE

PR #1667, opened 2023-05-01
update round images to jpg

codenamejason:
This changes the image format from `.png` to `.jpg` for the default round images.


vacekj:
@nfrgosselin this just changes the image format used for people's banners when it's missing. to test: go to main grants page and verify that the default banners are still showing.

PR #1668, opened 2023-05-01
feat: performance improvements

### LINKED ISSUE #1666: Split up svgs from the main bundle
Currently, we are loading the SVG assets for every page load in the main bundle. These need to be lazy loaded, so that they are only loaded when needed. I shouldn't be loading the welcome banner when i'm on the project details page.

vacekj:
fixes #1666 

- split up svgs into their own chunks
- don't flash and download 404 in ViewRound
- change default banner to jpg
also some automated prettier formatting


codenamejason (approved):

packages/grant-explorer/src/assets/default_banner.jpg
boudra, line 1: Not a huge win but just wanted to add that you can further reduce this file a 7% by using https://imageoptim.com/mac

boudra (approved):
🚀

packages/grant-explorer/src/assets/default_banner.jpg
vacekj: Good tip, I think the next step will be WebP which will further reduce it down by about 30%, and is supported in every browser now.

hmrtn (approved):

PR #1672, opened 2023-05-02
feat: new finalize round results page

### LINKED ISSUE #1475: Finalize Round Results
**User Story:**
As a program manager
I want to view and confirm my round results
So that I can finalize and payout my round

**Acceptance Criteria from #1455**
- [x] the table should have fixed height and scrollbars like in the stats page and figma
- [x] implements CSV downloads

**Acceptance Criteria**
- [x] GIVEN that I am on the Round Results page AND my round has ended
WHEN I click on the Finalize Results button
THEN the current distribution in the preview table is queued for upload
AND I see the Warning modal

- [x] GIVEN that I'm looking at the Warning modal
WHEN I click Cancel
THEN the modal closes
AND I see the latest version of the distribution in the preview table

- [x] GIVEN that I'm looking at the Warning modal
WHEN I click Continue
THEN I am guided through the finalization process and wallet transactions

- [x] GIVEN that I've initiated the finalization process
WHEN the process finishes
THEN I see the completion toast
AND I see the finalized version of the Round Results page
AND the round is marked ready for payout

- [x] GIVEN that I've finalized the round
WHEN I click on the "View on Etherscan" button
THEN a new tab opens with the finalization transaction on the relevant block explorer

**Product & Design Links:**
[Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=1086-43069&t=N9E3HUHadbyKhS3U-4)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

vacekj:
fixes #1475


packages/eslint-config-gitcoin/index.js
```diff
@@ -17,21 +21,14 @@ module.exports = {
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-non-null-assertion": "off",
         "@typescript-eslint/no-empty-function": "off",
-        // "@typescript-eslint/no-unused-vars": [
```
boudra, line 20: did you leave this commented out by accident?

boudra (commented):
I have to say all of this reformatting is making reviewing real changes really difficult, I was talking to @gravityblast about adding a formatting check in CI and reformat everything in the same PR so future PRs will never have any reformatting.

vacekj:
We already have a prettier precommit hook. So every staged file that goes in should be formatted. We can do a pass on the whole repo, but after that it should be automatic. Only reason why it wouldn't work is someone commiting without the hook.

boudra:
Yeah that is what I meant, so using prettier in CI to check formatting with `--check` not to reformat and commit (that would be weird 🙃) which is what the eslint plugin does in the end.

Hooks can be skipped so in my opinion it's as if they don't exist at all and to be honest I find them super annoying and they're a security risk, git commit should just commit and every developer should be responsible to run tests, lint and format as they see fit, and put those things in CI as a quality barrier that cannot be skipped, but that might be personal preference.

Also I would rather not check formatting in CI at all and trust the team, but we have no other choice since  I can see the team is not formatting files properly.

gravityblast:
I agree with removing the hooks and having CI failing if the formatting is not correct. Hooks are personal and it's true for security reasons most people will skip them. We are already doing the same in `builder`, and if you configure your editor in the right way following the repo settings it will format it in the right way. Having CI checking and failing is the only way to be sure all the contributions to the repo are cheked.

vacekj:
There's no security issue with hooks, since they are opt-in and you can inspect what they are doing before you first run them. `pnpm install` is a much bigger security hole with post-install scripts being run automatically and with full permissions. I would leave them in, make them not auto-install but they are still useful - much faster feedback that some lint or formatting is wrong than waiting 5 minutes for CI.

packages/eslint-config-gitcoin/index.js
vacekj: Yeah, i'll fix.

boudra:
thank you guys I agree it's a personal thing and they can be useful for some people so they can be left in but not automatically installed, we can do this in another PR though

re security: I think it's a long stretch to say there are no security issues, it's not so black or white, anything that let's other people execute code in your computer is surface area that can be abused in one way or another and I personally want to keep that surface as small as possible, and having some security risk from npm scripts doesn't mean it's OK to add more, especially when it's unnecessary

vacekj:
@nfrgosselin ready for review

packages/eslint-config-gitcoin/index.js
boudra: ah yes I just noticed, weird maybe we're adding this rule somewhere else, I think it's useful to have

boudra:
I noticed we're not using the common UI components like buttons, shouldn't we try to have new code using it instead of having to rewrite later on?

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -261,17 +373,44 @@ export default function ViewRoundResults() {
...
-                <button
-                  onClick={onFinalizeResults}
-                  className="self-end w-fit bg-white hover:bg-pink-200 border border-pink-400 text-pink-400 py-2
+                {!isReadyForPayout && (
+                  <>
+                    <button
+                      onClick={() => {
+                        setWarningModalOpen(true);
+                      }}
+                      className="self-end w-fit bg-white hover:bg-pink-200 border border-pink-400 text-pink-400 py-2
```
boudra, line 382: the button needs to be purple as per the figma

boudra (changes requested):

melissa-neira (changes requested):
A few minor changes:
- The overall spacing on the page does not match the spacing in [Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=1327-41795&t=WkNBsSZIzrx9xbwT-4). There were some dividing lines that were added to better distinguish the sections that should be there. 
- Button sizes are too big, check out the Figma file above for reference
- The "recalculate results" button appears only after a user has selected a file to upload. Also, the copy for that button has changed. [Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=1086-44901&t=WkNBsSZIzrx9xbwT-4)
- The font size for the radio button text in the round saturation text is too big and looks out of place, let's make sure it's the same font size as the rest of the body text (should be 14px)
- Screenshot 2023-05-03 at 18 14 53
The gray text should be right-aligned underneath the "finalize" button. The "finalize" button should also be purple - [Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=1327-41928&t=WkNBsSZIzrx9xbwT-4)

boudra:
@vacekj I'll fix the recalculate results button in my PR: https://github.com/gitcoinco/grants-stack/pull/1687

vacekj:
> @nfrgosselin @vacekj @melissa-neira I'll deal with the styling related to the revise results section in my PR: https://github.com/gitcoinco/grants-stack/pull/1687 and @hmrtn is adding the saturation stuff here: https://github.com/gitcoinco/grants-stack/pull/1693
> 
> 
> 
> I am not sure why this PR has the upload CSV section or the saturation stuff, the task is only to implement the round finalization without any settings, we either pretend those sections don't exist in this PR and test them in their own PRs or we merge the other PRs here and test everything together like I suggested yesterday at standup

Talked about this already a while ago, Hans implemented the whole page, let's test them in their respective PRs to keep separation.

vacekj:
> Loom

Thank you for the loom, looking into it now. it would be very helpful if you could always:
- open the devtools console (Cmd + Option + J) for the whole loom, that way I can have some info of what's going wrong
- paste the address of the round/project/w/e you're testing into the comment

PR #1655, opened 2023-04-28
estimate gas price

KurtMerbeth:
fixes #1652

PR #1673, opened 2023-05-02
remove gas fix

codenamejason:
Removes the previous PR(#1655) for the gas fix since Metamask updated their application.


thelostone-mc (approved):
locally this works.

PR #1691, opened 2023-05-03
Prevent project owners from reapplying

### LINKED ISSUE #1526: Builder: need to prevent project owners from reapplying
**Overview**
Currently, even if I've applied with a project, I am able to apply again with the same project. This should not be allowed, and should've been addressed with https://github.com/gitcoinco/grants-stack/issues/1327

Figma: https://www.figma.com/file/U4uRlyJJn92QiNUuyRxSvX/%5BWIP%5D-Grants-Application-%2F-Edit-Flow?node-id=687-5686&t=CZez4N3cfxLiTsF6-0 

**Steps To Reproduce**
1. Go to a round application 
2. Apply successfully with one project 
3. Go back to the round application and select the same project 
4. You will see that I can continue to the application, when it should've gated me from accessing the application. 

**Observed  behavior**

**Expected behavior**

**Screenshots**

**Additional context**

bhargavaparoksham:
##### Description

Builder: need to prevent project owners from reapplying

##### Refers/Fixes
#1526

### DIFF
--- a/packages/builder/src/utils/graphql.ts
+++ b/packages/builder/src/utils/graphql.ts
@@ -15,21 +15,19 @@ export enum ChainId {
  * @param chainId
  * @returns GraphEndpoint
  */
-const getGraphQLEndpoint = (
-  chainId: number,
-): string | undefined => {
+const getGraphQLEndpoint = (chainId: number): string | undefined => {
   switch (chainId) {
     case ChainId.MAINNET_CHAIN_ID:
       // eslint-disable-next-line max-len
-      return `${process.env.REACT_APP_SUBGRAPH_URL_MAINNET}`;
+      return process.env.REACT_APP_SUBGRAPH_URL_MAINNET;
     case ChainId.GOERLI_CHAIN_ID:
-      return `${process.env.REACT_APP_SUBGRAPH_URL_GOERLI}`;
+      return process.env.REACT_APP_SUBGRAPH_URL_GOERLI;
     case ChainId.OPTIMISM_MAINNET_CHAIN_ID:
-      return `${process.env.REACT_APP_SUBGRAPH_URL_OPTIMISM_MAINNET}`;
+      return process.env.REACT_APP_SUBGRAPH_URL_OPTIMISM_MAINNET;
     case ChainId.FANTOM_MAINNET_CHAIN_ID:
-      return `${process.env.REACT_APP_SUBGRAPH_URL_FANTOM_MAINNET}`;
+      return process.env.REACT_APP_SUBGRAPH_URL_FANTOM_MAINNET;
     case ChainId.FANTOM_TESTNET_CHAIN_ID:
-      return `${process.env.REACT_APP_SUBGRAPH_URL_FANTOM_TESTNET}`;
+      return process.env.REACT_APP_SUBGRAPH_URL_FANTOM_TESTNET;
     default:
       throw new Error(
         `Chain id (${chainId}) is invalid or subgraph is not deployed on requested chain`
@@ -40,7 +38,7 @@ const getGraphQLEndpoint = (
 export const graphqlFetch = async (
   query: string,
   chainId: number,
-  variables: object = {},
+  variables: object = {}
 ) => {
   const endpoint = getGraphQLEndpoint(chainId);
   if (endpoint) {
--- a/packages/builder/src/utils/graphql.ts
+++ b/packages/builder/src/utils/graphql.ts
@@ -17,8 +17,6 @@ export enum ChainId {
  */
 const getGraphQLEndpoint = (
   chainId: number,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  reactEnv?: any // ProcessEnv
 ): string | undefined => {
   switch (chainId) {
     case ChainId.MAINNET_CHAIN_ID:
@@ -43,9 +41,8 @@ export const graphqlFetch = async (
   query: string,
   chainId: number,
   variables: object = {},
-  reactEnv?: any // ProcessEnv
 ) => {
-  const endpoint = getGraphQLEndpoint(chainId, reactEnv);
+  const endpoint = getGraphQLEndpoint(chainId);
   if (endpoint) {
     return fetch(endpoint, {
       method: "POST",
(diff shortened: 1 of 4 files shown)


packages/builder/src/utils/graphql.ts
```diff
@@ -19,21 +17,21 @@ export enum ChainId {
  */
 const getGraphQLEndpoint = (
   chainId: number,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   reactEnv?: any // ProcessEnv
```
boudra, line 21: why don't we just delete the variable here and where it's called?

packages/builder/src/utils/graphql.ts
```diff
@@ -19,21 +17,21 @@ export enum ChainId {
...
 ): string | undefined => {
-  const environment = reactEnv || env;
   switch (chainId) {
     case ChainId.MAINNET_CHAIN_ID:
       // eslint-disable-next-line max-len
-      return environment.REACT_APP_SUBGRAPH_URL_MAINNET;
+      return `${process.env.REACT_APP_SUBGRAPH_URL_MAINNET}`;
```
boudra, line 26: This functions returns `string` or `undefined`, now it always returns a string, the function below this one is checking the return value, which now will not work as expected.

Check here: https://github.com/gitcoinco/grants-stack/blob/b338d74a47edf37b9864dccbe8f74393cf2f05d1/packages/builder/src/utils/graphql.ts#LL49C18-L49C18

boudra (changes requested):

packages/builder/src/utils/graphql.ts
bhargavaparoksham: It was being used in graphqlFetch function as well & I was not sure if removing it will cause issues in other places or lead to failing tests etc, so left it there for now.

michellema1208:
hey @bhargavaparoksham - can you share the deploy link for Builder? I tried looking here (https://github.com/gitcoinco/grants-stack/pull/1691/checks?check_run_id=13233329467) but didn't find it.

thelostone-mc:
@michellema1208 https://bafybeihxgvf6fxep7hpze62e4sli7zpefldm7v5elbnuscootzrokxlphu.on.fleek.co
image

packages/builder/src/utils/graphql.ts
boudra: yes the error  is on purpose, why do you want to skip the error? you should have the environment variables defined properly, if you don't get an error here stuff will just not work instead, making things hard to debug

packages/builder/src/utils/graphql.ts
boudra: the parameter is optional, and now we're not even using it, so not sure what kind of issues removing it will cause, did you check? removing it should be quite easy and will save us a headache in the future, as now we have a parameter in the function signature but it's not use, the caller will not expect this unless they read the code

packages/builder/src/utils/graphql.ts
boudra: but also if you want to remove the error, please remove the error handling logic properly, this will just silence it

bhargavaparoksham: I'm getting the error for goerli where the environment variable is defined, which is the issue. If it had been for an undefined evv, why would I try to change it.

packages/builder/src/utils/graphql.ts
gravityblast: I agree let's remove it, and with a quick search we can check all the calls to the function and remove the param from the calls, since it's not even used here.

packages/builder/src/utils/graphql.ts
gravityblast: @bhargavaparoksham  if you see the error it means the variable is not set. If you change it to return an empty string it means you are going to have an http request to `""`.

Check the `graphqlFetch` function:

```
  const endpoint = getGraphQLEndpoint(chainId, reactEnv);
  if (endpoint) {
    return fetch(endpoint, {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
```

If you `console.log` that env var do what do you have? Which one is it?

gravityblast (dismissed):

packages/builder/src/utils/graphql.ts
boudra: I just tested in main and if you have `REACT_APP_SUBGRAPH_URL_GOERLI` defined it works fine, otherwise it throws that error, I also tested in staging and it works well, there must be something else going on

bhargavaparoksham:
@gravityblast @boudra i've removed the reactEnv & updated getGraphQLEndpoint so if env is undefined it throws the correct errors etc. Please check & let me know if everything looks good. Thanks.

bhargavaparoksham:
@michellema1208 here's the updated link to test, https://bafybeibgbx2gprbpx3nvfyxhym66gavbjtplllolddgdcp7bdoufxv4pm4.on.fleek.co

michellema1208 (commented):
@bhargavaparoksham the overall functionality here is working, great job!! 

I do want to callout that the design doesn't fully match what @willsputra has in Figma: 

From my testing: 
Screen Shot 2023-05-05 at 11 48 27 AM

Figma: 
Screen Shot 2023-05-05 at 11 48 27 AM

Primarily: 
- Font text should be black 
- Different font type 
- Red outline on the project dropdown 
- Disabled preview button 

I don't think these are blocking bugs so I'm not marking these as "Request changes", but would be good for us to match UX as much as possible.

thelostone-mc:
@michellema1208 Could we have this as a follow up ticket ? It looks like we are using multiple alerts everywhere.
We should instead standardize alerts so that we can reuse code as opposed to making custom alert for every page.

michellema1208:
@thelostone-mc - gotcha, that's super fair! I've stubbed out a follow-up ticket here: https://github.com/gitcoinco/grants-stack/issues/1712 

@willsputra two questions for you: 
1. I've tagged you to a comment in this follow-up ticket: https://github.com/gitcoinco/grants-stack/issues/1712 --> would love your thoughts/help here! 
2. Do you have any thoughts about keeping the disabled Preview button from UX POV? From Product POV, I think functionality works either way.

michellema1208 (approved):

PR #1692, opened 2023-05-03
1458 add round saturation control

hmrtn:


packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -36,22 +37,49 @@ const distributionOptions = [
...
-        }
-      };
-      reader.readAsText(file);
-    });
-  }, []);
+  useEffect(() => {
+    if (matches && round) {
+      console.log(round)
+      console.log(matches)
+      const matchTotal = matches.reduce(
```
hmrtn, line 70: Here is a point of confusion: 

when we sum the match total (+= each project match amount) then this should not necessarily add up to the match pool amount - but in every case it seems to. Which will then never enable the radio buttons for recalculations. 

Is this correct?

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -61,14 +89,19 @@ export default function ViewRoundResults() {
 
   const onRecalculateResults = () => {
     // Logic for recalculating results goes here
+    // DEBUG: timeout for testing
```
hmrtn, line 92: Another: 
What should actually happen here? 

Are we making a POST request to `/chains/:chainId/rounds/:roundId/matches`?

hmrtn:
ngl - at a little bit of a loss with the context here. UI is gtg - but the implementation logic for radio button switching (between keep and recalc) and the saturation total does not seem correct. 

This should involve pluralistic, allo, the client and the grants stack itself. I am not entirely sure what needs to be altered here to proceed correctly. Any hand on this would be awesome.

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
gravityblast: yes!

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
gravityblast: do you have a match URL to test the results please?

PR #1706, opened 2023-05-05
feat: saturation controls

### LINKED ISSUE #1458: Add Round Saturation control
**User Story:**
As a program manager
I want to view and confirm my round results
So that I can finalize and payout my round

**Acceptance Criteria**
GIVEN that I am on the Round Results tab
WHEN I look at the Round Saturation section
THEN I see the current round saturation percentage
AND the amounts of how much of the matching fund will be distributed

GIVEN that I am on the Round Results tab
WHEN the current matching distribution shows the round is fully saturated (100%)
THEN the saturation radio buttons are grayed out

GIVEN that I am on the Round Results tab
WHEN the current matching distribution shows the round isn't fully saturated (<100%)
THEN the saturation radio buttons are active
AND the "Keep distribution..." button is selected

GIVEN that the current matching distribution isn't fully saturated
WHEN I click the "scale up..." button
THEN the current matching distribution is recalculated to redistribute the whole pool
AND the Revised Matching distribution is shown (with original / revised matching amounts)

**Product & Design Links:**
[Figma](https://www.figma.com/file/V113F2Pp5m2qDM5odXjfFu/Manager-S17?node-id=1086-46100&t=N9E3HUHadbyKhS3U-4)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

hmrtn:
fixes #1458 

Saturation controls.

This depends on this allo indexer client commit / PR: https://github.com/gitcoinco/allo-indexer-client/pull/21

### DIFF
--- a/packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
+++ b/packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
@@ -127,19 +127,17 @@ export default function ViewRoundResults() {
     undefined
   );
   const [overrideSaturation, setOverrideSaturation] = useState<boolean>(false);
-  const overridesJSON = JSON.stringify({
-    overrideSaturation,
-    overridesFile,
-  });
-  const overridesBlob = new Blob([overridesJSON], { type: 'application/json' });
-
   const {
     matches,
     isRevised: areMatchingFundsRevised,
     error: matchingFundsError,
     isLoading: isLoadingMatchingFunds,
     mutate: mutateMatchingFunds,
-  } = useRevisedMatchingFunds(roundId, overridesFile); // TODO: This should be the overrides JSON blob of both saturation and overrides file
+  } = useRevisedMatchingFunds(
+    roundId,
+    { query: { ignoreSaturation: overrideSaturation } },
+    overridesFile
+  );
   const debugModeEnabled = useDebugMode();
 
   const { data: round, isLoading: isLoadingRound } = useRound(roundId);
@@ -170,20 +168,18 @@ export default function ViewRoundResults() {
   useEffect(() => {
     if (round && matches) {
       const sumTotalMatch = matches?.reduce(
-        (acc, match) => acc + Number(match.matchedUSD),
+        (acc: number, match: Match) => acc + Number(match.matchedUSD),
         0
       );
       setSumTotalMatch(sumTotalMatch);
       setRoundSaturation(sumTotalMatch / round.matchAmountUSD);
+      setOverrideSaturation(distributionOption === "keep");
     }
-  }, [round, matches]);
-
+  }, [round, matches, distributionOption]);
 
   useEffect(() => {
-    // when distribution option is keep -> ignoreSaturation=true else false
-    setOverrideSaturation(distributionOption === "keep");
-  }, [distributionOption]);
-
+    mutateMatchingFunds();
+  }, [distributionOption, mutateMatchingFunds]);
 
   const isBeforeRoundEndDate = round && new Date() < round.roundEndTime;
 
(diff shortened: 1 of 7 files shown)


packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -48,21 +48,29 @@ export default function ViewRoundResults() {
...
   const [overridesFile, setOverridesFile] = useState<undefined | File>(
     undefined
   );
-  const matchingTableRef = useRef<HTMLDivElement>(null);
+  const [overrideSaturation, setOverrideSaturation] = useState<boolean>(false);
+  const overridesJSON = JSON.stringify({
+    overrideSaturation,
+    overridesFile,
+  });
+  const overridesBlob = new Blob([overridesJSON], { type: 'application/json' });
```
hmrtn, line 63: This blob should be used with the useRoundMatchingFunds hook, but fails.

nfrgosselin (changes requested):
Looks like the data source might be off a bit (not sure how the price of Goerli ETH is calculated). Also it doesn't look like the amounts are updating to fully distribute the round. See [Loom](https://www.loom.com/share/941d144d7ad44c6ebb549e1d3e325dd3)

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -153,6 +163,36 @@ export default function ViewRoundResults() {
...
   >("keep");
+  const [roundSaturation, setRoundSaturation] = useState<number>(0);
+  const [sumTotalMatch, setSumTotalMatch] = useState<number>(0);
+
+  const mutateMatchingFundsCallback = useCallback(mutateMatchingFunds, [
+    mutateMatchingFunds,
+  ]);
+
+  useEffect(() => {
+    mutateMatchingFundsCallback();
```
boudra, line 174: is the mutate here needed?

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -153,6 +163,36 @@ export default function ViewRoundResults() {
...
+    if (round && matches) {
+      const sumTotalMatch = matches?.reduce(
+        (acc: number, match) =>
+          acc +
+          (areMatchingFundsRevised
+            ? Number(match.revisedMatch) / 10 ** 18
```
boudra, line 180: revisedMatch is not guaranteed to fit into a number

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -153,6 +163,36 @@ export default function ViewRoundResults() {
...
+            : match.matchedUSD),
+        0
+      );
+      setSumTotalMatch(sumTotalMatch);
```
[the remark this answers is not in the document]
boudra: `useMemo` will be called when matches are updated to recalculate the sum so no need for a useEffect + setState

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
boudra: we should use token amounts exclusively, you can use decimal places to represent decimal percentages

nfrgosselin (approved):

melissa-neira (approved):

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -115,46 +130,66 @@ function useRevisedMatchingFunds(roundId: string, overridesFile?: File) {
...
+      4
+    )} ${matchToken?.name}`;
+  }
+
   const isBeforeRoundEndDate = round && new Date() < round.roundEndTime;
 
+  const sumOfMatches = useMemo(() => {
+    return (
+      matches?.reduce(
+        (acc: bigint, match) => acc + (match.revisedMatch ?? BigInt(0)),
```
boudra, line 187: revised match should always be a bigint, if it's undefined we have a bigger problem here, we should investigate

hmrtn: while this is valid - I am choosing violence and ignoring it because we need to role this out

PR #1708, opened 2023-05-05
fix: use bignumbers in cart

### LINKED ISSUE #1699: Explorer: Rounding error causes failed transactions
If I want to donate 0.0015 ETH to a project, the checkout amount will be rounded up to 0.002 ETH. At contract level we check that the vote amount and the sent amount are exactly the same. Since there are discrepancies due to rounding, the transaction fails.

Proposed solution:
Behind the interface, wei should always be used as unit.
Maybe we can also limit the decimals in the form.

image

vacekj:
fixes #1699


boudra (approved):

thelostone-mc (approved):
haven't tested but LGTM

PR #1711, opened 2023-05-05
fix: ensure show 404 page if on wrong network

### LINKED ISSUE #1534: Manager: Wrong Chain injected into "View on Explorer" URL
**Overview**
The URL should pull the chain that the round is deployed on, not the round that the person's wallet is connected to. See [Loom](https://www.loom.com/share/8ee72ef52d4e4747a987b326070188ce)

**Details**
Say more

thelostone-mc:
fixes #1534

### DIFF
--- a/packages/round-manager/src/context/round/RoundContext.tsx
+++ b/packages/round-manager/src/context/round/RoundContext.tsx
@@ -159,7 +159,8 @@ export const useRoundById = (roundId?: string) => {
   useEffect(() => {
     if (roundId) {
       const existingRound = context.state.data.find(
-        (round) => round.id === roundId
+        (round) =>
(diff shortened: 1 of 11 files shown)


michellema1208 (changes requested):
hey @thelostone-mc - just tested this as well, and it looks like the 404 page is working when on the wrong network, but the page is not updating when the user switches to the right network. I believe even after I've changed my network to the correct chain the round is deployed on, the URL doesn't update to the correct chain. 

Loom: https://www.loom.com/share/a58ac508a4b94988936b2964aa29de6c

KurtMerbeth (approved):

michellema1208 (approved):

PR #1757, opened 2023-05-09
Fix saturation

boudra:
##### Description

- Fix infinite lifecycle
- Simplify lifecycle
- Use match amounts instead of USD to calculate saturation

I didn't have time to fix the tests but i am sure they'r easy to fix, I would merge this PR and fix in the origin


boudra:
I sneaked in a commit that fixes the join between revised and original matches, there's an edge case where if a project is deleted when revised it won't be shown in the table

gravityblast (approved):

PR #1769, opened 2023-05-10
feat: separate ipfs and contract upload steps during finalization

### LINKED ISSUE #1764: Finalize process modal should have two steps
**Details**
When finalizing a round on the Round Results tab, the user has to sign two transactions if they have revised the distribution (one to upload revised distribution, one to finalize the round). The progress modal only shows one step, but we need to have a second step to account for this.

vacekj:
modifies the modal so that there is a new step.

Upload distribution to IPFS -> Set hash in Contract -> Set as Ready for Payout -> Final step (don't what what it's for but it's in the designs and all the other modals)

the code isn't ideal, but the whole context and handling of finalizing to contract needs to be redone anyway, so this is temporary.

fixes #1764

### DIFF
--- a/packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
+++ b/packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
@@ -1,11 +1,14 @@
-import { useCallback, useState, useRef, useMemo } from "react";
+import { useCallback, useMemo, useRef, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 import { BigNumber, utils } from "ethers";
 import { RadioGroup, Tab } from "@headlessui/react";
 import { ExclamationCircleIcon as NoInformationIcon } from "@heroicons/react/outline";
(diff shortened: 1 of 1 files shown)


nfrgosselin (approved):

melissa-neira (approved):

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -275,10 +289,17 @@ export default function ViewRoundResults() {
...
     Number(
-      ((sumOfMatches * BigInt(10000)) / round.matchAmount) * BigInt(10000)
-    ) / 1000000;
+      ((sumOfMatches * BigInt(10_000)) / round.matchAmount) * BigInt(10_000)
+    ) / 1_000_000;
+
+  const disableRoundSaturationControls = Math.round(roundSaturation) >= 100;
 
-  const disableRoundSaturationControls = roundSaturation >= 100;
+  console.log(
```
gravityblast, line 297: @vacekj  I think we can remove this log right?

gravityblast (approved):

PR #1778, opened 2023-05-10
Fund Grantees Fixes

boudra:
##### Description

I noticed we're using the match percentage in the Fund grantees page: https://github.com/gitcoinco/grants-stack/blob/0e74007cff1b29d130c160dfe402ece47c758a48/packages/round-manager/src/features/round/ViewFundGrantees.tsx#L386

I would rather calculate it using the round Match amount but this should suffice for now.

This also fixes the USD conversion.


hmrtn (dismissed):
🔥

hmrtn (approved):

gravityblast (approved):

PR #1781, opened 2023-05-11
feat: add bread crumbs :bread:

hmrtn:
also removes back buttons outside of design spec

Screenshot 2023-05-10 at 5 01 25 PM
Screenshot 2023-05-10 at 5 01 19 PM


thelostone-mc (approved):

melissa-neira (changes requested):
my bad - can we pls make sure the explorer logo leads users back to the round landing page they were browsing on and not the round discovery page? missed that before! 😅

nfrgosselin (approved):

PR #1788, opened 2023-05-12
Edit flow group transactions

KurtMerbeth:


codenamejason (approved):
Looks good, we just have to add tests and fix the requirements section. We can do this from the integration branch directly. This addresses/adds multicall for our edits.

PR #1790, opened 2023-05-13
enable rolling rounds

owocki:
This is a change to the grants stack round application that will enable applications to a round AFTER the crowdfund portion of the round has started.

This change will enable round operators to run rolling rounds (eg rounds where applications can be accepted after the crowdfund has started).  If this was available to me when we were running the metacrisis round, then we would not have had as many grant owners who missed the round because they didnt realize the deadline until too late.  I also ran into a similar situation with the Metacamp 2023 round wherein ppl asked to be included AFTER the round had already started.

I asked nate about this and he said "This design decision really stemmed from a desire to make the rounds legitimate / fair — we thought that allowing late acceptances would disadvantage those late grantees,".  IMO that is a good consideration, and perhaps there is an opportunity to educate round operators here about the tradeoffs of rolling rounds.  But I dont think a paternalistic "dont do this, its entirely disabled" is the answer.  So thats why I'm opening this PR.

Up to Gitcoin whether they want to merge this or not.  I would very much like to be able to run rolling rounds though.

### DIFF
--- a/packages/round-manager/src/features/round/RoundDetailForm.tsx
+++ b/packages/round-manager/src/features/round/RoundDetailForm.tsx
@@ -388,7 +388,7 @@ export function RoundDetailForm(props: RoundDetailFormProps) {
                             className:
                               "block w-full border-0 p-0 text-gray-900 placeholder-grey-400 focus:ring-0 text-sm",
                           }}
-                          isValidDate={disableBeforeApplicationEndDate}
+                          isValidDate={disableBeforeApplicationStartDate}
                           utc={true}
                           dateFormat={"YYYY-MM-DD"}
                           timeFormat={"HH:mm UTC"}


packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -388,7 +388,7 @@ export function RoundDetailForm(props: RoundDetailFormProps) {
                             className:
                               "block w-full border-0 p-0 text-gray-900 placeholder-grey-400 focus:ring-0 text-sm",
                           }}
-                          isValidDate={disableBeforeApplicationEndDate}
+                          isValidDate={disableBeforeApplicationStartDate}
```
owocki, line 391: allo smart contract validation allows this case if im reading the code correctly https://github.com/allo-protocol/contracts/blob/main/contracts/round/RoundImplementation.sol#L223-L242

thelostone-mc: I am supportive of this feature but this enables people to re-apply their application and this can lead to complications that the product is yet to support :

- notifying round operator an approved application has reapplied
- auto declin-ing duplicate applications made by the same project on approving an application
- handling if an application gets changes mid-round -> how we'd ensure old / new votes get counted for calculations 

^ Would love to reach a path forward on this front before merging this

Defer to @meg-fsc + @michellema1208 on decision here

michellema1208: hey @thelostone-mc ! Thanks for looping me in on this, and calling out the edge cases above! 

Since we've made an interim decision to not allow grantees to reapply to any given round, we can allow for rolling rounds per @owocki 's request without triggering the edge cases above. However, it does bring up some existential questions around if/how we want to enable grantees to reapply in the future that Product team still needs to research and validate. 

FYI @meg-fsc and I have **enabling rolling rounds** in our top 5 pieces of feedback to address post-Beta round, but are still firming up what the exact timeline and roadmap will be. Will sync up with her on this and report back!

meg-fsc: Hey all -- few thoughts
`1` thanks @owocki, love to see community PRs!
`2` as @michellema1208 mentioned, accepting applications after a round has started is on our list of things we want to address
`3` we've been thinking a lot about the role of rounds as we work on direct grants/etc, and i'm not sure rolling rounds (and opening up the reapplication possibility) is the way to go about this

@owocki if you're open to taking a diff approach with this work would love to chat -- happy to open a discord message, you know where to find me!

thelostone-mc (approved):

packages/round-manager/src/features/round/RoundDetailForm.tsx
vacekj: Progress tracked here: https://discord.com/channels/562828676480237578/1116454693481222286

PR #1801, opened 2023-05-15
feat: unify footer

thelostone-mc:
##### Testing

fixes https://github.com/gitcoinco/grants-stack/issues/1611


michellema1208 (changes requested):
hey @thelostone-mc ! 

I just tested this, and recorded a Loom here: https://www.loom.com/share/e56ef975eb6b4377b91d8806bd7db72b 

Overall, I think what I'm seeing is: 
- [ ] Footer is missing from the "View All" page for rounds with open applications 
- [ ] Bottom padding inconsistency across Explorer, Builder, Manager (+ @melissa-neira to weigh in from Design side) 
- [ ] I would recommend some more right padding so it doesn't look like the Intercom widget is covering the footer (+ @melissa-neira to weigh in from Design side) 

I think the apps included in the footer are right from product POV, so no changes there!

melissa-neira (dismissed):
@michellema1208 For some reason on my computer, I'm seeing the correct horizontal padding of 80px and the Intercom widget isn't covering/close to any of the footer icons. I'm on a 14" MacBook pro so maybe that's why I'm seeing it a bit wider? Regardless, I'd say the more responsive we can make the footer the better!

michellema1208:
oooh interesting - I'm also on a 14" Macbook Air, that's strange. +1 to overall responsiveness! 

@melissa-neira are you seeing the bottom padding issue on Explorer & Builder when you test? Curious what the right padding should be (or if it's just happening on my side as well)

melissa-neira:
@michellema1208 No, I'm not seeing the padding issue at all, which is strange! The padding should be 80px on the right - from what I can see in testing, the intercom icon fits right into that negative space and there's enough room fo the footer icons. This could also be a browser thing too? Not sure 😅

thelostone-mc:
Builder
image

Explorer
image

Manager
image

PR #1803, opened 2023-05-15
feat: make url checking in manager allow urls without protocols

### LINKED ISSUE #1538: Manager: Friendlier link verification for URL during round creation
See [Loom](https://www.loom.com/share/408a73025bd64567b69cdf0facdb8ea5) for context

**User Story:**
As a
I want to
So that

**Acceptance Criteria**
GIVEN
WHEN
THEN

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

vacekj:
fixes #1538 

- [ ] Fix test and add new ones for the regex
- [ ] Remove ftp from regex

### DIFF
--- a/packages/round-manager/src/features/round/RoundDetailForm.tsx
+++ b/packages/round-manager/src/features/round/RoundDetailForm.tsx
@@ -60,7 +60,7 @@ export const RoundValidationSchema = yup.object().shape({
             .string()
             /*Matches www.example.com, example.com, http and https prefixes, but not www.invalid */
             .matches(
-              /^(http:\/\/|https:\/\/|ipfs:\/\/)\S+\.\S+$|^(ipfs:\/\/)\S+$/,
+              /^(http:\/\/|https:\/\/|ipfs:\/\/)?\S+\.\S+$|^(ipfs:\/\/)\S+$/,
               "Must be a valid URL"
             )
             .required("You must provide a valid URL."),
--- a/packages/round-manager/src/features/round/RoundDetailForm.tsx
+++ b/packages/round-manager/src/features/round/RoundDetailForm.tsx
@@ -60,7 +60,7 @@ export const RoundValidationSchema = yup.object().shape({
             .string()
             /*Matches www.example.com, example.com, http and https prefixes, but not www.invalid */
             .matches(
-              /((http|https):\/\/)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@\-/]))?/,
+              /^(http:\/\/|https:\/\/|ipfs:\/\/)\S+\.\S+$|^(ipfs:\/\/)\S+$/,
               "Must be a valid URL"
             )
             .required("You must provide a valid URL."),
--- a/packages/round-manager/src/features/round/RoundDetailForm.tsx
+++ b/packages/round-manager/src/features/round/RoundDetailForm.tsx
@@ -56,7 +56,14 @@ export const RoundValidationSchema = yup.object().shape({
         })
         .when("type", {
           is: (val: string) => val && val != "Email",
-          then: yup.string().url().required("You must provide a valid URL."),
+          then: yup
+            .string()
+            /*Matches www.example.com, example.com, http and https prefixes, but not www.invalid */
+            .matches(
+              /^((ftp|http|https):\/\/)?(www.)?(?!.*(ftp|http|https|www.))[a-zA-Z0-9_-]+(\.[a-zA-Z]+)+((\/)[\w#]+)*(\/\w+\?[a-zA-Z0-9_]+=\w+(&[a-zA-Z0-9_]+=\w+)*)?$/,
+              "Must be a valid URL"
+            )
+            .required("You must provide a valid URL."),
         }),
     }),
   }),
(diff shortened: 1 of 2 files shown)


packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -57,7 +57,14 @@ const ValidationSchema = yup.object().shape({
         })
         .when("type", {
           is: (val: string) => val && val != "Email",
-          then: yup.string().url().required("You must provide a valid URL."),
+          then: yup
+            .string()
+            /*Matches www.example.com, example.com, http and https prefixes, but not www.invalid */
+            .matches(
+              /^((ftp|http|https):\/\/)?(www.)?(?!.*(ftp|http|https|www.))[a-zA-Z0-9_-]+(\.[a-zA-Z]+)+((\/)[\w#]+)*(\/\w+\?[a-zA-Z0-9_]+=\w+(&[a-zA-Z0-9_]+=\w+)*)?$/,
```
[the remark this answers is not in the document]
vacekj: No sir, feel free to delet 🫡

hmrtn:
hmmm. i think we might want to reconsider using regex. URLs are pretty diverse and this regex pattern could run into a few issues. More importantly the current regex doesn't guard against XSS attacks that could be embedded within URLs, and could be a serious security risk. 
imho, at the very least, we should use js builtin URL parser as it is well defined and very flexible - along with some regex checking (for xss scripts, typos, etc). 

But even more of imho - i think we should absolutely enforce strict url matching in general and use yup - regardless of the friction this causes the creator. There are just too many security risks to make this flexible - especially when users are expected to click on it.

gravityblast:
@hmrtn I don't see XSS problems here unless we execute or render that value somewhere in the UI. 
As long as we just escape any possible tag when we render the field in the UI, this is just a link that at worst can be broken. 
Do you have a specific example in mind that you thought about?

vacekj:
@hmrtn Good callout on the XSS, luckily React escapes HTML automatically, so we are safe in this case (unless we use this url value in some `dangerouslySetInnerHTML` which we don't and have no reason to). We can add the protocol prefix automatically if the user doesn't type/paste it, to always have well-formed URLs, otherwise I don't see a reason we can't use this regex. It handles all tlds correctly as far as my testing goes.

vacekj:
Last commit adds automatically prepending `https://` to links without a protocol. Discussed possible edge cases with Hans:
- if a website doesn't support https -> bad security practice thus wontfix
- if a url starts with ipfs:// -> we support this and don't prepend https
- we don't support other protocols

melissa-neira:
hmm this isn't working for me - anyone else having the same issue? [Loom](https://www.loom.com/share/ea3dead3053247a9a2be6a93505d7ae5)

vacekj:
@melissa-neira please recheck

melissa-neira (commented):
It works for me now! thank you :) one little thing - it seems like the input box isn't aligned anymore with the component on the right.
Screenshot 2023-05-24 at 16 42 38

nfrgosselin (changes requested):
This functionality works for me — agree with @melissa-neira that we need to adjust formatting before merging

michellema1208:
@vacekj - just tested this! Two things to callout

- [ ] Similar to @melissa-neira - I'm seeing the misaligned input boxes (see screenshot below). Since this isn't happening on staging, let's make sure we address it in this PR before merging 
Screen Shot 2023-05-30 at 3 33 20 PM

- [ ] This isn't a bug per se, but I notice that if I enter a website URL like "gitcoin.co" and fill out other required info, click "Next", and then click "Back", then I see that "http://gitcoin.co" is my new input. Not sure if we want it to appear in the UI? + @melissa-neira for a second opinion & @gravityblast for any technical edge cases that could emerge (cc: @nfrgosselin who drafted OG ticket)

Loom: https://www.loom.com/share/6aa81a0f18df4e3cbefdc94ccc0c8302 

Other than those two UX nitpicks, this looks great and functionality-wise is working for me!

vacekj:
@michellema1208 spacing should be fixed now, re: second point - I think it's best that what the user sees = what gets submitted. But I can also make it so that it only gets added at the last step.

michellema1208 (approved):
@vacekj totally, I think your solution works great! LGTM [🚀](https://emojipedia.org/rocket/)[🚀](https://emojipedia.org/rocket/)[🚀](https://emojipedia.org/rocket/)

vacekj:
@nfrgosselin need you to approve please, @hmrtn @boudra @gravityblast eng review please 🙏🏻

michellema1208 (dismissed):

packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -56,7 +56,14 @@ export const RoundValidationSchema = yup.object().shape({
         })
         .when("type", {
           is: (val: string) => val && val != "Email",
-          then: yup.string().url().required("You must provide a valid URL."),
+          then: yup
+            .string()
+            /*Matches www.example.com, example.com, http and https prefixes, but not www.invalid */
+            .matches(
+              /^(http:\/\/|https:\/\/|ipfs:\/\/)?\S+\.\S+$|^(ipfs:\/\/)\S+$/,
```
DanieleSalatti, line 63: This does not allow for URLs not ending after the TLD? `https://blabla.com/ciao` would not pass if I'm reading this correctly. Should we allow paths in the URL?

DanieleSalatti (approved):
Left one comment, if we are fine with that then LGTM!

packages/round-manager/src/features/round/RoundDetailForm.tsx
vacekj: According to regex101, paths in the url are fine.

Issue #1540, opened 2023-04-20
Manager: Friendlier date / time picker during round creation


nfrgosselin:
See [Loom](https://www.loom.com/share/b317a184891c4dad8d2c611d02f2a663) for context

**User Story:**
As a
I want to
So that

**Acceptance Criteria**
GIVEN
WHEN
THEN

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

PR #1755, opened 2023-05-09
Edit flow integration

codenamejason:
Closes #1646 #1701

PR #1805, opened 2023-05-16
feat: fixing up date picker

hmrtn:
This PR addresses #1540 and uses a new date time picker component. Specifically, it utilizes the browsers native datetime picker.
The control's UI varies in general from browser to browser and is not exactly customizable. But it is well defined. In browsers with no support (like old ie), these degrade gracefully to simple text input. 
Screenshot 2023-05-17 at 11 53 07 AM

One of the consequences of this is the loss of input UTC time. Instead, it uses the clients local. But this is handled properly throughout the application - so it should not be a problem, and imo improves the usability. 

After creating the component I took the opportunity of the "down time" to commit a much needed refactor of the application form. This involved adding a couple of new base components, pulling out the form sub components into its own file, and pulled out the validation schema into its own file. It is now sane to navigate and modify ;)

### DIFF
--- a/packages/round-manager/src/features/common/FormInputField.tsx
+++ b/packages/round-manager/src/features/common/FormInputField.tsx
@@ -1,0 +1,36 @@
+import { FC } from "react";
+import { UseFormRegisterReturn, FieldErrors } from "react-hook-form";
+import { Input } from "common/src/styles";
+import { Round } from "../api/types";
+import _ from "lodash";
+
+export const FormInputField: FC<{
+  register: UseFormRegisterReturn<string>;
+  errors: FieldErrors<Round>;
+  label: string;
+  id: string;
+  placeholder?: string;
+}> = ({ register, errors, label, id, placeholder }) => {
+  const errorMessage = _.get(errors, id)?.message;
+  const hasError = Boolean(errorMessage);
+
+  return (
+    <div className="col-span-6 sm:col-span-3">
+      <div className="flex justify-between">
+        <label htmlFor={id} className="text-sm">{label}</label>
+        <span className="text-right text-violet-400 float-right text-xs mt-1">*Required</span>
+      </div>
+      <Input
+        {...register}
+        className={"h-10"}
+        $hasError={hasError}
+        type="text"
+        id={id}
+        placeholder={placeholder}
+      />
+      {hasError && (
+        <p className="text-xs text-pink-500">{errorMessage}</p>
+      )}
+    </div>
+  );
+};
--- a/packages/round-manager/src/features/common/FormInputField.tsx
+++ b/packages/round-manager/src/features/common/FormInputField.tsx
@@ -27,6 +27,7 @@ export const FormInputField: FC<{
         type="text"
         id={id}
         placeholder={placeholder}
+        data-testid={`${id}-testid`}
       />
       {hasError && (
         <p className="text-xs text-pink-500">{errorMessage}</p>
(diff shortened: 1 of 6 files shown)


DanieleSalatti:
`react-date-range` would be kinda cool too, but it would require a slight rethinking of the UX

vacekj:
Let's disable the lint overrides and fix up the `any`s otherwise LGTM! Looks great thanks for taking the time to clean up this part of the codebase 🙌

hmrtn:
> Let's disable the lint overrides and fix up the `any`s otherwise LGTM! Looks great thanks for taking the time to clean up this part of the codebase 🙌

Unfortunately I think many of the any's are a consequence of our form register usage. I did not see any clear way to resolve this without looking more deeply into it - because I am not exactly keen on anys!

codenamejason:
@hmrtn do we want to add the updated date-pickers to the round edit flow in round settings?

vacekj:
Let's use @ts-expect-error when we can't solve a type error, better than typing as any and disabling type checking altogether. As for the other type errors, let's pair on fixing them.

hmrtn:
> @michellema1208 @hmrtn do we want to add the updated date-pickers to the round edit flow in round settings? We are going to have loads of conflicts here #1755

From discord: 
okay after a lengthy discussion with <@346985800145043467> and <@132595722649468928> here is the gist of the issue and plan to resolve it: 
- https://github.com/gitcoinco/grants-stack/pull/1755 is the edit flow PR. there will be a lot conflicts with https://github.com/gitcoinco/grants-stack/pull/1805, as this includes a refactor and new base components. To temporarily resolve the conflicts, we are going to create a new branch based on 1755 (or main when merged) that include hardcoded copies of the component dependencies introduced in the edit-flow (there are only a couple). Once this PR is merged, we can then proceed with merging 1805 on top, resolving the conflicting changes, as they are now entirely independent. After this process is complete, then we can at some point go back and use the 1805 refactored components throughout the application.

packages/round-manager/src/features/round/ApplicationFormValidationSchema.tsx
```diff
@@ -0,0 +1,71 @@
+import * as yup from "yup";
```
boudra, line 1: this file is not a react component, the name needs to be camel case and extension `.ts`

packages/round-manager/src/features/round/ApplicationFormComponents.tsx
```diff
@@ -0,0 +1,226 @@
...
+interface BetterDatetimeProps {
+  control: Control<Round>;
+  name: any;
+  label: string;
+  date?: moment.Moment;
+  setDate?: (date: moment.Moment) => void;  
+  minDate?: moment.Moment;
+}
+
+export const BetterDatetime: FC<BetterDatetimeProps> = ({ control, name, label, date, setDate, minDate }) => {
```
boudra, line 194: shouldn't this be named just `Datetime`, what do you think?

packages/round-manager/src/features/common/FormInputField.tsx
```diff
@@ -0,0 +1,37 @@
+import { FC } from "react";
+import { UseFormRegisterReturn, FieldErrors } from "react-hook-form";
+import { Input } from "common/src/styles";
+import { Round } from "../api/types";
+import _ from "lodash";
```
boudra, line 5: since we're only using a few functions form lodash here we should import them individually to avoid having a large bundle size: https://www.blazemeter.com/blog/import-lodash-libraries

boudra (changes requested):

hmrtn:
All, this branch closed in favor of this branch: https://github.com/gitcoinco/grants-stack/tree/1540-better-forms

PR #1848, opened 2023-05-22
1797 upgrade to wagmi v1 + viem 133 = 27 kb

### LINKED ISSUE #1797: upgrade to wagmi v1 + viem, (133 => 27 kB)
https://wagmi.sh/react/migration-guide
https://www.rainbowkit.com/docs/migration-guide
https://viem.sh/docs/ethers-migration.html

**Overview**
- [x] Upgrade from wagmi v0 to v1 (this drops ethers for viem in the meantime)
- [x] Remove ethers dependency
- [x] Replace all BigNumber usage with bigint
- [x] bump es version to bigint literals
- [x] upgrade rainbowkit to v1
- [x] get typescript to pass
- [x] get build to pass
- [x] get app working
- [x] get tests to pass
- [x] verify everything that is put on-chain
- [x] verify calculations
- [x] Replace all old ABIs with new ones, delete unused

**Projects**
- [x] Round Manager

vacekj:
fixes #1797 

# Docs
[wagmi.sh/react/migration-guide](https://wagmi.sh/react/migration-guide)
[rainbowkit.com/docs/migration-guide](https://www.rainbowkit.com/docs/migration-guide)
[viem.sh/docs/ethers-migration.html](https://viem.sh/docs/ethers-migration.html)

# Objectives
- Take advantage of viem's static type checking to ensure data is passed to and from the chain in the correct format, to prevent typos and radically increase developer productivity when interacting with the chain.
- Drop ethers dependency and as a result reduce bundle size significantly (-100 kB)
- Don't rely on ethers' abstraction and instead come closer to the JSON-RPC underlying both viem and ethers
- Be able to upgrade Rainbowkit and Wagmi to their v1 stable releases and suppor their future versions
- Take advantage of doing a pass of the codebase to reduce tech debt and cleanup where it doesn't affect functionality or other PRs, including resolving the mess around signer, provider, signerOrProvider etc.

# Overview of changes
- `ethers.constants.AddressZero` -> `zeroAddress`
- `provider` -> `publicClient`
- `signer` -> `walletClient`
- `new ethers.Contract` -> `getContract`
- use native bigint instead of ethers BigNumber everywhere
- ABIs have been updated to the latest versions and now use the object format to enable strict typing
- Redux Router removed as unused
- standalone jest is now used for testing, due to jest version in craco being too old and not supporting `exports` in modules, and the jest version being pinned inside craco thus us not being able to update it as a peerDep
- the new jest config is mostly an adaptation of the one used internally by react-scripts
- use builtin fantom testnet chain from `viem/chains` instead of declaring our own
- bump ES version target of esbuild to es2020 to support bigint literals
- bump ES version in tsconfig
- update tests to mock wagmi and viem correctly
- clean up mocking in tests a little bit
- resolve eslint ignores by explicitly type casting roundIds to string

### DIFF
--- a/packages/round-manager/src/features/common/Option.tsx
+++ b/packages/round-manager/src/features/common/Option.tsx
@@ -1,19 +1,15 @@
 import { XIcon } from "@heroicons/react/solid";
-import { SchemaQuestion } from "../api/utils";
 
-// eslint-disable-next-line
 function Option({
   index,
   value,
   onChange,
   onDeleteOption,
-  options,
 }: {
   index: number;
   value: string;
-  onChange: (event: any) => void;
+  onChange: (event: React.ChangeEvent) => void;
   onDeleteOption: (index: number) => void;
-  options: SchemaQuestion[];
 }) {
   const renderDeleteOption = (index: number) => {
     return (


boudra:
@vacekj you have a conflict - also the build also doesn't seem to build, do you know why?

melissa-neira:
defer to eng review 🕺

vacekj:
@boudra RM build is fine, other projects are not touched by this PR. RM build from fleek is deployed here: https://bafybeig2o3chwjt6yucmlzzp6sqy2gwxkjwqvvngwjrzobx2geucbaqdbi.on.fleek.co/.
Conflict resolved.

boudra:
> @boudra RM build is fine, other projects are not touched by this PR. RM build from fleek is deployed here: https://bafybeig2o3chwjt6yucmlzzp6sqy2gwxkjwqvvngwjrzobx2geucbaqdbi.on.fleek.co/. Conflict resolved.

@vacekj the build is failing no? https://github.com/gitcoinco/grants-stack/actions/runs/5135107742/jobs/9240073731?pr=1848

packages/grant-explorer/package.json
```diff
@@ -131,7 +131,7 @@
     "tailwindcss": "^3.0.24"
   },
   "engines": {
-    "node": ">=16.15.0",
+    "node": ">=20",
```
boudra, line 134: I see the current Node LTS version is 18, why is it necessary to upgrade?

packages/round-manager/package.json
```diff
@@ -69,39 +64,26 @@
...
-    }
-  },
   "browserslist": {
     "production": [
-      "chrome >= 67",
-      "edge >= 79",
-      "firefox >= 68",
-      "opera >= 54",
-      "safari >= 14"
+      "last 1 chrome version",
```
boudra, line 84: is this safe to do? the `defaults` option seems like a more sensible setting for production, what do you think?

boudra (commented):
We need to go through the whole flow of creating, editing, approving applications and finalizing a round before merging this, did you have a chance to test everything @vacekj ?

packages/round-manager/src/context/program/__tests__/CreateProgramContext.test.tsx
```diff
@@ -87,7 +100,7 @@ describe("<CreateProgramProvider />", () => {
 
       const createContract = screen.getByTestId("create-program");
       fireEvent.click(createContract);
-
+      screen.logTestingPlaygroundURL();
```
boudra, line 103:
```suggestion
```

boudra (commented):
It would be very useful to understand how this impacted the bundle size if you can share your findings with bundle analyzer

packages/round-manager/src/features/api/abi/RoundFactoryABI.ts
```diff
@@ -0,0 +1,227 @@
+export default [
```
boudra, line 1: I noticed that with these new ABIs we now might have some duplication, do we still need the human readable ABIs? this one for example: https://github.com/gitcoinco/grants-stack/blob/7d9ef9c1f5d0beda8fcd1f36452ae00756e81c56/packages/round-manager/src/features/api/abi/roundFactory.ts

packages/grant-explorer/package.json
vacekj: I see no reason not to upgrade, v18 will be LTS only by October '23 and v20 will be the new LTS by then. Unless something specifically requires us to stay on v18, I'd say let's keep it up to date. Not a strong opinion though, this is a minor thing.

packages/round-manager/src/context/round/UpdateRoundContext.tsx
```diff
@@ -136,41 +156,37 @@ const _updateRound = async ({
...
     if (editedGroups.StartAndEndTimes) {
-      console.log("updating start and end times")
-      transactionBuilder.add(UpdateAction.UPDATE_ROUND_START_AND_END_TIMES, [
-        Date.parse(round?.applicationsStartTime.toString()) / 1000,
-        Date.parse(round?.applicationsEndTime.toString()) / 1000,
-        Date.parse(round?.roundStartTime.toString()) / 1000,
-        Date.parse(round?.roundEndTime.toString()) / 1000,
+      console.log("updating start and end times");
+      transactionBuilder.add("updateStartAndEndTimes", [
+        BigInt(Date.parse(round?.applicationsStartTime.toString()) / 1000),
```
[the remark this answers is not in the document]
vacekj: If I understand this correctly, this is taking the epoch (ms) representation of the date and dividing by 1000 to seconds. I would be fine by rounding to the nearest second to avoid BigInt errors. Is that the best solution here?

packages/round-manager/package.json
vacekj: good catch, this was a remnant of development, reverted back to defaults.

michellema1208:
@vacekj I'm seeing some of the tasks in the original ticket have not yet been completed - should this ticket be tested after those have been completed? 

Also, how are we determining AC for this ticket (am not seeing in the OG ticket)? Is it just if we are able to do all of the core actions that Mo mentioned above?

packages/round-manager/jest.config.ts
```diff
@@ -0,0 +1,34 @@
...
+
+module.exports = {
+  preset: "ts-jest/presets/js-with-ts",
+  testEnvironment: "<rootDir>/src/tests/fixedEnvironment.js",
+  globals: {
+    Uint8Array: Uint8Array,
+    ArrayBuffer: ArrayBuffer,
+    "ts-jest": {
+      diagnostics: {
+        exclude: ["**"],
```
boudra, line 20: what does this do? does it disable type checking for tests?

boudra:
you have some conflicts

packages/round-manager/jest.config.ts
vacekj: Yes. For some reason, typechecking is behaving differently in Jest and in the rest of the project, most likely because of the different environments. I'll try enabling it again and fixing the errors after I test all the onchain interactions.

packages/round-manager/src/context/program/CreateProgramContext.tsx
```diff
@@ -268,6 +270,7 @@ async function waitForSubgraphToUpdate(
       payload: { indexingStatus: ProgressStatus.IS_SUCCESS },
     });
   } catch (error) {
+    debugger;
```
vacekj, line 273: Remove

packages/round-manager/src/context/round/UpdateRoundContext.tsx
```diff
@@ -121,13 +138,16 @@ const _updateRound = async ({
 
     // direct contract updates
     if (editedGroups.MatchAmount) {
-      console.log("updating match amount")
-      const decimals = getPayoutTokenOptions(chainId).find(token => token.address === round.token)?.decimal
+      console.log("updating match amount");
+      const decimals = getPayoutTokenOptions(chainId).find(
+        (token) => token.address === round.token
+      )?.decimal;
       // use ethers to convert amount using decimals
```
vacekj, line 145: delete comment

packages/round-manager/src/features/api/application.ts
```diff
@@ -326,27 +327,26 @@ const updateApplicationStatusFromContract = async (
...
+    statuses.map((status) => ({
+      index: BigInt(status.index),
+      statusRow: BigInt(status.statusRow),
+    })),
+  ]);
+  console.log("✅ Transaction hash: ", tx);
 
   return {
-    transactionBlockNumber: blockNumber,
+    transactionBlockNumber: BigInt(0),
```
vacekj, line 349: Check if this is actually used

packages/round-manager/src/features/api/application.ts
```diff
@@ -428,64 +428,57 @@ export const updateApplicationList = async (
...
-  let receipt: any = {};
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  let tx: any = {};
-
-  if (payoutToken.address === ethers.constants.AddressZero) {
+  amount: bigint
+): Promise<{ txBlockNumber: bigint; txHash: string }> => {
+  let receipt: TransactionReceipt;
+  let txHash: Hex;
+  roundId = getAddress(roundId);
```
vacekj, line 437: check this

packages/round-manager/src/features/api/round.ts
```diff
@@ -448,27 +496,27 @@ function convertStatus(status: string | number) {
 async function loadApprovedProjects(
   round: RoundResult,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  chainId: any,
+  chainId: any
```
vacekj, line 499: remove any

packages/round-manager/src/features/api/round.ts
```diff
@@ -533,44 +581,41 @@ export async function getProjectOwners(
...
+      await payoutStrategy.read.distributionMetaPtr();
+    const distributionMetaPtr = distributionMetaPtrRes[1];
 
     if (distributionMetaPtr !== "") {
-      // fetch distribution from IPFS
       const matchingDistributionRes = await fetchFromIPFS(distributionMetaPtr);
       matchingDistribution = matchingDistributionRes.matchingDistribution;
 
-      // parse matchAmountInToken to a valid BigNumber
+      /* TODO: use viem to deserialize bigints here */
```
vacekj, line 615: remove any, remove note

packages/round-manager/src/features/api/round.ts
```diff
@@ -582,25 +627,65 @@ export async function fetchMatchingDistribution(
...
+    );
+
+  const payoutEvents = await publicClient.getFilterLogs({
+    filter: payoutFilter,
+  });
 
-  const tx = await roundImplementation.setReadyForPayout();
-  return tx.wait();
-};
+  if (payoutEvents.length > 0) {
```
vacekj, line 652: is there a better way to do this?

packages/round-manager/src/features/round/FundContract.tsx
```diff
@@ -534,7 +527,7 @@ export default function FundContract(props: {
         payoutToken: matchingFundPayoutToken!,
       });
     } catch (error) {
-      if (error === Logger.errors.TRANSACTION_REPLACED) {
+      if (error === "replaced") {
```
vacekj, line 530: intrinsic to RPC, check what is this in ethers and if viem has that?

packages/round-manager/src/features/round/ViewRoundResults.tsx
```diff
@@ -149,16 +86,16 @@ function useRevisedMatchingFunds(
...
       setFinalizedMatchesLoading(true);
-      const res = await fetchFinalizedMatches(roundId, signer);
+      const res = await fetchFinalizedMatches(roundId, publicClient);
       setFinalizedMatches(res);
     } finally {
       setFinalizedMatchesLoading(false);
     }
   }
 
+  // TODO: ???
```
vacekj, line 98: drop a link to react docs here

packages/round-manager/src/features/api/utils.ts
```diff
@@ -364,7 +365,7 @@ export const pinToIPFS = (obj: IPFSObject): Promise<{ IpfsHash: string }> => {
         ...params.headers,
         "Content-Type": "application/json",
       },
-      body: JSON.stringify({ ...params.body, pinataContent: obj.content }),
+      body: serialize({ ...params.body, pinataContent: obj.content }),
```
boudra, line 368: More context: https://wagmi.sh/core/faq#bigint-serialization

I don't think using the `serialize` / `deserialize` functions from Wagmi is a good idea. I understand why they implemented it but it doesn't apply for our case.

- We sometimes might upload user content, and since with this a string is not really a string, what happens when I answer `#bigint.000` or something to one of the round application questions?
- This was probably designed to encode and decode arbitrary data, in our case we know exactly the shape of the data, we can decode bigints manually, so it's never "lossy" like they mention in the FAQ
- This metadata is part of the protocol, if we agree that we want to represent bigint as `#bigint.xxx` then let's do it, but we're inventing a new convention that depends on the Wagmi library
- We're changing expectations of how JSON encoding / decoding is behaving, we're not expecting a normal JSON anymore, this will break either compatibility or in the future if we move from this


```typescript
type MyObject = {value: bigint};

const objectA: MyObject = { value: 69420n };

const serialized = JSON.stringify({ value: objectA.value.toString() });
const deserialized = JSON.parse(serialized) as { value: string };

const objectB = { value: BigInt(deserialized.value) };
```

hmrtn:
After testing on round manager here is what definitely works: 

- Create program flow
- Create round flow 
- Edit Round

What does not work, and may not be related to this PR is the fund contract flow. The txn and transfer are successful, but it is not displayed/updated on the FE.

packages/round-manager/src/features/round/ViewRoundResults.tsx
vacekj: @boudra could you please drop the link to the explanation to this here

packages/round-manager/src/features/round/ViewRoundSettings.tsx
```diff
@@ -251,6 +251,7 @@ export default function ViewRoundSettings(props: { id?: string }) {
 
   const updateRoundHandler = async () => {
     try {
+      // @ts-expect-error need help debugging this
```
vacekj, line 254: check this

packages/round-manager/src/features/api/application.ts
vacekj: Now actually using the transaction block number

packages/round-manager/jest.config.ts
vacekj: The issue only exists in jest, and after some investigation I'm fairly confident it's due to Jests's module handling (CommonJS/ES) and other intricacies. We can't even ignore the error because it wouldn't pass tsc/eslint because it's "ignoring an error that isn't there" for normal tsconfig. The errors happen when using the `Transition` component from headless UI, and setting the `as` props to React's Fragment.

vacekj:
Discord discussion: https://discord.com/channels/562828676480237578/1116333300135493662

Deprio for now.

Plan: 
0. Upgrade lit js sdk https://github.com/LIT-Protocol/js-sdk to potentially remove dep on ethers (or pass in ethers as in builder)
1. Use https://wagmi.sh/react/ethers-adapters to do a small PR that just install viem and upgrades wagmi. No conversions yet. Bootnode can use this in their new code.
2. See if we can refactor away some existing uses of ethers where it's not necessary (replace with wagmi hook/wagmi cli/ just remove)
3. Start refactoring one feature at a time (can reuse this PR code), testing and then merging.

jxom:
Curious what the blockers are for getting this merged? Happy to help.

gravityblast:
closed PR but kept the branch to be used in case we want to cherry pick changes

PR #1854, opened 2023-05-22
1841 - Update calculation banner decimals and update value

### LINKED ISSUE #1841: Manager: errors with matching cap calculation when editing a round
**Overview**
Two minor errors on the Funding Settings tab: 

- [ ] If I edit the "matching funds available" amount, the matching cap banner says "0% (0 DAI)"  -> it only updates after I edit the matching cap value 
Screen Shot 2023-05-19 at 10 47 48 AM

- [ ] Round up the matching cap amount to the hundredths decimal place 
Screen Shot 2023-05-19 at 10 49 41 AM

**Steps To Reproduce**

**Observed  behavior**

**Expected behavior**

**Screenshots**

**Additional context**

codenamejason:
closes #1841 

Fixed decimals to 4 places

Screenshot 2023-05-22 at 12 44 05 PM


thelostone-mc (dismissed):

michellema1208:
@codenamejason @DanieleSalatti it looks like I'm still waiting for the Fleek builds above to finish before I can test - maybe we can get the leads to help upgrade our Fleek pricing plan?

michellema1208 (changes requested):
hey @DanieleSalatti @codenamejason - great work here! Two minor pieces of feedback:

- [ ] If the number is a whole number, don't show any decimal places (i.e. show 50 DAI, not 50.0000 DAI) 
Screen Shot 2023-05-25 at 2 07 41 PM

- [ ] Reduce to only showing two decimal places 
Screen Shot 2023-05-25 at 2 08 52 PM

> **Reason:** I know this is backtracking what we talked about in standup - the reason is I'm realizing through testing that **multiplying by a percentage 1-100 never gets you more than two decimal places** (see above for example)*. Therefore, let's reduce the decimal places to two, as the third and fourth decimal places will never be used. 

*I'm not sure what elementary school math principle this is that I'm forgetting, but I tried multiple examples with ETH and DAI and I only get up to two decimal places. You get 2-4+ decimal places when dividing by a decimal/percentage.

codenamejason:
@michellema1208 when you use smaller amounts when testing this like 0.001 ETH we needed that decimal place otherwise we would just see 0's. I can update to 2 decimals for I don't think round operators will use small amounts like mentioned.

michellema1208 (approved):
hey @codenamejason - thank you for your hard work here!! 

**TL;DR**: I'm going ahead and approving this PR, given it satisfies all functional criteria. The aesthetic changes I called out also need to be fixed on the Create flow, which is out of scope for this ticket, so I'm pulling that out into a follow up ticket: https://github.com/gitcoinco/grants-stack/issues/1884

Issue #1656, opened 2023-04-28
Improve the round metadata 


thelostone-mc:
I saw that we're working on a Round settings page, if we're going to work on this area I wanted to start a discussion on the metadata format and raise some issues I noticed, it could be a good opportunity to improve it.

At the moment this is how the quadraticFundingConfig value looks like:

{
   "matchingFundsAvailable":1000,
   "matchingCap":true,
   "minDonationThreshold":true,
   "sybilDefense":true,
   "matchingCapAmount":10,
   "minDonationThresholdAmount":1
}

I think there are a couple of issues with it

- matchingCapAmount is actually a percentage and not an amount, we should either rename it to matchingCapPercentage or store an actual amount in wei
- minDonationThresholdAmount and minDonationThreshold are redundant because minDonationThresholdAmount can just be 0 to disable it
- minDonationThresholdAmount should be a wei big int, not a number
- matchingFundsAvailable is a duplicate of the RoundImplementation.matchAmount property, do they serve different purposes? which one is the source of truth? at the moment in the indexer we're using the matchAmount in the contract since it has events to update it, we should have only one to avoid confusion

any thoughts? is this something we can do now with the round settings page or is it a separate tech debt task?

hmrtn:
We should have a x-team discussion about what _exactly_ needs to be altered here before proceeding, as this change is not trivial.

boudra:
@hmrtn can you come up with a suggestion looking at the description of the issue to move this forward?

hmrtn:
From what you said I think this data schema is appropriate.
```
{
  "hasSybilDefense": Bool,
  "matchingCapPercentage": Float
  "minDonationThresholdAmount": BigInt
}

```

boudra:
I agree we should store the matchingCap as an amount and not percentage.

for the passport option I would go for `enableSybilDefense` or `sybilDefenseEnabled`, what do you think?

hmrtn:
@gravityblast @thelostone-mc What do you guys think of the proposed schema? Would like your input if anything is missing. 

I am being deliberately conscientious here because this will likely require altering quite a number of files and their dependent usages.

gravityblast:
LGTM!

boudra:
@hmrtn just fixed a bug in the calculator where were treating the min amount as a token value and didn't realize it was USD, can we name it something like:

```
minAmountUSD
```

so it's clear it's USD to the next person who looks at this

boudra:
[...]

PR #1862, opened 2023-05-22
1540-better-datetime (and form)

hmrtn:
This PR addresses #1540 and uses a new date time picker component. Specifically, it utilizes the browsers native datetime picker.
The control's UI varies in general from browser to browser and is not exactly customizable. But it is well defined. In browsers with no support (like old ie), these degrade gracefully to simple text input. 
Screenshot 2023-05-17 at 11 53 07 AM

One of the consequences of this is the loss of input UTC time. Instead, it uses the clients local. But this is handled properly throughout the application - so it should not be a problem, and imo improves the usability. 

After creating the component I took the opportunity of the "down time" to commit a much needed refactor of the application form. This involved adding a couple of new base components, pulling out the form sub components into its own file, and pulled out the validation schema into its own file. 

Note: The round edit flow also includes some of the removed hardcoded components previously used (@codenamejason!) When we get the chance, lets update this page to use the base components as well.

DISCORD CONVO HERE: https://github.com/gitcoinco/grants-stack/pull/1862

### DIFF
--- a/packages/round-manager/src/features/common/FormDropDown.tsx
+++ b/packages/round-manager/src/features/common/FormDropDown.tsx
@@ -11,6 +11,7 @@ import { CheckIcon, SelectorIcon } from "@heroicons/react/solid";
 import { Round } from "../api/types";
 import { classNames } from "common";
 import { Fragment } from "react";
+import { get } from "lodash";
 
 export const FormDropDown: FC<{
   register: UseFormRegisterReturn<string>;
@@ -22,7 +23,7 @@ export const FormDropDown: FC<{
   defaultValue: string;
 }> = ({ errors, control, label, id, options, defaultValue }) => {
   // @ts-expect-error appears on the error id key-value for the errors object
-  const errorMessage = errors[id]?.message;
+  const errorMessage = get(errors, id)?.message;
   const hasError = Boolean(errorMessage);
   const { field } = useController({
     name: id,
--- a/packages/round-manager/src/features/common/FormDropDown.tsx
+++ b/packages/round-manager/src/features/common/FormDropDown.tsx
@@ -1,0 +1,129 @@
+import { FC } from "react";
+import {
+  Control,
+  useController,
+  UseFormRegisterReturn,
+  FieldErrors,
+  FieldPath,
+} from "react-hook-form";
+import { Listbox, Transition } from "@headlessui/react";
+import { CheckIcon, SelectorIcon } from "@heroicons/react/solid";
+import { Round } from "../api/types";
+import _ from "lodash";
+import { classNames } from "common";
+import { Fragment } from "react";
+
+export const FormDropDown: FC<{
+  register: UseFormRegisterReturn<string>;
+  errors: FieldErrors<Round>;
+  control: Control<Round>;
+  label: string;
+  id: FieldPath<Round>;
+  options: string[];
+  defaultValue: string;
+}> = ({ errors, control, label, id, options, defaultValue }) => {
+  // @ts-expect-error "required" appears on the error type, not sure where from
+  const errorMessage = _.get(errors, id)?.message;
+  const hasError = Boolean(errorMessage);
+  const { field } = useController({
+    name: id,
+    defaultValue: defaultValue,
+    control: control,
+    rules: {
+      required: true,
+    },
+  });
+
+  return (
+    <div className="col-span-6 sm:col-span-3 relative mt-2">
+      <Listbox {...field}>
+        {({ open }) => (
+          <div>
+            <Listbox.Label className="text-sm mt-4 mb-2">
+              <p className="text-sm">
+                <span>{label}</span>
+                <span className="text-right text-violet-400 float-right text-xs mt-1">
+                  *Required
+                </span>
+              </p>
+            </Listbox.Label>
+            <div className="mt-1 mb-2 shadow-sm block rounded-md border-gray-300 focus:border-indigo-500 focus:ring-indigo-500 sm:text-sm">
+              <Listbox.Button
+                className={`relative w-full cursor-default rounded-md border h-10 bg-white py-2 pl-3 pr-10 text-left shadow-sm ${
+hasError
+? "border-red-300 text-red-900 placeholder-red-300 focus-within:outline-none focus-within:border-red-500 focus-within: ring-red-500"
+: "border-gray-300 focus:border-indigo-500 focus:outline-none focus:ring-1 focus:ring-indigo-500 sm:text-sm"
+}`}
+                data-testid={`${id}-testid`}
+                id={id}
+              >
+                <span className="flex items-center">
+                  {/*@ts-expect-error field.value includes the Date type, but we never actually render it */}
+                  <span className="ml-3 block truncate">{field.value}</span>
+                </span>
+                <span className="pointer-events-none absolute inset-y-0 right-0 ml-3 flex items-center pr-2">
+                  <SelectorIcon
+                    className="h-5 w-5 text-gray-400"
+                    aria-hidden="true"
+                  />
+                </span>
+              </Listbox.Button>
+              <Transition
+                show={open}
+                as={Fragment}
+                leave="transition ease-in duration-100"
+                leaveFrom="opacity-100"
+                leaveTo="opacity-0"
+              >
+                <Listbox.Options className="absolute z-10 mt-1 w-full bg-white shadow-lg max-h-60 rounded-md py-1 text-base ring-1 ring-black ring-opacity-5 overflow-auto focus:outline-none sm:text-sm">
+                  {options.map((option) => (
+                    <Listbox.Option
+                      key={option}
+                      className={({ active }) =>
+                        classNames(
+                          active ? "text-white bg-indigo-600" : "text-gray-900",
+                          "cursor-default select-none relative py-2 pl-10 pr-4"
+                        )
+                      }
+                      value={option}
+                      data-testid={`${id}-option-${option}`}
+                    >
+                      {({ selected, active }) => (
+                        <>
+                          <span
+                            className={classNames(
+                              selected ? "font-medium" : "font-normal",
+                              "block truncate"
+                            )}
+                          >
+                            {option}
+                          </span>
+                          {selected ? (
+                            <span
+                              className={classNames(
+                                active ? "text-white" : "text-indigo-600",
+                                "absolute inset-y-0 left-0 flex items-center pl-3"
+                              )}
+                            >
+                              <CheckIcon
+                                className="h-5 w-5"
+                                aria-hidden="true"
+                              />
+                            </span>
+                          ) : null}
+                        </>
+                      )}
+                    </Listbox.Option>
+                  ))}
+                </Listbox.Options>
+              </Transition>
+            </div>
+            {hasError && (
+              <p className="mt-2 text-xs text-pink-500">{errorMessage}</p>
+            )}
+          </div>
+        )}
+      </Listbox>
+    </div>
+  );
+};
--- a/packages/round-manager/src/features/common/FormDropDown.tsx
+++ b/packages/round-manager/src/features/common/FormDropDown.tsx
@@ -9,7 +9,6 @@ import {
 import { Listbox, Transition } from "@headlessui/react";
 import { CheckIcon, SelectorIcon } from "@heroicons/react/solid";
 import { Round } from "../api/types";
-import _ from "lodash";
 import { classNames } from "common";
 import { Fragment } from "react";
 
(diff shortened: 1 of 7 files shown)


packages/round-manager/src/features/common/FormDropDown.tsx
```diff
@@ -0,0 +1,129 @@
...
+  register: UseFormRegisterReturn<string>;
+  errors: FieldErrors<Round>;
+  control: Control<Round>;
+  label: string;
+  id: FieldPath<Round>;
+  options: string[];
+  defaultValue: string;
+}> = ({ errors, control, label, id, options, defaultValue }) => {
+  // @ts-expect-error "required" appears on the error type, not sure where from
+  const errorMessage = _.get(errors, id)?.message;
```
DanieleSalatti, line 26: qq: what are the benefits of using lodash instead of directly doing `const errorMessage = errors[id]?.message;`? I don't get the "required" error either if I try that locally, so we might be able to remove the `@ts-expect-error` comment.

packages/round-manager/src/features/common/FormDropDown.tsx
```diff
@@ -0,0 +1,129 @@
...
+hasError
+? "border-red-300 text-red-900 placeholder-red-300 focus-within:outline-none focus-within:border-red-500 focus-within: ring-red-500"
+: "border-gray-300 focus:border-indigo-500 focus:outline-none focus:ring-1 focus:ring-indigo-500 sm:text-sm"
+}`}
+                data-testid={`${id}-testid`}
+                id={id}
+              >
+                <span className="flex items-center">
+                  {/*@ts-expect-error field.value includes the Date type, but we never actually render it */}
+                  <span className="ml-3 block truncate">{field.value}</span>
```
DanieleSalatti, line 62: Can we do a type check on this field and figure out what to render if it's not a string? Just in case future changes make the problem resurface.

DanieleSalatti: `{typeof field.value === 'string' ? field.value : field.value.toISOString()}` or thereabouts

packages/round-manager/src/features/common/FormDropDown.tsx
DanieleSalatti: Alternatively, can we `import { get } from "lodash";` like in `FormInputField` instead of importing everything? (I still prefer the solution above if possible though)

packages/round-manager/src/features/common/FormInputField.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { get } from "lodash";
+
+export const FormInputField: FC<{
+  register: UseFormRegisterReturn<string>;
+  errors: FieldErrors<Round>;
+  label: string;
+  id: string;
+  placeholder?: string;
+}> = ({ register, errors, label, id, placeholder }) => {
+  const errorMessage = get(errors, id)?.message;
```
DanieleSalatti, line 14: We could use optional chaining here as well? See comment above.

packages/round-manager/src/features/round/ApplicationFormComponents.tsx
```diff
@@ -0,0 +1,227 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
DanieleSalatti, line 1: Instead of doing this on the whole file, would it be better to be explicit on which lines we are applying this lint exclusion to?

packages/round-manager/src/features/round/ApplicationFormComponents.tsx
```diff
@@ -0,0 +1,227 @@
...
+  );
+}
+
+
+interface DatetimeProps {
+  control: Control<Round>;
+  name: any;
+  label: string;
+  date?: moment.Moment;
+  setDate?: (date: moment.Moment) => void;
```
DanieleSalatti, line 190: If this gets the `date` prop as a parameter which is defined as optional, I think it should be able to take null as a parameter too?

i.e.:

`setDate?: (date: moment.Moment | null) => void;`

DanieleSalatti (commented):
Left some comments - I know you refactored a lot of this so I likely have commented on older code, but figured that would be useful still.

Love the refactoring, good job!

boudra:
@hmrtn is there a reason why there's a new PR / branch made for this issue? I see it's not the first time, is it a problem with git?

melissa-neira (dismissed):
lgtm 🕺
question - will UTC time still be displayed on a round's Explorer and Manager page?

packages/round-manager/src/features/round/ApplicationFormComponents.tsx
hmrtn: I do not think so - the anys are necessary in this case for each component. See: https://github.com/gitcoinco/grants-stack/pull/1805#issuecomment-1551915471

hmrtn:
> lgtm 🕺 question - will UTC time still be displayed on a round's Explorer and Manager page?

This is converted to UTC on store! So it should not affect the application at all.

nfrgosselin (dismissed):
LGTM - thanks Hans!

packages/round-manager/src/features/round/ApplicationFormComponents.tsx
boudra: @hmrtn you can remove the anys like this:

```diff
diff --git a/packages/round-manager/src/features/round/ApplicationFormComponents.tsx b/packages/round-manager/src/features/round/ApplicationFormComponents.tsx
index 631233ce..92de79f3 100644
--- a/packages/round-manager/src/features/round/ApplicationFormComponents.tsx
+++ b/packages/round-manager/src/features/round/ApplicationFormComponents.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 import { FC } from "react";
 import "react-datetime/css/react-datetime.css";
 import {
@@ -7,6 +6,7 @@ import {
   FieldErrors,
   useController,
   UseFormRegisterReturn,
+  Path,
 } from "react-hook-form";
 import { Listbox, RadioGroup } from "@headlessui/react";
 import { Program, Round } from "../api/types";
@@ -15,15 +15,17 @@ import moment from "moment";
 import { FormInputField } from "../common/FormInputField";
 import { FormDropDown } from "../common/FormDropDown";
 
-export function RoundName(
-  props: any
-) {
+export function RoundName(props: {
+  register: UseFormRegisterReturn<string>;
+  errors: FieldErrors<Round>;
+}) {
   return (
     <FormInputField
       {...props}
       label="Round Name"
       id="roundMetadata.name"
-      placeholder="Enter round name here."    />
+      placeholder="Enter round name here."
+    />
   );
 }
 
@@ -61,7 +63,10 @@ export function ProgramChain(props: { program: Program }) {
   );
 }
 
-export const ContactInformation = (props: any) => (
+export const ContactInformation = (props: {
+  register: UseFormRegisterReturn<string>;
+  errors: FieldErrors<Round>;
+}) => (
   <FormInputField
     {...props}
     label="Contact Information"
@@ -83,7 +88,7 @@ export function Support(props: {
     "Telegram Group Invite Link",
     "Google Form Link",
     "Other (please provide a link)",
-  ]
+  ];
 
   return (
     <FormDropDown
@@ -181,36 +186,40 @@ export function RoundType(props: {
   );
 }
 
-
 interface DatetimeProps {
   control: Control<Round>;
-  name: any;
+  name: Path<Round>;
   label: string;
   date?: moment.Moment;
-  setDate?: (date: moment.Moment) => void;  
+  setDate?: (date: moment.Moment) => void;
   minDate?: moment.Moment;
 }
 
-export const Datetime: FC<DatetimeProps> = ({ control, name, label, date, setDate, minDate }) => {
-  const now = moment().format('YYYY-MM-DDTHH:mm');
+export const Datetime: FC<DatetimeProps> = ({
+  control,
+  name,
+  label,
+  date,
+  setDate,
+  minDate,
+}) => {
+  const now = moment().format("YYYY-MM-DDTHH:mm");
   return (
     <div className="relative w-full border-0 p-0 placeholder-grey-40 focus:ring-0 text-sm">
-      <label
-        htmlFor={name}
-        className="block text-[10px]"
-      >
+      <label htmlFor={name} className="block text-[10px]">
         {label}
       </label>
       <Controller
         control={control}
         name={name}
-        defaultValue={date?.format('YYYY-MM-DDTHH:mm')}
+        defaultValue={date?.format("YYYY-MM-DDTHH:mm")}
         render={({ field }) => (
           <input
-            data-testid={`${name}-testid`} 
+            data-testid={`${name}-testid`}
             type="datetime-local"
             {...field}
-            min={minDate ? minDate.format('YYYY-MM-DDTHH:mm') : now}
+            value={field.value?.toString()}
+            min={minDate ? minDate.format("YYYY-MM-DDTHH:mm") : now}
             className="block w-full border-0 p-0 focus:ring-0"
             onChange={(e) => {
               // Convert the local datetime to UTC
@@ -224,4 +233,3 @@ export const Datetime: FC<DatetimeProps> = ({ control, name, label, date, setDat
     </div>
   );
 };
-
```

boudra: I also noticed you might not be using prettier, do you mind running it

packages/round-manager/src/features/round/__tests__/RoundDetailForm.test.tsx
```diff
@@ -129,121 +128,164 @@ describe("<RoundDetailForm />", () => {
...
       name: /next|launch/i,
     });
-    const supportSelection = screen.getByTestId("support-type-select");
+
+    const supportSelection = screen.getByTestId("roundMetadata.support.type-testid");
     fireEvent.click(supportSelection);
-    const firstSupportOption = screen.getAllByTestId("support-type-option")[0];
-    fireEvent.click(firstSupportOption);
+
+    const emailSupportOption = screen.getByTestId("roundMetadata.support.type-option-Email");
```
boudra, line 135: It would be good if we could avoid using test ids unless we cannot avoid it, in this case I see we have labels, can we not use those?

michellema1208:
hey @hmrtn - just tested this! Overall, I'm loving the new date picker UI and _definitely_ think it's a huge improvement from what we had before. A few minor issues I found: 

- [ ] Add default buffer time if I select the current date for any of the fields -> I would recommend at least adding a 15 min buffer by default, user can go back and adjust it if needed (more context and why on this in the Loom) 

- [ ] When I click next and then go back to Round Details page where I can configure dates, my previously selected dates are wiped in the UI 

- [ ] Duplicate error messages - I would recommend keeping just our existing error messages to keep the style consistent with other parts of the flow. + @melissa-neira @willsputra for a second op from design POV! I also want to note that the copy on the new error alert (see screenshot below) is actually incorrect - I can't set the times to be equal, it must be greater than the value of the start time. 

Screen Shot 2023-05-30 at 4 50 48 PM

- [ ] Is there a way to trigger the date picker if I click anywhere on the input field instead of only if I click on the calendar icon? It feels more intuitive to me that clicking anywhere should trigger the date picker, but also would love second op from design (+ @melissa-neira @willsputra ) 

Loom walkthrough: https://www.loom.com/share/42c391fbfae3447f9e9b36339040715f

michellema1208:
one other thing to note: since we're also shipping Round Edit flow this week, we would also want to make sure this new date picker is reflected in the edit flow. My soft suggestion is that once issues have been addressed in this PR, we merge it into a temporary integration branch, and follow up with a second PR adding this new date picker to the round edit flow as well once that's shipped. Then, we would merge the integration branch -> staging -> prod. 

+ @gravityblast @boudra - ultimately defer to you both how we want to handle this! Just want to make sure an edge case doesn't emerge from the round edit flow having the old date picker and the round create flow having the new date picker.

packages/round-manager/src/features/round/ApplicationFormComponents.tsx
hmrtn: Resolved in #1656!

melissa-neira:
I agree with @michellema1208! I'd especially love it if the date picker could be clicked from anywhere in the input field, not just the calendar.

willsputra:
> Duplicate error messages - I would recommend keeping just our existing error messages to keep the style consistent with other parts of the flow. + @melissa-neira @willsputra for a second op from design POV! I also want to note that the copy on the new error alert (see screenshot below) is actually incorrect - I can't set the times to be equal, it must be greater than the value of the start time.

Agree! Let's keep the existing error message and remove the new one

>  Is there a way to trigger the date picker if I click anywhere on the input field instead of only if I click on the calendar icon? It feels more intuitive to me that clicking anywhere should trigger the date picker, but also would love second op from design (+ @melissa-neira @willsputra )

+1!

packages/round-manager/src/features/round/__tests__/RoundDetailForm.test.tsx
boudra: I don't think it's a personal preference, the testing library is designed around the idea that tests behave like users, so we're testing the UI as close as to how it's used, and test IDs go against that completely, users don't see test IDs, they see buttons, links, labels, text etc. That's how tests should query the DOM. Test IDs should only be used in a pinch for very specific cases.

I recommend going through the testing library documentation: https://testing-library.com/docs/guiding-principles

This article summarizes the approach pretty well: https://derekndavis.com/posts/getbytestid-overused-react-testing-library

vacekj:
Next steps: Split into problem fix and Refactor (native date picker) ticket. Decide if the native experience is better.

vacekj:
Why close? This PR is still very relevant.

JV

PR #1873, opened 2023-05-25
1656 improve round meta data

hmrtn:


packages/round-manager/src/features/common/Datetime.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+interface DatetimeProps<T extends FieldValues> {
+    control: Control<T>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    name: any;
+    label: string;
+    date?: moment.Moment;
+    setDate?: (date: moment.Moment) => void;  
+    minDate?: moment.Moment;
+  }
```
gravityblast, line 13: there's something wrong with the indentation here, it should be 2 spaces instead of 4, with `}` being at column 0

packages/round-manager/src/features/common/Datetime.tsx
```diff
@@ -0,0 +1,47 @@
...
+  
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  export const Datetime = <T extends FieldValues>({ control, name, label, date, setDate, minDate }: DatetimeProps<T>): ReactElement => {
```
gravityblast, line 16: I see `date` is declared but never used

packages/round-manager/src/features/common/Datetime.tsx
```diff
@@ -0,0 +1,47 @@
+import moment from "moment";
+import { ReactElement } from "react";
+import { Control, Controller, FieldValues } from "react-hook-form";
+
+interface DatetimeProps<T extends FieldValues> {
+    control: Control<T>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    name: any;
```
gravityblast, line 8: shall we put a type instead of any?

PR #1883, opened 2023-05-26
remove husky and lint staged, disable git hooks

### LINKED ISSUE #1814: rework git hooks
- [x] Remove husky + lint-staged in favour of left-hook
- [x] Make left-hook opt-in
- [x] Remove all hooks, only leave prettier before commit

https://github.com/evilmartians/lefthook/wiki/Migration-from-husky-with-lint-staged

=> Hooks will be fast, will never fail and opt-in

vacekj:
fixes #1814 

left in a lefthook to trigger prettier for those who want


boudra (changes requested):
Since hooks are not installed automatically now, we should add information of how to use lefthook in the README and once we merge we can share the link to the team so they know how to enable it if they want to.

melissa-neira:
defer to eng review 🕺

gravityblast (dismissed):

gravityblast:
I think we should put the prettier check in CI but it's longer, let start with what @boudra said

DanieleSalatti (approved):

PR #1889, opened 2023-05-30
feat: load intercom script async

### LINKED ISSUE #1822: Async load Intercom
put the async attr on both of those scripts

vacekj:
fixes #1822 

eng review only

### DIFF
--- a/packages/builder/public/index.html
+++ b/packages/builder/public/index.html
@@ -54,7 +54,7 @@
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
 
-    <script>
+    <script async>
       window.intercomSettings = {
         api_base: "https://api-iam.intercom.io",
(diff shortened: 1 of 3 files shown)


boudra (changes requested):
The `async` attribute is only relevant for scripts that have to be downloaded (scripts with a `src` attribute), here nothing is downloaded because the source is inline.

Furthermore, if you look closely to the code that actually loads Intercom (minified), it already loads asynchronously.

vacekj:
> The `async` attribute is only relevant for scripts that have to be downloaded (scripts with a `src` attribute), here nothing is downloaded because the source is inline.
> 
> Furthermore, if you look closely to the code that actually loads Intercom (minified), it already loads asynchronously.

🤦🏻‍♂️ my bad, didn't realize that it doesn't apply to inline scripts. Closing this PR then 🤝

PR #1935, opened 2023-06-09
[1583] feat: hide past rounds from builder project page

### LINKED ISSUE #1583: Builder: hide past rounds on My Applications section
**Overview**
On my Project Details page, I can still see past rounds (rounds that are already over) in the My Applications section

**Steps To Reproduce**

1. Go to https://builder.gitcoin.co/#/chains/10/registry/0x8e1bD5Da87C14dd8e08F7ecc2aBf9D1d558ea174/projects/0
2. Notice that many of the rounds under "My Applications" have already ended 

**Expected behavior**
I should only see rounds in the "My Applications" section that are upcoming or active rounds (rounds that have not reached its end date). 

**Screenshots**
Image

DanieleSalatti:
##### Description

What it says on the tin.

##### Refers/Fixes

Closes #1583 

##### Testing

Tested locally

### DIFF
--- a/packages/builder/src/actions/rounds.ts
+++ b/packages/builder/src/actions/rounds.ts
@@ -9,9 +9,9 @@ import { RootState } from "../reducers";
 import { Status } from "../reducers/rounds";
 import PinataClient from "../services/pinata";
 import { MetaPtr, ProgramMetadata, Round, RoundMetadata } from "../types";
-import { getProviderByChainId } from "../utils/utils";
 import { RoundApplicationMetadata } from "../types/roundApplication";
 import { parseRoundApplicationMetadata } from "../utils/roundApplication";
+import { getProviderByChainId } from "../utils/utils";
 
 export const ROUNDS_LOADING_ROUND = "ROUNDS_LOADING_ROUND";
 interface RoundsLoadingRoundAction {
@@ -331,4 +331,12 @@ export const loadRound =
     };
 
     dispatch(roundLoaded(address, round));
+  };
+
+export const loadRounds =
+  (addressAndChain: Map<string, number>) => async (dispatch: Dispatch) => {
+    addressAndChain.forEach((chainId, address) => {
+      console.log("DASA loading round", address, chainId);
+      dispatch<any>(loadRound(address, chainId));
+    });
   };
--- a/packages/builder/src/actions/rounds.ts
+++ b/packages/builder/src/actions/rounds.ts
@@ -336,7 +336,6 @@ export const loadRound =
 export const loadRounds =
   (addressAndChain: Map<string, number>) => async (dispatch: Dispatch) => {
     addressAndChain.forEach((chainId, address) => {
-      console.log("DASA loading round", address, chainId);
       dispatch<any>(loadRound(address, chainId));
     });
   };
--- a/packages/builder/src/components/grants/About.tsx
+++ b/packages/builder/src/components/grants/About.tsx
@@ -8,10 +8,10 @@ import { RootState } from "../../reducers";
 import colors from "../../styles/colors";
 import { FormInputs, Metadata, Project } from "../../types";
 import { formatDateFromMs } from "../../utils/components";
+import GreenVerifiedBadge from "../badges/GreenVerifiedBadge";
 import Calendar from "../icons/Calendar";
 import LinkIcon from "../icons/LinkIcon";
 import ApplicationCard from "./ApplicationCard";
-import GreenVerifiedBadge from "../badges/GreenVerifiedBadge";
 
 export default function About({
   project,
@@ -56,7 +56,7 @@ export default function About({
         <span className="text-[20px]">My Applications</span>
       </Box>
       <Box>
-        {props.applications.map((application, index) => {
+        {props.applications.map((application) => {
           const roundID = application?.roundID;
           const cardData = {
             application,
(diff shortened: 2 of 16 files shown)


packages/builder/src/components/grants/About.tsx
```diff
@@ -56,19 +56,15 @@ export default function About({
...
             roundID,
             chainId: application.chainId,
           };
 
-          return (
-            <Box key={[roundID, index].join("-")} m={2}>
-              <ApplicationCard applicationData={cardData} />
-            </Box>
-          );
+          return <ApplicationCard applicationData={cardData} />;
```
gravityblast, line 67: @DanieleSalatti don't we still need a key here?

gravityblast (changes requested):
with this implementation we are adding a hidden class name directly in the card component, which means we can never use the Card component with past applications, even in other possible pages. 

Instead of rendering all the applications and hiding them, I would suggest to move the logic in the `map` loop and skipping the Card component for past rounds

melissa-neira (changes requested):
While testing, I noticed that each project that I had rendered the application column differently. 
This project rendered correctly (although it took a few extra seconds to load):
screencapture-bafybeiatebykyd3tudj75zfy2owwytuo5r4zihgnavkdsvs5janlp2xzau-on-fleek-co-2023-06-14-13_29_53
This project's application column was empty:
screencapture-bafybeiatebykyd3tudj75zfy2owwytuo5r4zihgnavkdsvs5janlp2xzau-on-fleek-co-2023-06-14-13_29_00
This project didn't even have a "my applications" column.
screencapture-bafybeiatebykyd3tudj75zfy2owwytuo5r4zihgnavkdsvs5janlp2xzau-on-fleek-co-2023-06-14-13_29_18

Anyone else experiencing this?

DanieleSalatti:
I see what the problem is, that needs to be fixed.

Two questions:
* Would it be better to rename "My Applications" with current applications? So it's more obvious why applications disappear.
* When no applications are displayed, should we hide the column or should we leave the header? Since it takes some time to load the applications (even when there is only one), I think we should show the header with a spinner when loading, and then a message of some sort if there are no active rounds.

Thinking about it, maybe we leave "My Applications" at the top, never hide the column, and add "This grant has not applied to active rounds yet" as a message then the column is empty?

@melissa-neira

vacekj:
@melissa-neira @DanieleSalatti it would be great if we could start using loading skeletons instead of spinners. https://www.npmjs.com/package/react-loading-skeleton. It looks much better than a generic spinner.

willsputra:
> Thinking about it, maybe we leave "My Applications" at the top, never hide the column, and add "This grant has not applied to active rounds yet" as a message then the column is empty?

@DanieleSalatti I like the idea! created the updated Figma screen here: https://www.figma.com/file/obBzCCL8KwoTlVBb5g3ccN/Builder-S17?type=design&node-id=5183-45653&t=x3oIeUUcFuwdjU3g-4

@melissa-neira @michellema1208 would love your eyes especially on the copy! feel free to edit :)

melissa-neira:
@willsputra looks good to me!

packages/builder/src/actions/rounds.ts
```diff
@@ -332,3 +332,11 @@ export const loadRound =
 
     dispatch(roundLoaded(address, round));
   };
+
+export const loadRounds =
+  (addressAndChain: Map<string, number>) => async (dispatch: Dispatch) => {
+    addressAndChain.forEach((chainId, address) => {
+      console.log("DASA loading round", address, chainId);
```
gravityblast, line 339: @DanieleSalatti we have to remove this log

packages/builder/src/components/grants/About.tsx
```diff
@@ -34,8 +39,46 @@ export default function About({
...
+
+  useEffect(() => {
+    const roundToChain = new Map<string, number>();
+    props.applications.forEach((application) => {
+      roundToChain.set(application.roundID, application.chainId);
+    });
+    dispatch(loadRounds(roundToChain));
+  }, [dispatch, props.applications]);
+
+  useEffect(() => {
```
boudra, line 57: we could do `const appsToShow = useMemo(..)` here instead of `useEffect + useState`

DanieleSalatti:
> > Thinking about it, maybe we leave "My Applications" at the top, never hide the column, and add "This grant has not applied to active rounds yet" as a message then the column is empty?
> 
> @DanieleSalatti I like the idea! created the updated Figma screen here: https://www.figma.com/file/obBzCCL8KwoTlVBb5g3ccN/Builder-S17?type=design&node-id=5183-45653&t=x3oIeUUcFuwdjU3g-4
> 
> @melissa-neira @michellema1208 would love your eyes especially on the copy! feel free to edit :)

Looks good, on it.

DanieleSalatti:
Tests failing now that I'm using the chakra skeleton.
Relevant: https://github.com/chakra-ui/chakra-ui/issues/6036

bhargavaparoksham (approved):
LGTM.

DanieleSalatti:
> @DanieleSalatti loading time is really long, over 2 minutes on my end. Project application status appears like this: 
[...]
 It still hasn't loaded yet 😅

I'll take a look to see if we are doing any unnecessary operation. We fetch that data from IPFS, which can be slow, but 2 minutes is a really long time.

vacekj:
@meglister @michellema1208 same request - can you please confirm that the issue this PR was trying to solve in the first place is still present in our latest deployment? If so, then I can drive this PR over the line.

thelostone-mc:
Closing this out as the PR has gone stale. Will revisit when if and when it gets added to the backlog

PR #1942, opened 2023-06-14
fix: error page fix

### LINKED ISSUE #1890: Manager: remove error page popping up every time I take an action in Manager
**Overview**
Problem: Every time after I take an action within Manager (i.e. approve/reject a project, edit a round, etc.), I briefly see the Manager app error page flash up on my screen, and then after a second it refreshes. Let's remove that error page from popping up, which is confusing. 

**Steps To Reproduce**
1. Go to https://rmgitcoin.on.fleek.co/ 
2. Try to Approve or Reject a project, Edit a round, etc. 
3. When the transaction is successful, the page refreshes and briefly flashes the error page, before returning to the correct page

vacekj:
fixes #1890

fixes error page popping up on navigation due to incorrect permissions check in ViewRoundPage.tsx

## PR checklist
- [x] Includes only changes relevant to the original ticket. Refactoring and formatting needs to be separated.
- [x] Doesn't contain `any`
- [x] Doesn't contain unnecessary type casts and non-null asserts
- [x] Doesn't add `@ts-ignore`
- [x] Doesn't disable lints
- [x] Doesn't use `useState` just for computation
- [x] Doesn't add components inside of components that depend on variables in the parent scope
- [x] Doesn't propagate optional values without good reason
- [x] Doesn't duplicate existing code
- [x] Validates out-of-domain data
- [x] Does things as simple as possible, but not simpler.
- [x] Doesn't reinvent the wheel or create premature and unnecessary abstractions.
- [x] Is properly formatted.

### DIFF
--- a/packages/round-manager/src/features/round/ViewRoundPage.tsx
+++ b/packages/round-manager/src/features/round/ViewRoundPage.tsx
@@ -15,7 +15,6 @@ import {
 } from "@heroicons/react/solid";
 import { formatUTCDateAsISOString, getUTCTime } from "common";
 import { Button } from "common/src/styles";
-import { useEffect, useState } from "react";
 import { Link, useParams } from "react-router-dom";
 import tw from "tailwind-styled-components";
 import { ReactComponent as GrantExplorerLogo } from "../../assets/grantexplorer-icon.svg";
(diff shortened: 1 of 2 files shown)


packages/round-manager/src/features/round/ViewRoundPage.tsx
```diff
@@ -49,53 +47,28 @@ import ViewRoundStats from "./ViewRoundStats";
 export default function ViewRoundPage() {
   datadogLogs.logger.info("====> Route: /round/:id");
   datadogLogs.logger.info(`====> URL: ${window.location.href}`);
-  const { id } = useParams();
+
+  const { id } = useParams() as { id: string };
```
DanieleSalatti, line 51: Nice!

DanieleSalatti (approved):
LGTM

michellema1208 (approved):
No longer seeing the error page! When I approve/reject a project, I'm still seeing an additional page refresh (Loom: https://www.loom.com/share/33da8c457dcd4ce7bacbc86b2935b38a), in case that's a quick fix. 

Still going to approve this though - it solves the original issue!

melissa-neira (approved):

PR #1984, opened 2023-06-27
feat: add pgn testnet to common and builder package

### LINKED ISSUE #1961: [PGN] Builder: add PGN testnet network
**User Story**
As a user of Builder 
I want to be able to switch onto PGN testnet 
so that I can create a project or apply to a round on PGN testnet 

**Acceptance Criteria** 
GIVEN that I am on Builder
WHEN I click on the Network selection toggle 
THEN I see an option to switch to PGN testnet 

GIVEN that I want to switch onto PGN testnet
WHEN I click on PGN testnet 
THEN I am prompted to switch my network 

GIVEN that I successfully switch my network to PGN testnet
WHEN I look at the Network selection toggle
THEN I see the network that I am currently on is PGN testnet 

**Tech Details**

vacekj:
fixes #1961


melissa-neira:
is there anything to test here? if not, I defer to the devs 😎

bhargavaparoksham (dismissed):
lgtm!

packages/builder/src/utils/graphql.ts
```diff
@@ -17,8 +8,7 @@ export enum ChainId {
  */
 const getGraphQLEndpoint = (chainId: number): string | undefined => {
```
gravityblast, line 9: we need to add the subgraph for PGN

gravityblast (changes requested):

packages/builder/src/utils/graphql.ts
vacekj: Yes - and I will change this object to be typesafe once we have that URL - meaning it will fail typecheck if the subgraph URL isn't specified for all chains. (and I will also check that it's not an empty string so that we don't "cheat" it by just putting an empty string hehe)

packages/builder/src/actions/projects.ts
```diff
@@ -8,7 +8,7 @@ import { global } from "../global";
 import { RootState } from "../reducers";
 import { AppStatus, Application, ProjectStats } from "../reducers/projects";
 import { ProjectEvents, ProjectEventsMap } from "../types";
-import { graphqlFetch } from "../utils/graphql";
+import graphqlFetch from "../utils/graphql";
```
bard, line 11: Mostly personal preference, but I tend to always use named imports, because with default imports the name is up to the developer who writes the import, so there's a chance that different developers might pick different names in different files for the same imported thing. Could happen with named imports, too, but it takes an extra step so it's a little less likely.

packages/builder/src/contracts/deployments.ts
```diff
@@ -33,9 +37,12 @@ export const addresses: DeploymentAddress = {
...
     projectRegistry: "0x8e1bD5Da87C14dd8e08F7ecc2aBf9D1d558ea174",
   },
+  pgnTestnet: {
+    projectRegistry: "0x6294bed5B884Ae18bf737793Ef9415069Bf4bc11",
+  },
 };
 
-export const addressesByChainID = (chainID: number) => {
-  const chainName: string = chains[chainID];
+export const addressesByChainID = (chainID: ChainId) => {
```
bard, line 45: Worth declaring the return type?

bard (approved):

packages/builder/src/actions/projects.ts
vacekj: I do the same and agree, however there is a lint enabled (only in Builder) to prefer default exports when there's only a single export (that I don't think is a good idea :D ) - I'd say let's disable the lint globally - it's being eslint-ignored in other places of the project anyway.

vacekj: - disabled the lint rule and changed back to named export

packages/builder/src/contracts/deployments.ts
vacekj: Yes and fixed.

packages/builder/src/contracts/deployments.ts
```diff
@@ -1,20 +1,25 @@
-export const chains: { [key: number]: string } = {
-  31337: "localhost",
+import { ChainId } from "common";
+
+export const chains = {
+  1: "mainnet",
   5: "goerli",
   10: "optimism",
-  69: "optimisticKovan", // todo: update to 420: "optimisticGoerli"
```
gravityblast, line 5: why are we removing optimism testnets?

packages/builder/src/contracts/deployments.ts
```diff
@@ -1,20 +1,25 @@
...
+
+export type ChainName = (typeof chains)[keyof typeof chains];
 
-export type DeploymentAddress = {
-  [key: string]: {
-    projectRegistry: string | undefined;
-  };
+type DeploymentAddresses = {
+  projectRegistry: string | undefined;
+};
```
gravityblast, line 17: let's leave an empty line here

gravityblast (approved):

packages/builder/src/contracts/deployments.ts
vacekj: Kovan is deprecated

vacekj: We can add support for optimistic goerli, but I feel like we're bloating the scope of this PR a bit already, so would leave that for another one.

gravityblast: ok, but just for the next time, if it's not related to the PR, don't even remove the deprecated one. if you remove it you need to add the new one

gravityblast (approved):

bard (approved):

PR #1986, opened 2023-06-27
feat: add pgn testnet to manager

### LINKED ISSUE #1962: [PGN Testnet] Manager: add PGN testnet network
**User Story**
As a user of Manager 
I want to be able to switch onto PGN testnet 
so that I can create a round on PGN testnet 

**Acceptance Criteria** 
GIVEN that I am on Manager 
WHEN I click on the Network selection toggle 
THEN I see an option to switch to PGN testnet 

GIVEN that I want to switch onto PGN testnet
WHEN I click on PGN testnet 
THEN I am prompted to switch my network 

GIVEN that I successfully switch my network to PGN testnet
WHEN I look at the Network selection toggle
THEN I see the network that I am currently on is PGN testnet 

**Tech Details**

vacekj:
closes #1962 and implements and closes https://github.com/gitcoinco/grants-stack/issues/1973

Testing
- [x] Program creation
- [x] Round Creation (on PGN testnet we only need one transation, and updated modal)
- [x] Round update
- [x] Round fund


vacekj:
Round creation is working, but subgraph isn't indexing the new rounds.

bhargavaparoksham (changes requested):
@vacekj round deployment is failing for me with the following error

Screenshot 2023-07-05 at 3 03 30 PM

bhargavaparoksham:
@vacekj We might have to update the copy for this modal as well,

Screenshot 2023-07-05 at 3 00 47 PM

vacekj:
The failure is likely due to your wallet not estimating gas correctly. I have to set gas manually in Frame.
I am removing the modal as it's no longer relevant.

packages/round-manager/src/features/api/round.ts
```diff
@@ -70,7 +72,7 @@ export class TransactionBuilder {
  */
 export async function getRoundById(
   signerOrProvider: Web3Provider,
-  roundId: string,
+  roundId: string
```
bard, line 75: Nothing to change but worth checking if a formatter went rogue anywhere as there are a few "comma diffs" and may cause noise in future reviews.

bard (dismissed):

packages/common/src/index.ts
```diff
@@ -155,6 +155,9 @@ const getGraphQLEndpoint = async (chainId: ChainId) => {
     case ChainId.FANTOM_TESTNET_CHAIN_ID:
       return `${process.env.REACT_APP_SUBGRAPH_FANTOM_TESTNET_API}`;
 
+    case ChainId.PGN_TESTNET:
+      return `${process.env.REACT_APP_SUBGRAPH_PGN_TESTNET_API}`;
```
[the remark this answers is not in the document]
vacekj: i can't see a reason not to just return variable. it's most likely just a remnant from the past.

packages/round-manager/src/context/round/CreateRoundContext.tsx
```diff
@@ -117,12 +121,14 @@ interface _createRoundParams {
   context: CreateRoundState;
   signerOrProvider: Signer;
   createRoundData: CreateRoundData;
+  useNewRoundInit: boolean;
```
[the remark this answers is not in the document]
vacekj: open to a diff. name, however this parameter is going to be removed once all networks support the aio round creation, which should be soon.

packages/round-manager/src/context/round/CreateRoundContext.tsx
```diff
@@ -139,10 +145,14 @@ const _createRound = async ({
...
-      roundMetadataWithProgramContractAddress.eligibility.requirements = roundMetadataWithProgramContractAddress.eligibility?.requirements.filter(
-        obj => obj.requirement != ""
-      );
+    if (
+      roundMetadataWithProgramContractAddress &&
+      roundMetadataWithProgramContractAddress.eligibility
+    ) {
+      roundMetadataWithProgramContractAddress.eligibility.requirements =
+        roundMetadataWithProgramContractAddress.eligibility?.requirements.filter(
+          (obj) => obj.requirement != ""
```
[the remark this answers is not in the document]
vacekj: I just formatted this code, didn't write it. Not sure why the original author used loose comparison here, but I'm too scared to make it strict because not sure if the requirements are actually just empty strings, or if they can be nulls etc. So would leave as-is for now.

packages/round-manager/src/features/api/round.ts
vacekj: I format everything using prettier with the default config. Prettier --check says that this file is okay, so it's likely that the formatting was not correct previously. and is now correct. Unfortunately, Prettier in default config only inserts trailing commas where es5 supports them, and function parameters with types are not one of those. I'd be happy to enable this option, but I fear it might open a pandora's box of bikeshedding over configuration.

packages/round-manager/src/context/round/CreateRoundContext.tsx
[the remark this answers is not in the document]
bard: Got it, no point in worrying them.

packages/round-manager/src/context/round/CreateRoundContext.tsx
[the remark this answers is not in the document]
bard: Agreed. Can you just add a comment so anybody who's tempted to "fix" knows why it's like that?

packages/round-manager/src/features/api/round.ts
bard: I wasn't aware the source file hadn't been previously fed to prettier. Good that it now is!

vacekj: We will do a prettier pass on the whole codebase soon + fail CI if it doesn't pass prettier check. There is a git hook to autoformat staged files too, see README for more info if you'd like.

bard (dismissed):

packages/round-manager/src/features/round/RoundApplicationForm.tsx
```diff
@@ -304,32 +307,21 @@ export function RoundApplicationForm(props: {
   };
 
   const formSubmitModals = () => (
-    <InfoModal
```
bhargavaparoksham, line 307: @vacekj rather than completely removing it, i think its better we make this conditional as well depending on useNewRoundInit?

bhargavaparoksham (changes requested):
Screenshot 2023-07-05 at 6 57 50 PM

As we are not deploying 3 different contracts directly, the ui would be cleaner, if we can just show only 1 deploying step i.e Deploying round. It's better to get product & design feedback on this before going ahead with current implementation.

cc: @michellema1208

packages/round-manager/src/features/round/RoundApplicationForm.tsx
vacekj: all the chains will use the new UI soon, useNewRoundInit will be removed and all chains will use the aio deploy soon. So no reason to keep it.

vacekj:
> Screenshot 2023-07-05 at 6 57 50 PM
> As we are not deploying 3 different contracts directly, the ui would be cleaner, if we can just show only 1 deploying step i.e Deploying round. It's better to get product & design feedback on this before going ahead with current implementation.
> 
> cc: @michellema1208

We actually are deploying 3 contracts, so the UI is true to what is actually happening. Unless @michellema1208 objects, I would keep it as-is.

packages/round-manager/src/features/round/RoundApplicationForm.tsx
```diff
@@ -223,11 +220,17 @@ export function RoundApplicationForm(props: {
         operatorWallets: props.initialData.program.operatorWallets,
       } as Round;
 
-      await createRound({
-        roundMetadataWithProgramContractAddress,
-        applicationQuestions,
-        round,
-      });
+      /*TODO: detect version from round impl contract? */
+      const useNewRoundInit = chain.id === pgnTestnet.id;
```
gravityblast, line 224: @vacekj we need to enable this for all the chains since Allo is deploying/upgrade the contracts everywhere

bhargavaparoksham (approved):

melissa-neira (approved):

bard (approved):

PR #2000, opened 2023-06-28
Migrate to WalletConnect v2

vacekj:
fixes walletconnect v2 for round manager and explorer. builder is not possible without considerable effort rn because it's using the lit-js-sdk with some custom overrides etc. and I couldn't get walletconnecto to correctly work there.
- upgrades wagmi & rainbowkit to their latest 0.x versions (before mandatory viem upgrade)
- adds pgn testnet to all frontends
- ViewFundGrantees is just prettier
- `as '0x{string}` is wagmi having stricter types for addresses than just strings - you're going to see a lot more of this in the viem migration
- removes custom Chain objects in favour of wagmi-exported ones.


packages/grant-explorer/src/app/wagmi.ts
```diff
@@ -1,13 +1,17 @@
...
+  getDefaultWallets,
+} from "@rainbow-me/rainbowkit";
 import {
   coinbaseWallet,
   injectedWallet,
   walletConnectWallet,
   metaMaskWallet,
 } from "@rainbow-me/rainbowkit/wallets";
-import { createClient, configureChains, chain } from "wagmi";
+import { mainnet, goerli, fantomTestnet, optimism } from "wagmi/chains";
```
gravityblast, line 13: what are the values from this variables. I wouldn't change the RPC we used so far just to remove some lines of code

packages/grant-explorer/src/app/wagmi.ts
```diff
@@ -79,21 +62,30 @@ const allChains: Chain[] =
 export const { chains, provider, webSocketProvider } = configureChains(
   allChains,
   [
-    infuraProvider({ apiKey: process.env.REACT_APP_INFURA_ID }),
+    infuraProvider({ apiKey: process.env.REACT_APP_INFURA_ID as string }),
```
gravityblast, line 65: it's better to have it failing if the var is not set. let's use something like 

```
const getEnv = (name: string): string => {
  const value = process.env[name];

  if (value === undefined) {
    throw new Error(`missing env var ${name}`);
  }

  return value;
};
```

packages/grant-explorer/src/app/wagmi.ts
```diff
@@ -79,21 +62,30 @@ const allChains: Chain[] =
...
     publicProvider(),
   ]
 );
 
+const projectId = "2685061cae0bcaf2b244446153eda9e1";
```
gravityblast, line 70: let's have it as an env var

packages/grant-explorer/src/features/round/ViewCartPage.tsx
```diff
@@ -98,6 +98,7 @@ export default function ViewCart() {
       ? { addressOrName: address }
       : { addressOrName: address, token: selectedPayoutToken.address };
 
+  // @ts-expect-error Temp until viem
```
gravityblast, line 101: what's the error we can have here?

gravityblast (changes requested):

packages/grant-explorer/src/app/wagmi.ts
vacekj: They are the exact same. They are just exported from a different part of wagmi.
You can check that by Go to definition in your editor.
Not sure what you mean by "just to remove some lines of code"

packages/grant-explorer/src/app/wagmi.ts
vacekj: Agree, but I'd argue that env var validation is out of scope of this ticket.

packages/grant-explorer/src/features/round/ViewCartPage.tsx
vacekj: @ts-expect-error silences the typescript error on the next line. unlike ts-ignore, it will error out if there is no error on the next line, making sure that you remove the ts-expect-error once it's not necessary anymore.
In this case, you can see the error by removing the @ts-expect-error line. 
The error in this case comes from wagmi requiring addresses to be of type ``0x${string}`` (a hexadecimal string). In viem, you can import this type (called `Hex`), however I couldn't find the import in our current version of wagmi. We silence the error here for now, and it will alert us to remove the expect-error once we adopt viem.

packages/grant-explorer/src/app/wagmi.ts
DanieleSalatti: Off topic, but for env variable validation I propose we explore Zod and [t3-env](https://github.com/t3-oss/t3-env).

vacekj: Sounds good, tracking ticket here for GS: https://github.com/gitcoinco/grants-stack/issues/1508

gravityblast: if it's out of scope let's remove the `as string` as well

vacekj: the function expects a string, while we are passing in `string | undefined`. usually we do `!` to assert that it's not undefined, but I find `as string` to be much more visible, but if you want, I can swap for `!`. Whatever you prefer.

DanieleSalatti (approved):
LGTM
Panding @gravityblast's answers on the open discussions

packages/grant-explorer/src/app/wagmi.ts
gravityblast: in both cases we would pass an empty string which doesn't make sense, that's why I think a simple check can help avoiding having an environment without that var

gravityblast (dismissed):

packages/builder/src/setupTests.ts
```diff
@@ -4,6 +4,8 @@
 // learn more: https://github.com/testing-library/jest-dom
 import "@testing-library/jest-dom";
 import "@testing-library/jest-dom/extend-expect";
-import { TextEncoder } from "util";
+import { TextEncoder, TextDecoder } from "util";
 
 global.TextEncoder = TextEncoder;
+// @ts-expect-error We love Jest
```
bard, line 10: :rofl:

packages/round-manager/src/features/round/FundContract.tsx
```diff
@@ -104,13 +101,16 @@ export default function FundContract(props: {
       matchingFundPayoutToken?.address ===
       "0x0000000000000000000000000000000000000000"
         ? undefined
-        : matchingFundPayoutToken?.address,
+        : (matchingFundPayoutToken?.address as `0x{string}`),
```
bard, line 104: For the future, is there any non-casting way of reading a string into a `0x{string}`? Couldn't find much about it

bard (approved):

melissa-neira (commented):
lgtm!

melissa-neira (approved):

michellema1208 (approved):
lgtm!

packages/builder/package.json
```diff
@@ -132,6 +134,7 @@
     "jest-environment-jsdom": "^27.0",
     "jest-transform-stub": "^2.0.0",
     "jest-when": "^3.5.1",
+    "lokijs": "^1.5.12",
```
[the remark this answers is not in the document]
vacekj: This was an attempt to fix WalletConnect for Builder, an error message from WC told me to install it, so I obliged. In the end, it only choked on another, unrelated error. Will remove in a future PR as to not invalidate the reviews on this one.

packages/builder/src/setupTests.ts
vacekj: One does not simply _use_ Jest. You must configure it. Understand it. Master it. Recite the source code from memory. Offer sacrifices to keep it running.

PR #2019, opened 2023-07-05
[round-manager] - 1 - Direct grant: round creation

nicosampler:
##### Description

This pull request adds the direct-round functionality under a feature flag. The env name to enable this new feature is `REACT_APP_DIRECT_GRANT_ENABLED=true`. Let me know if you prefer to use a different name.

* When a program has no rounds created, the screen was changed to offer the user the possibility to create a direct or quadratic round.
* On the viewProgramPage, the create button triggers a modal to offer the user the possibility to create a direct or quadratic round.
* The create round form is a bit different than the quadratic one. 

[Figma: Create direct round](https://www.figma.com/proto/rax6pBts8JLgCh1E58pqf7/Gitcoin?page-id=35%3A220&type=design&node-id=314-2104&viewport=294%2C521%2C0.32&t=ieHzN9o8AW9cMrrf-1&scaling=min-zoom&starting-point-node-id=35%3A221&show-proto-sidebar=1&mode=design).

Note: The changes applied were done to be compatible with the current code. 
Note: There are many files changed, but many of them were changed because I renamed a file. 
Note: The subgraph URL is hardcoded to support the new changes required for DR, but is compatible with QF. It was not feature-flagged to avoid requesting to change the ENV var or creating a new environment.


vacekj:
Quite a big PR, will mob-review with my teammates once we get the chance.

nicosampler:
yes, sorry for that, upcoming PR will be smaller, I promise. The reason this one is too big is because I wanted to wrap the whole "create direct round" functionality to have something that could be tested.

packages/round-manager/src/features/api/payoutStrategy/payoutStrategy.ts
```diff
@@ -63,6 +66,20 @@ export const deployMerklePayoutStrategyContract = async (
...
 };
 
+/**
+ * @param signerOrProvider
+ * @returns the factory address.
+ */
+export const getDirectPayoutFactoryAddress = async (
+  signerOrProvider: Signer
+): Promise<{ payoutContractAddress: string }> => {
+  const chainId = await signerOrProvider.getChainId();
```
bard, line 76: Since `signerOrProvider` has no other use here, would it make sense to turn this function into this?

```ts
export const getDirectPayoutFactoryContractAddress = (chainId: ChainId): string => { ... }
```

That way, we turn the function sync, and we avoid it returning something that sounds slightly different that what the name says it would return ("DirectPayoutFactoryAddress" vs "payoutContractAddress").

packages/round-manager/src/features/api/round.ts
```diff
@@ -271,9 +273,15 @@ export async function listRounds(
         roundMetadata,
         applicationMetadata,
         applicationsStartTime: new Date(round.applicationsStartTime * 1000),
-        applicationsEndTime: new Date(round.applicationsEndTime * 1000),
+        applicationsEndTime:
+          round.applicationsEndTime == ethers.constants.MaxUint256.toString()
```
bard, line 277: Please always use `===`

packages/round-manager/src/features/api/round.ts
```diff
@@ -271,9 +273,15 @@ export async function listRounds(
...
+            ? maxDate
+            : new Date(round.applicationsEndTime * 1000),
```
bard, line 279: Does the name `applicationsXYZTime` come from code generation or is it something we have control over and introduced in this PR? If the latter, we should hint at the unit in the name itself, e.g. `applicationsEndTimeS`

packages/round-manager/src/features/api/round.ts
```diff
@@ -300,57 +308,80 @@ export async function listRounds(
...
+    const roundFactory = RoundFactory__factory.connect(
+      roundFactoryAddress,
+      signerOrProvider
     );
 
-    if (!round.applicationsEndTime) {
-      round.applicationsEndTime = round.roundStartTime;
+    let initRoundTimes: string[] = [];
+    const parseDate = (date: Date) =>
+      (new Date(date).getTime() / 1000).toString();
```
bard, line 323: This would seem like a `formatDate` rather than a `parseDate`, no?

Also, could it be shortened to this?

```ts
const dateToSeconds = (date: Date): string => (date.getTime() / 1000).toString()
```

packages/round-manager/src/features/api/round.ts
```diff
@@ -300,57 +308,80 @@ export async function listRounds(
...
+    if (isQF) {
+      if (!round.applicationsEndTime) {
```
bard, line 325: Please make the condition explicit to avoid false positives, e.g.

```ts
if (round.applicationsEndTime === undefined) { ...
```

packages/round-manager/src/features/api/round.ts
```diff
@@ -300,57 +308,80 @@ export async function listRounds(
...
+        round.applicationsEndTime = round.roundStartTime;
+      }
+      initRoundTimes = [
+        parseDate(round.applicationsStartTime),
+        parseDate(round.applicationsEndTime),
+        parseDate(round.roundStartTime),
+        parseDate(round.roundEndTime as Date),
```
bard, line 332: If the cast is needed here because `roundEndTime` is actually a number? Then `parseDate` should take `(date: Date | number)`

packages/round-manager/src/features/api/round.ts
```diff
@@ -300,57 +308,80 @@ export async function listRounds(
...
-      token.decimal,
-    );
+    const token = ethers.constants.AddressZero;
+    let parsedTokenAmount = ethers.constants.Zero;
+
+    if (isQF) {
+      // Ensure tokenAmount is normalized to token decimals
+      const tokenAmount =
+        round.roundMetadata.quadraticFundingConfig?.matchingFundsAvailable || 0;
+      const pyToken = payoutTokens.filter(
```
bard, line 370: Was this meant to be `payToken`?

packages/round-manager/src/features/api/round.ts
```diff
@@ -300,57 +308,80 @@ export async function listRounds(
...
-      tokenAmount.toString(),
-      token.decimal,
-    );
+    const token = ethers.constants.AddressZero;
+    let parsedTokenAmount = ethers.constants.Zero;
+
+    if (isQF) {
+      // Ensure tokenAmount is normalized to token decimals
+      const tokenAmount =
+        round.roundMetadata.quadraticFundingConfig?.matchingFundsAvailable || 0;
```
bard, line 369: Please use `??` if the check is for `undefined`

packages/round-manager/src/features/api/votingStrategy/votingStrategy.ts
```diff
@@ -9,22 +12,32 @@ import { qfVotingStrategyFactoryContract } from "../contracts";
...
     const chainId = await signerOrProvider.getChainId();
 
-    const _QFVotingStrategyFactory = qfVotingStrategyFactoryContract(chainId); //roundFactoryContract(chainId);
-    const qfVotingStrategyFactory = new ethers.Contract(
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      _QFVotingStrategyFactory.address!,
-      _QFVotingStrategyFactory.abi,
+    if (!isQF) {
+      // Note: In direct-rounds, we use a dummy contract for the voting contract.
+      return Promise.resolve({
```
bard, line 24: Since `deployVotingContract` is `async`, the return value should already be wrapped in a promise, so probably the `Promise.resolve` call here can be skipped.

packages/round-manager/src/features/api/votingStrategy/votingStrategy.ts
```diff
@@ -9,22 +12,32 @@ import { qfVotingStrategyFactoryContract } from "../contracts";
...
+        votingContractAddress: dgVotingStrategyDummyContract(chainId),
+      });
+    }
+
+    const _votingStrategyFactory = qfVotingStrategyFactoryContract(chainId);
+    if (!_votingStrategyFactory.address)
```
bard, line 30: Please use an explicit check such as `if (_votingStrategyFactory.address === undefined)`

packages/round-manager/src/features/common/FormWizard.tsx
```diff
@@ -6,9 +6,14 @@ export interface FormWizardProps {
   initialCurrentStep?: number;
   // eslint-disable-next-line @typescript-eslint/ban-types
   initialData?: object;
+  configuration?: object;
```
bard, line 9: Is the configuration data actually free from or is there an expected shape and content? If so, it would be better to specify it, otherwise we're not allowing TypeScript to help us.

packages/round-manager/src/features/round/CreateRoundPage.tsx
```diff
@@ -32,6 +33,20 @@ export default function CreateRound() {
 
   const [searchParams] = useSearchParams();
   const programId = searchParams.get("programId");
+  const roundCategoryParam = searchParams.get("roundCategory");
+  const roundCategory =
+    roundCategoryParam == "direct"
```
bard, line 38: Please use the exact comparison operator `===`

packages/round-manager/src/features/round/RoundApplicationForm.tsx
```diff
@@ -114,9 +181,16 @@ export function RoundApplicationForm(props: {
 
   const navigate = useNavigate();
 
-  const defaultQuestions: ApplicationMetadata["questions"] =
-    // @ts-expect-error TODO: either fix this or refactor the whole formstepper
-    formData?.applicationMetadata?.questions ?? initialQuestions;
+  const roundCategory =
+    props.configuration?.roundCategory || RoundCategory.QuadraticFunding;
```
bard, line 185: `||` → `??`

packages/round-manager/src/features/round/RoundApplicationForm.tsx
```diff
@@ -114,9 +181,16 @@ export function RoundApplicationForm(props: {
...
+
+  // @ts-expect-error TODO: either fix this or refactor the whole formstepper
+  const questionsArg = formData?.applicationMetadata?.questions;
+  const defaultQuestions: ApplicationMetadata["questions"] = questionsArg
+    ? questionsArg
+    : roundCategory == RoundCategory.QuadraticFunding
```
bard, line 191: `==` → `===`

packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -67,59 +68,91 @@ export const RoundValidationSchema = yup.object().shape({
...
 interface RoundDetailFormProps {
   stepper: typeof FormStepper;
   initialData?: { program?: Program };
+  configuration?: { roundCategory?: RoundCategory };
 }
 
 export function RoundDetailForm(props: RoundDetailFormProps) {
   const program = props.initialData?.program;
+  const roundCategory =
+    props.configuration?.roundCategory || RoundCategory.QuadraticFunding;
```
bard, line 154: `||` → `??`

packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -205,19 +243,31 @@ export function RoundDetailForm(props: RoundDetailFormProps) {
...
                 />
                 {program && <ProgramChain program={program} />}
               </div>
 
-              <p className="mt-6 mb-2 text-sm text-grey-400">
-                Where can applicants reach you and/or your team if support is
-                needed?
-              </p>
+              {/* vault Address */}
+              {roundCategory == RoundCategory.Direct && (
```
bard, line 257: `==` → `===`

bard: Will skip marking the remaining ones (if any) as by now you're probably getting sick of me doing it :)

packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -234,326 +284,411 @@ export function RoundDetailForm(props: RoundDetailFormProps) {
...
+                              className="h-4 w-4 text-indigo-600 focus:ring-indigo-500 border-gray-300 rounded"
+                              checked={field.value}
+                              onChange={(e) => {
+                                field.onChange(e.target.checked);
+                                if (e.target.checked) {
+                                  // reset input and trick type validation
+                                  // for direct round, when date is empty it will be set to maxDate
+                                  setValue(
+                                    "roundEndTime",
+                                    null as unknown as Date
```
bard, line 580: Was it not possible to have the `roundEndTime` value as `null | Date`?

packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -234,326 +284,411 @@ export function RoundDetailForm(props: RoundDetailFormProps) {
...
+                              }}
+                              onChange={(date) => {
+                                field.onChange(moment(date));
+                                setRoundEndDate(moment(date));
+                                if (rollingApplications) {
+                                  setApplicationEndDate(moment(date));
+                                  setValue(
+                                    "applicationsEndTime",
+                                    moment(date).toDate()
+                                  );
```
bard, line 639: Can we find a way of letting the reader know:

- what type is `date` in `(date) => {`?
- why are we passing `date` to `moment()` repeatedly? What type is that converting to, if any? 
- why are we converting from that (whatever that is) to `date` before giving it to `setValue`?

I'm mostly concerned that we're making it very easy for the reader to introduce bugs here due to not understanding what's going on.

packages/round-manager/src/features/round/RoundDetailForm.tsx
```diff
@@ -234,326 +284,411 @@ export function RoundDetailForm(props: RoundDetailFormProps) {
...
+                                }
+                              }}
+                              isValidDate={disableBeforeRoundStartDate}
+                              utc={true}
+                              // we use renderInput because there is a bug with the library
+                              // if the input is cleared programmatically the value is removed
+                              // but the visual date is not updated
+                              // ref: https://stackoverflow.com/a/64972324/2524608
```
bard, line 647: :+1: for providing context

bard (changes requested):

packages/round-manager/src/features/api/round.ts
nicosampler: This code was like so before, It refers to the tokens that can be selected on the creating round form.

packages/round-manager/src/features/api/round.ts
nicosampler: no, these dates were always like so. The only changes we've done was to set the proper values depending on the type of the round.

packages/common/src/index.ts
```diff
@@ -157,7 +160,11 @@ const getGraphQLEndpoint = async (chainId: ChainId) => {
 
     case ChainId.GOERLI_CHAIN_ID:
     default:
-      return `${process.env.REACT_APP_SUBGRAPH_GOERLI_API}`;
+      return process.env.REACT_APP_DIRECT_GRANT_ENABLED
+        ? TMP_DIRECT_ROUND_SUBGRAPH_URL
+        : `${process.env.REACT_APP_SUBGRAPH_GOERLI_API}`;
+    // TODO: rollback this change once the subgraph is deployed
+    // return `${process.env.REACT_APP_SUBGRAPH_GOERLI_API}`;
```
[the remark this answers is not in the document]
nicosampler: sure

packages/round-manager/src/features/api/payoutStrategy/payoutStrategy.ts
nicosampler: good call :+1:

packages/round-manager/src/features/round/RoundDetailForm.tsx
nicosampler: it was, but if I changed it to be null | Date in the Round type, there were a lot of places breaking because of it.
I could have changed/fixed them, but I didn't for the following reasons:

* This trick is not the best, I admit it, but it is well contained as we are handling all the dates before we send the creating the round tx. This `null` should not have impact beyond the creating form. 
* Setting the roundEndTime as null, would require a LOT of changes in many parts of the code as the Round type is widely used. And to be honest I didn't want to add more noise to this PR as it is a bit complex and already big enough.
* Not sure if making roundEndTime `null | Date` is something that we want for the rest of the app. This is the only place were it need to be possibly `null`. Perhaps an alternative could be defining a new Round type and overriding the roundEndTime.

packages/round-manager/src/features/api/contracts.ts
```diff
@@ -181,6 +177,39 @@ export const qfVotingStrategyFactoryContract = (
...
+      address = "";
+      break;
+    }
+    case ChainId.FANTOM_TESTNET_CHAIN_ID: {
+      address = "";
+      break;
+    }
+    case ChainId.GOERLI_CHAIN_ID:
+    default: {
+      address = "0x717A2cCDD81944e64c8BD9BB1D179A241dE14B46";
```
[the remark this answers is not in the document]
nicosampler: I totally agree with you. 
I tried to do it as you suggested but there are many tests breaking as they are not passing any chainId. 

Although doing this change is possible I would suggest to addressing it in a different PR.

PR #2031, opened 2023-07-07
feat: env var verification webpack plugin + CI changes + misc

vacekj:
- Adds a webpack plugin for verifying env variables. This extracts all occurences of `process.env.SOMETHING` from the codebase, using a regex, and checks that they are defined at build time.
- Unifies .env.examples, by doing `vercel env pull`, then cleaning up and commenting the resulting .env.local file
- Unifies environment variable names to take full advantage of shared environments in vercel
- Changes CI such that: 
	-  workflow names are fixed and correspond to what is actually happening
	- disables building, as that check is done in both vercel and fleek
	- enables typechecking, linting and testing in all apps
- updates docs on DAI and adds docs on USDC.
- fixes scripts in global `package.json` to use turbo
- removes an unused file in GE
- fixes type-errors in tests


packages/verify-env/src/index.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  const consoleMock = vi
+    .spyOn(console, "error")
+    .mockImplementation(() => undefined);
+  const processMock = vi.spyOn(process, "exit").mockImplementation(() => {
+    return undefined as never;
+  });
+
+  verify(__dirname + "/test");
+  expect(processMock).toHaveBeenCalledOnce();
+  expect(processMock).toHaveBeenCalledWith(1);
```
bard, line 45: Isn't there any canonical way for plugins to signal non-recoverable errors and let webpack exit? Calling `process.exit` from inside a plugin seems very unplugin-y.

packages/verify-env/src/test/file.ts
```diff
@@ -0,0 +1,7 @@
+const allChains = process.env.REACT_APP_ENV === "development";
```
bard, line 1: Nit: consider naming this in a way that makes it clear that it's a boolean, see e.g. the `isXYZ` convention in Chakra. `doesXYZ` or `shouldXYZ` are good too.

packages/verify-env/src/testSource.txt
```diff
@@ -0,0 +1,91 @@
+import "@rainbow-me/rainbowkit/styles.css";
```
bard, line 1: I *think* it should be possible to save this as a `ts` but disable typechecking with a directive at the top, if that helps.

packages/verify-env/src/verify.ts
```diff
@@ -0,0 +1,83 @@
...
+      "Error: Some environment variables are not set or are empty:"
+    );
+    if (badVars.length > 0) {
+      console.error(`\n${badVars.join("\n")}\n`);
+    }
+    process.exit(1);
+  }
+}
+
+const getFilesWithExtensions = (
```
bard, line 39: Can probably replace this with one of https://www.npmjs.com/search?q=glob (ideally whatever is already used by webpack). Less code = fewer things that can break

packages/verify-env/src/verify.ts
```diff
@@ -0,0 +1,83 @@
...
+  return envVars;
+};
+
+export function validateVariables(variables: string[]) {
+  return variables.filter((variable) => {
+    const value = process.env[variable];
+    return (
+      value === undefined ||
+      value === "" ||
+      value === "undefined" ||
```
bard, line 79: Hmm, are there actually cases where env variables would get the values "undefined" and "null"?

.github/workflows/grant-explorer.yml
```diff
@@ -30,11 +30,14 @@ jobs:
         run: |
           pnpm install
 
-
-      - name: Build Grant Explorer
+      - name: Lint Explorer
         run: |
-          pnpm re-build
+          pnpm re-lint
```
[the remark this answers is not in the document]
vacekj: agreed. will fix

package.json
```diff
@@ -14,17 +14,21 @@
...
+    "rm-typecheck": "turbo run typecheck --filter=round-manager",
     "// grant explorer script": "====== packages/grant-explorer specific ======",
     "re-build": "turbo run build --filter=grant-explorer",
     "re-test": "turbo run test --filter=grant-explorer",
     "re-start": "pnpm --filter grant-explorer run start",
     "re-typecheck": "turbo run typecheck --filter=grant-explorer",
+    "re-lint": "turbo run lint:ci --filter=grant-explorer",
     "// builder script": "====== packages/builder specific ======",
-    "b-start": "pnpm --filter builder run start"
+    "b-start": "pnpm --filter builder run start",
```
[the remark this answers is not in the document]
vacekj: it does, but since the whole point of turbo is to cache and skip tasks that shouldn't be run (like lint, build, test etc.) and we always want to run start, i would rather have pnpm manage that.

packages/builder/src/index.tsx
```diff
@@ -80,8 +80,9 @@ if ((queryError || queryCode) && queryState && /^github-.*/.test(queryState)) {
   // always close the redirected window
   window.close();
 }
-
-const pathname = process.env.REACT_APP_PATHNAME;
+/* this makes env var validation miss this var and therefore make it optional */
+// eslint-disable-next-line @typescript-eslint/dot-notation
+const pathname = process.env["REACT_APP_PATHNAME"];
```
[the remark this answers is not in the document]
vacekj: Yeah, open to ideas on how to handle those elegantly.

packages/grant-explorer/vitest.config.ts
```diff
@@ -11,6 +11,8 @@ export default defineConfig({
     ],
     globals: true,
     environmentMatchGlobs: [
+      ["**/voting.test.tsx", "happy-dom"],
```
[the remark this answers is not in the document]
vacekj: Updated.

packages/verify-env/src/index.test.ts
vacekj: there seems to be a way, but in the interest of making this plugin bail as soon as possible (not wait for others to finish), i'm gonna leave it as this for now.

packages/verify-env/src/test/file.ts
vacekj: done.

packages/verify-env/src/testSource.txt
vacekj: it is, using `// @ts-nocheck`, but in the interest of not tripping up the rest of my (and others') IDEs, i'm gonna leave it as .txt.

packages/verify-env/src/verify.ts
vacekj: true and fixed.

packages/verify-env/src/verify.ts
vacekj: not if you manually edit them, but as it goes in javascript land, a null or undefined might sneak in anywhere, so i'd rather guard against this.

packages/builder/src/utils/datadog.tsx
```diff
@@ -23,7 +23,7 @@ const initDatadog = () => {
 
   // Init datadog-logs
   datadogLogs.init({
-    clientToken: process.env.REACT_APP_DATADOG_LOGGER_CLIENT_TOKEN || "",
+    clientToken: process.env.REACT_APP_DATADOG_CLIENT_TOKEN || "",
```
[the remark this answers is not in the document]
vacekj: not sure what you mean here - the var is still in .env.example

packages/builder/src/index.tsx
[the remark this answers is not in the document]
bard: An explicit schema would be ideal. That way not only you can handle optional variables, but you can actually validate the variables' values, not just their presence.

bard (approved):

michellema1208 (approved):

PR #2052, opened 2023-07-14
add pgn mainnet to all apps

### LINKED ISSUE #2046: Enable PGN Mainnet on Explorer, Builder and Manager

vacekj:
fixes #2046 

- makes ChainId a number (a TS enum of numbers, which compiles down to a number) to align with the rest of the package ecosystem
- fixes tests and some other places where a string chainid was expected
- Replaces a lot of functions interacting with ChainIds with respective type-safe variants
- adds the pgn chain to common package and adds it to all apps
- adds env vars to .env.samples and .env.example
- removes unused code in common package
- removes unused code in explorer package that was copy-pasted from manager previously


gravityblast:
@vacekj in the title we wrote testnet but here we are talking about mainnet right?

michellema1208 (changes requested):
am running into two issues: 

- [ ] PGN not available as a network selection on Builder

Screen Shot 2023-07-17 at 5 49 38 PM

- [x] When I successfully switch the network to PGN, I can't take any user action in Manager, i.e. create a Program (see below). If I switch my network back to Optimism, for example, then I see that I can create a program, create a round, etc. 

Screen Shot 2023-07-17 at 5 49 26 PM

vacekj:
Both should be fixed now @michellema1208 , including the testnet and mainnet tokens in explorer and manager.

packages/builder/src/utils/graphql.ts
```diff
@@ -1,10 +1,6 @@
 import { ChainId } from "common";
 
-export type GraphQLEndpoints = {
-  [key in ChainId]: string;
-};
-
-const graphQlEndpoints: GraphQLEndpoints = {
+const graphQlEndpoints: Record<ChainId, string> = {
```
bard, line 3: Great catch here... I guess the type was too wide and let some things slip through, whereas inferring it from the value makes it the tightest it could be.

packages/common/src/chains.ts
```diff
@@ -31,3 +37,29 @@ export const pgnTestnet: Chain = {
   },
   testnet: true,
 };
+
+export const pgn: Chain = {
+  id: 424,
```
bard, line 42: Could we use `ChainId.PGN` here, so that numeric id's are only declared in the ChainId enum?

(Ah, but I see that the `ChainId` enum wasn't introduced in this PR, just moved around.)

packages/grant-explorer/src/context/__tests__/QFDonationContext.test.tsx
```diff
@@ -78,6 +78,8 @@ describe("<QFDonationProvider />", () => {
 
       fireEvent.click(screen.getByTestId("submit-qf-donations"));
 
+      screen.logTestingPlaygroundURL();
```
bard, line 81: Leftover?

packages/builder/src/utils/wagmi.ts
```diff
@@ -18,6 +18,27 @@ const ftmMainnetIcon = FantomFTMLogo;
 const alchemyId = process.env.REACT_APP_ALCHEMY_ID;
 const infuraId = process.env.REACT_APP_INFURA_ID;
 
+export const pgn: Chain = {
```
bhargavaparoksham, line 21: can we import this from common to avoid duplication?

packages/grant-explorer/src/features/api/utils.ts
```diff
@@ -194,59 +212,52 @@ export const payoutTokens = [
...
-      return `${process.env.REACT_APP_SUBGRAPH_FANTOM_TESTNET_API}`;
-
-    case ChainId.PGN_TESTNET_CHAIN_ID:
-      return `${process.env.REACT_APP_SUBGRAPH_PGN_TESTNET_API}`;
-
-    case ChainId.GOERLI_CHAIN_ID:
-    default:
-      return `${process.env.REACT_APP_SUBGRAPH_GOERLI_API}`;
-  }
+const getGraphQLEndpoint = (chainId: ChainId) => `${graphQlEndpoints[chainId]}`;
```
bard, line 251: Are the backticks necessary here?

bhargavaparoksham (changes requested):

bard (approved):

packages/common/src/chains.ts
vacekj: We could, but this custom definition is going to be upstreamed to wagmi anyway, so I would leave it.

packages/grant-explorer/src/features/api/utils.ts
vacekj: Yes, this is the solution to the fetch testing bug we chatted about yesterday. The best solution is to improve the test, but the backticks preserve existing behaviour.

packages/grant-explorer/src/context/__tests__/QFDonationContext.test.tsx
vacekj: Yes, deleted, thanks.

packages/grant-explorer/src/features/api/utils.ts
bard: If you feel like pushing one more time (can't blame you if you don't :)) please add a comment so one can see the reason for the apparent redundancy.

vacekj: yep, fixed

michellema1208:
@vacekj confirming that the Builder issue is now fixed! 

I'm still running into the Manager issue though - when I load Manager connected to PGN, the page is missing all of the core actions (i.e. create a program). See my screenshot above

bhargavaparoksham:
@vacekj when creating a new program, the program chain shows up as wrong network,

Screenshot 2023-07-18 at 8 19 48 PM

cc: @michellema1208

michellema1208:
@gravityblast @vacekj - I'm able to successfully donate to a round on PGN testnet on ETH, but within testing on this PR, I don't see my results anywhere -- I don't see the donation amount on the Donation History page or the project pages, not on Manager under Results, not on Builder under Stats. Is this a bug? Or is this because the PR environment is not connected to the Indexer in some way? 

Loom: https://www.loom.com/share/8cebe7a8cc124123b4ea5e061a1e6f82

bhargavaparoksham (changes requested):
https://github.com/gitcoinco/grants-stack/pull/2052#issuecomment-1640380477

this needs to be fixed.

vacekj:
@bhargavaparoksham @michellema1208 both are now fixed.

bhargavaparoksham (approved):

gravityblast (approved):

michellema1208 (approved):

PR #2053, opened 2023-07-14
chore: split up Apply component and disable default export eslint rule

vacekj:
- splits out the huge Apply.tsx in builder (1200-line files are not ok 😬)
- disables the eslint rule to prefer default exports on files with single export (prefer named exports to prevent mis-naming imports)
- removes eslint-ignores of the aforementioned rule
- fixes the error for applications not open

### DIFF
--- a/packages/builder/src/actions/roundApplication.ts
+++ b/packages/builder/src/actions/roundApplication.ts
@@ -127,6 +127,7 @@ const dispatchAndLogApplicationError = (
   datadogLogs.logger.error(error, {
     roundAddress,
   });
+  debugger;
   dispatch(applicationError(roundAddress, error, step));
 };
 
@@ -233,6 +234,7 @@ export const submitApplication =
 
       deterministicApplication = objectToDeterministicJSON(application as any);
     } catch (error) {
+      console.error(error);
       dispatchAndLogApplicationError(
         dispatch,
         roundAddress,
--- a/packages/builder/src/actions/roundApplication.ts
+++ b/packages/builder/src/actions/roundApplication.ts
@@ -127,7 +127,6 @@ const dispatchAndLogApplicationError = (
   datadogLogs.logger.error(error, {
     roundAddress,
   });
-  debugger;
   dispatch(applicationError(roundAddress, error, step));
 };
 
--- a/packages/builder/src/actions/roundApplication.ts
+++ b/packages/builder/src/actions/roundApplication.ts
@@ -233,7 +233,6 @@ export const submitApplication =
 
       deterministicApplication = objectToDeterministicJSON(application as any);
     } catch (error) {
-      console.error(error);
       dispatchAndLogApplicationError(
         dispatch,
         roundAddress,
--- a/packages/builder/src/components/application/Form.tsx
+++ b/packages/builder/src/components/application/Form.tsx
@@ -3,47 +3,31 @@ import { datadogRum } from "@datadog/browser-rum";
 import { ExclamationCircleIcon } from "@heroicons/react/20/solid";
 import {
   ExclamationTriangleIcon,
-  EyeIcon,
-  GlobeAltIcon,
   InformationCircleIcon,
 } from "@heroicons/react/24/solid";
-import { ChainId, renderToHTML } from "common";
 import { Fragment, useEffect, useState } from "react";
 import { shallowEqual, useDispatch, useSelector } from "react-redux";
 import { Link } from "react-router-dom";
-import { useEnsName, useNetwork } from "wagmi";
+import { useNetwork } from "wagmi";
 import { ValidationError } from "yup";
 import { fetchProjectApplicationInRound } from "../../actions/projects";
 import { resetApplicationError } from "../../actions/roundApplication";
-import {
-  DefaultProjectBanner,
-  DefaultProjectLogo,
-  GithubLogo,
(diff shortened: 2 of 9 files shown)


packages/builder/src/actions/roundApplication.ts
```diff
@@ -127,6 +127,7 @@ const dispatchAndLogApplicationError = (
   datadogLogs.logger.error(error, {
     roundAddress,
   });
+  debugger;
```
vacekj, line 130:
```suggestion
```

packages/builder/src/actions/roundApplication.ts
```diff
@@ -233,6 +233,7 @@ export const submitApplication =
 
       deterministicApplication = objectToDeterministicJSON(application as any);
     } catch (error) {
+      console.error(error);
```
vacekj, line 236:
```suggestion
```

packages/builder/src/utils/RoundApplicationBuilder.ts
```diff
@@ -27,7 +27,7 @@ export default class RoundApplicationBuilder {
     this.ram = ram;
     this.roundAddress = roundAddress;
     this.chainName = chainName;
-
+    debugger;
```
vacekj, line 30:
```suggestion
```

bhargavaparoksham (approved):

packages/builder/src/components/application/Form.tsx
```diff
@@ -687,6 +333,9 @@ export default function Form({
   const needsProject = schema.questions.find((q) => q.type === "project");
   const now = new Date().getTime() / 1000;
 
+  console.log(now, round.applicationsEndTime);
+  console.log(round.applicationsEndTime > now);
+
```
vacekj, line 338:
```suggestion
```

packages/builder/src/components/application/FullPreview.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { EyeIcon } from "@heroicons/react/24/solid";
+import { ChainId, renderToHTML } from "common";
+import { Metadata, RoundApplicationQuestion } from "../../types";
+import { RoundApplicationAnswers } from "../../types/roundApplication";
+import { DefaultProjectBanner, DefaultProjectLogo } from "../../assets";
+import Button, { ButtonVariants } from "../base/Button";
+import { AboutProject } from "./AboutProject";
+import { ProjectTitle } from "./ProjectTitle";
+
+export function FullPreview(props: {
```
bard, line 11: We should state return types, otherwise a function could return something that is not React-renderable data and cause a runtime error. This might sound unlikely, but it's enough to forget a `return` inside a component that conditionally renders based on `if` or `switch` for things to explode (`undefined`is one such invalid data type).

I usually do:

```tsx
export const Foo: React.FC<{ bar: string, baz: number}> = ({ bar, baz}) => {
  // ...
}
```

That takes care of both argument and return type. Not sure how to write the same without `const`, though.

packages/builder/src/components/application/FullPreview.tsx
```diff
@@ -0,0 +1,177 @@
...
+  project: Metadata;
+  answers: RoundApplicationAnswers;
+  questions: RoundApplicationQuestion[];
+  handleSubmitApplication: Function;
```
bard, line 15: Can we tighten this up? E.g. `(submitData: ...) => Promise<void>`

bard (approved):
I left a couple of comments but they're not related to anything introduced by this PR, so please disregard, we can and should address those things separately.

Were there any non-mechanical changes other than flipping the `> now` / `< now` condition? Are there any tests exercising that particular scenario?

packages/builder/src/components/application/FullPreview.tsx
vacekj: Good catch, I actually saw an instance of this in our codebase, where someone forgot to put return in a conditional (maybe they were thinking about Rust-style returns :D ), luckily there it just results in a part of the component not rendering I think. 

I'll resolve.

vacekj: Actually, i just now realized - undefined is a valid type tha React can render. if you look at the ReactNode type: https://reactnative.dev/docs/react-node (applies to react-dom too), undefined and null are just ignored. So I think it's safe to just type the return value as a React.ReactNode to ensure we don't accidentally return a function or something that react can't render. What do you think?

bard: Yes, you're right, `undefined` has become ok in React 18 and that's what the project is using.

I think the best would still be to type the function itself as `React.FC<PropType>`, as that will always be valid regardless of React version, but if not possible, adding types manually is still better than nothing.

vacekj:
> I left a couple of comments but they're not related to anything introduced by this PR, so please disregard, we can and should address those things separately.
> 
> Were there any non-mechanical changes other than flipping the `> now` / `< now` condition? Are there any tests exercising that particular scenario?

No other non-mechanical changes, tests are all passing so I assume there was nothing that tested this condition directly, but our testing strategy rn is not that great, because we mostly have huge components with lots of implicit dependencies and then have to mock half the application to make it testable. This PR sets the stage for an example of colocated, small, component-only tests to form the base of a better testing approach.

bhargavaparoksham (approved):

Issue #2084, opened 2023-07-24
Manager: funding amount not showing up in UI even though contract has been funded 


michellema1208:
**Overview**

**Problem**: We're getting reports from partners that after they funded their round contracts, the funding amount is not showing up on the Fund Contracts tab in Manager UI. See links below: 

Mantle Grants: 
* https://etherscan.io/address/0xf1c021df6dc6b2dc2e5a837cdfddc2f42503233b#tokentxns 
* Screenshot: https://share.getcloudapp.com/d5uDy5Oj 

Citizens Round: 
* https://optimistic.etherscan.io/tx/0xc0d110fa576ca891aaeec3a5879c03455397296f369ae4b677d12d26c0990e71 
* Screenshot: 
<img width="568" alt="Screen Shot 2023-07-24 at 4 23 10 PM" src="https://github.com/gitcoinco/grants-stack/assets/26017180/b0d6cf0b-c769-43e7-a405-e5fc8230e61a">

**Steps To Reproduce**

**Observed  behavior**
A clear and concise description of what the observed behavior of the system.

**Expected behavior**
A clear and concise description of what you expected the system to do.

**Screenshots**

**Additional context**
Add any other context about the problem here that will assist in triaging the bug. I.e. Browser information, unique identifiers, etc.

meg-fsc:
Note: follow ups in GGOG telegram channel, this is blocking round managers from finalizing rounds. Need to pick up on Monday cc @michellema1208 @gravityblast

meg-fsc:
Noting that this seems to be preventing managers from distributing funds, not just a UI bug.

To reproduce go to: https://rmgitcoin.on.fleek.co/#/round/0x859FaeAa266BA13bd1E72eB6dd7A223902d1adFE?debug=true

Notice that contract is fully funded
Notice that round results are finalized
Notice that grantees are unable to be paid (SS attached

![Image](https://github.com/gitcoinco/grants-stack/assets/72808678/c0ad3e73-2fd6-4c79-b87e-79d94e3ce3fb)

)

PR #2105, opened 2023-08-02
fix useBalance params to show funds in contract in Explorer

gravityblast:
fixes: #2084


vacekj (approved):

vacekj (approved):

packages/round-manager/src/features/round/FundContract.tsx
```diff
@@ -116,7 +116,10 @@ export default function FundContract(props: {
     data: balanceData,
     isError: isBalanceError,
     isLoading: isBalanceLoading,
-  } = useBalance(tokenDetail);
+  } = useBalance({
+    address: tokenDetail.addressOrName as Address,
+    token: tokenDetail.token,
+  });
```
bard, line 122: What is the difference between the old and new code? They seem equivalent.

Also, I realize it's like this in several areas already, but this would be safer:

```ts
    address: assertAddress(tokenDetail.addressOrName),
    token: tokenDetail.token,
  });

// ...


function assertAddress(value: unknown): Address {
  if (typeof value === "string" && value.slice(0, 2) === "0x") {
    return value as Address;
  } else {
    throw new Error(`Value is not an address (${value})`);
  }
}
```

gravityblast: > What is the difference between the old and new code? They seem equivalent.

the new wagmi version needs `{ address: ...., token: ...}`, the previous one had `addressOrName` instead of `address`

bard: Duh, apparently I'm code-blind today. Then I guess they could be renamed in the `tokenDetail` and `tokenDetailUser` objects themselves, since they're not used elsewhere, or they could be got rid of entirely?

bard (approved):

michellema1208 (approved):
Confirming that it looks like it's working for one of the rounds that reported this issue! 

Link: https://bafybeifvlnsylvnsgowfxuxj6efbe3qzunxutxukhbyrfxie7m7edzkedq.on.fleek.co/#/round/0xf1c021df6dc6b2dc2e5a837cdfddc2f42503233b?debug=true

Screen Shot 2023-08-02 at 9 21 58 AM

packages/round-manager/src/features/round/FundContract.tsx
gravityblast: I also missed a few things, I thought the object was created in other places but most of them are just created for that call, I'm checking them all now

gravityblast: @bard I updated it using your function

gravityblast:
@michellema1208 since I made some changes can you check it again please?

packages/round-manager/src/features/round/FundContract.tsx
bard: > @bard good point I'm going to check it. would you still add the assert in that case?

Yes, I would use the runtime assert in place of every `as Address` we have.

bard (approved):

PR #2159, opened 2023-08-09
[round-manager] - 7. Unit tests

nicosampler:
This PR adds unit tests for the Direct round feature.

Unfortunately, I haven't had the time to have a full coverage but the main uses cases were covered. 

I would like to have the time to add 2 more UT for:
- Creating a direct round (form and the call to action)
- usePayout hook.


bard:
@nicosampler if this is work in progress, please mark it as draft

nicosampler:
hi @bard I tried to do it, but it seems like I don't have permission. Anyways, it's ready now.

PR #2389, opened 2023-09-26
Add data-testid attributes to some form elements

giliomeejg:
##### Description

Add some attributes to buttons and links in order to identify them for the purposes of setting up a conversion funnel and tracking things using Mixpanel.

##### Testing

Small changes that do not influence the logic.

### DIFF
--- a/packages/builder/src/components/base/VerificationForm.tsx
+++ b/packages/builder/src/components/base/VerificationForm.tsx
@@ -185,10 +185,21 @@ export default function VerificationForm({
         <Button
           variant={ButtonVariants.outline}
           onClick={() => setVerifying(ProjectFormStatus.Metadata)}
+          dataTestId="project-create-social-back"
         >
           Back
         </Button>
-        <Button variant={ButtonVariants.primary} onClick={saveAndPreview}>
+        <Button
+          variant={ButtonVariants.primary}
+          onClick={saveAndPreview}
+          dataTestId={
+            !props.formMetaData.userGithub &&
+            !props.formMetaData.projectGithub &&
+            +!props.formMetaData.projectTwitter
+              ? "project-create-social-skip"
+              : "project-create-social-next"
+          }
+        >
           {!props.formMetaData.userGithub &&
           !props.formMetaData.projectGithub &&
           !props.formMetaData.projectTwitter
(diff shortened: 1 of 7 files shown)


melissa-neira (dismissed):

vacekj (dismissed):

gravityblast (changes requested):
thank you @giliomeejg for the contribution! I see some data-testid attributes added but not used in tests, what's the goal of this PR? 
we also talked about avoiding adding these kind of attributes only for tests and test those components in other way, let's talk about it! /cc @bard @vacekj @bhargavaparoksham @boudra

gravityblast:
@giliomeejg if it's only for Mixpanel, shall we decide an other way to do it and keep the same convention for future changes?

boudra:
> @giliomeejg if it's only for Mixpanel, shall we decide an other way to do it and keep the same convention for future changes?

I would suggest using the attribute `data-track-event`, whis way we can separate tracking attributes from test IDs and not delete all your tracked events by mistake!

bard:
I don't see a problem with using `data-testid` for testing, but if used for tracking, then intention and name don't align, and that puts people unfamiliar with the code at higher risk of breaking something.

giliomeejg:
Thanks for all the comments.

My primary concern is to setup Mixpanel, for which I need some descriptors on the buttons.

The idea behind data-testid is that the same identifiers could be used for both testing and e.g. setting up Mixpanel, but happy to rename to anything else.  Perhaps data-identifier would make it more generic?

gravityblast:
if we can add something like `data-track-event` like @boudra said, I feel it will be clear to everyone we use it to track events

giliomeejg:
Renamed them all to 'data-track-event'

packages/builder/src/components/base/VerificationForm.tsx
```diff
@@ -185,10 +185,21 @@ export default function VerificationForm({
...
           Back
         </Button>
-        <Button variant={ButtonVariants.primary} onClick={saveAndPreview}>
+        <Button
+          variant={ButtonVariants.primary}
+          onClick={saveAndPreview}
+          dataTrackEvent={
+            !props.formMetaData.userGithub &&
+            !props.formMetaData.projectGithub &&
+            +!props.formMetaData.projectTwitter
```
gravityblast, line 198: we forgot a `+` here

giliomeejg: Tnx, sorted.

giliomeejg:
> if we can add something like `data-track-event` like @boudra said, I feel it will be clear to everyone we use it to track events

Should be good to go now.

vacekj (approved):

PR #2413, opened 2023-10-03
make sure pinata requests don't fail if base URL ends with a '/'

gravityblast:


packages/builder/src/services/pinata.ts
```diff
@@ -27,7 +27,11 @@ export default class PinataClient {
   }
 
   fileURL(cid: string) {
-    return `${GATEWAY}/ipfs/${cid}`;
+    // if GATEWAY is true it throws an exception in the constructor
+    const baseURL = new URL(GATEWAY!);
```
bard, line 31: The non-null assertion would not be needed here if `this.gateway` were used, since that has already been validated as being non-null in the constructor.

packages/builder/src/services/pinata.ts
```diff
@@ -27,7 +27,11 @@ export default class PinataClient {
...
+    // using the baseURL makes sure that we don't have a double '/'
+    // for the path in case GATEWAY ends with one
+    return new URL(`/ipfs/${cid}`, baseURL).toString();
```
bard, line 34: If the issue is that the user/operator can enter a gateway with a trailing slash, and the trailing slash shouldn't be allowed, then rather than accepting invalid data and fixing it at the points of usage (which can be many), I'd rather either reject or correct it at the point where it's read (which is one).

bard (approved):
Approved by default — what to pick and what to leave is up to the author.

packages/builder/src/services/pinata.ts
vacekj: I'd correct it here and just parse it into a format that we accept (remove the trailing slash if present)

packages/builder/src/services/pinata.ts
gravityblast: @bard nice catch thank you! in fact I added the comment but I then used the other variable instead of the instance one thank you!

packages/builder/src/services/pinata.ts
gravityblast: @bard @vacekj it's in the env file, so passing the baseURL as second parameter is exactly to remove the trailing slash. do you mean this way is not clear? it's done on purpose in the URL api.

packages/builder/src/services/pinata.ts
gravityblast: I updated it using `const baseURL = new URL(this.gateway);`

packages/builder/src/services/pinata.ts
bard: I meant that, if the trailing slash is a problem, you can ensure it's absent in the constructor when the variable is first read, and not worry about it at any point of usage, as in:

```diff

  constructor() {
    if (JWT === undefined || GATEWAY === undefined) {
      throw new Error(
        "remember to set the REACT_APP_PINATA_JWT and REACT_APP_PINATA_GATEWAY env vars"
      );
    }
    this.jwt = JWT!;
-   this.gateway = GATEWAY!;
+   this.gateway = GATEWAY.replace(/\/$/, '');
+// or even better, throws not just if GATEWAY is unset but also if it's not a valid URL:
+// this.gateway = z.string().url().parse(process.env.GATEWAY).replace(/\/$/, '');
    this.pinJSONToIPFSURL = PIN_JSON_TO_IPFS_URL;
    this.pinFileToIPFSURL = PIN_FILE_TO_IPFS_URL;
  }

  fileURL(cid: string) {
-   // if GATEWAY is true it throws an exception in the constructor
-   const baseURL = new URL(this.gateway);
-   // using the baseURL makes sure that we don't have a double '/'
-   // for the path in case GATEWAY ends with one
-   return new URL(`/ipfs/${cid}`, baseURL).toString();
+   return `${this.gateway}/ipfs/${cid}`
  }
```

gravityblast: I felt using the URL API was more elegant than a regexp in the same way we use `path.join(base, path)` in other languages, but I'm happy to use `replace` if it feels more clear.

> base Optional
A string representing the base URL to use in cases where url is a relative URL. If not specified, it defaults to undefined.

gravityblast: I pushed the change, and I'm happy to add `zod` when we want to validate all vars

bard: It was more about validating / parsing at the point of read than at the point of use.

bard (approved):

vacekj (approved):

Issue #2382, opened 2023-09-25
Explorer: update branding and page layout 


michellema1208:
**User Story:**
As a donor 
I want to have a delightful experience browsing Explorer 
So that I stay and fund more projects and rounds 

**Acceptance Criteria**
GIVEN that I am on the Explorer Homepage 
WHEN I look at the page 
THEN I see updated branding on the page 
AND I see a Home tab which displays rounds categorized by groups 
AND an "Explore Rounds" tab which displays all rounds with sort/filter functionality 

**Product & Design Links:**
* [Figma](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=300-14464&mode=design&t=os8gMWwBouzyFg6y-0)
* [Sample staging environment to see existing homepage](https://explorer-staging-git-bug-2394-grants-stack.vercel.app/#/)

**Tech Details:**
* From @gravityblast: for the round groupings with different card sizes, we can build those basic components as a part of this ticket

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

carlbarrdahl:
As mentioned in the call, it would make sense to create a layout component that contains all the shared parts between the different routes.

Right now there are some duplications in these routes, for example, this code to render the navbar and the footer is in each of these files.
```tsx
<Navbar />
...
 <div className="relative top-16 lg:mx-20 h-screen px-4 py-7"> <-- This is different in some of the pages
    <main>

...
<div className="my-11">
  <Footer />
</div>
```

We can create a shared layout component that renders the Navbar, Footer, padding, max-width, etc. This way we only need to make adjustments in one file and ensure consistency across pages.

PR #2463, opened 2023-10-16
[2383] Round card redesign

### LINKED ISSUE #2383: Explorer: update round card design and add labels
**User Story:**
As a donor 
I want to have a delightful experience browsing Explorer 
So that I stay and fund more projects and rounds 

**Acceptance Criteria**
GIVEN that I am on the Explorer Homepage 
WHEN I look at any round card 
THEN I see a labels for the type of grant round it is (QF or Direct Grants) 
AND I see a label for total project count
AND I see the icon for the network the round is deployed on 

GIVEN that the round is a **QF round** 
WHEN I look at the round card 
THEN I see an additional label for for matching funds available 

GIVEN that the round is **still accepting applications** 
WHEN I look at the round card  
THEN I see an apply label 

GIVEN that the round is **currently active** 
WHEN I look at the round card 
THEN I see "countdown copy" on the card that tells me how many days are left in the round 

**Product & Design Links:**
* [Figma](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=300-14464&mode=design&t=os8gMWwBouzyFg6y-0)
* [Sample staging environment to see existing homepage](https://explorer-staging-git-bug-2394-grants-stack.vercel.app/#/)

**Tech Details:**
* This ticket includes updating the round cards to match current designs, which includes rounded edges 

**Open Questions:**
*

**Notes/Assumptions:**

carlbarrdahl:
- Closes #2383 

### Todo
- [x] Apply badge
- [x]  Days left to apply
- [x] Different colors for different grants (add yellow for "direct grant")

image

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


packages/grant-explorer/src/features/discovery/RoundCard.tsx
```diff
@@ -22,86 +23,143 @@ type RoundCardProps = {
...
+          <RoundBanner roundId={id} />
+          {daysLeftToApply > 0 ? (
+            <Badge
+              color="green"
+              rounded="full"
+              className="absolute top-3 right-3"
+            >
+              Apply!
+            </Badge>
+          ) : null}
```
vacekj, line 79: `condition ? <element/> : null` can be simplified as `condition && <element/>`

```suggestion
          {daysLeftToApply > 0 && 
            <Badge
              color="green"
              rounded="full"
              className="absolute top-3 right-3"
            >
              Apply!
            </Badge>}
```

packages/grant-explorer/src/features/discovery/RoundCard.tsx
```diff
@@ -22,86 +23,143 @@ type RoundCardProps = {
...
+          <CardTitle
+            data-testid="round-name"
+            className="absolute bottom-3 left-3"
+          >
+            {roundMetadata?.name}
```
vacekj, line 84: this will display `undefined` if roundMetadata is falsy. If roundmetadata is always defined, remove the optional chaining, or if it can truly be undefined, add a simple skeletontext for loading state: https://chakra-ui.com/docs/components/skeleton/usage#circle-and-text-skeleton

vacekj:
@carlbarrdahl great work, left some comments, otherwise lgtm.

packages/grant-explorer/src/features/discovery/RoundCard.tsx
carlbarrdahl: Round metadata should always be there since it is fetched in the same function as the rest of the round data.
The typescript type types it as optional (I guess if the fetch fails). To me, it doesn't make sense with a Skeleton loading state because the RoundCard doesn't have a loading state.

Undefined should resolve into empty.

carlbarrdahl:
> @carlbarrdahl great work, left some comments, otherwise lgtm.

Excellent, tysm!

melissa-neira (changes requested):
hey! overall looking good, but wanted to call out a few things:
- there's a new image library that we're going to use for round cards (these will be randomized and can be found [here](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=300-15373&mode=design&t=t0DqvyYhlvyMgAu8-4))
- the round card image should have a blur to it as well, as seen here:
Screenshot 2023-10-17 at 09 53 33
- round titles should be truncated like this - see figma [here](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=159-24094&mode=design&t=t0DqvyYhlvyMgAu8-4)
Screenshot 2023-10-17 at 09 56 19
- I'm seeing some spacing & alignment issues here in the cards, especially toward the bottom. also some time issues (see second card in screenshot)
Screenshot 2023-10-17 at 09 51 06

carlbarrdahl:
> hey! overall looking good, but wanted to call out a few things:
> 
> * there's a new image library that we're going to use for round cards (these will be randomized and can be found [here]

Thank you @melissa-neira!
I fixed all the issues except adding the new banners (couldn't figure out how to save them without the rounded border).

Noticed something else. The match amount pushes the chain icon into a small size if too long text string. Possible solutions:
- Shorten the match amount numbers (eg. 22k ETH match amount instead of 22,232)
- Shorten "match amount"
- Truncate
- Overflow scroll
- Collapse into 2 rows

Any suggestions?

melissa-neira:
@carlbarrdahl thank you! I'll review the changes in a bit! as for the image library, I'll work on getting those images without the border radius. 

thank you for thinking out those solutions - I think shortening the match amount numbers and match amount to just "match" could work here! let's go with those, wdyt?

carlbarrdahl:
> @carlbarrdahl thank you! I'll review the changes in a bit! as for the image library, I'll work on getting those images without the border radius.
> 
> thank you for thinking out those solutions - I think shortening the match amount numbers and match amount to just "match" could work here! let's go with those, wdyt?

I've updated the banners and shortened the match amount to just "match". I do think we need to make some other adjustments.
image

I suggest we look more into this in the #2382 ticket. We might want to change to a 2-column grid on lower screen sizes like in the screenshot above.

carlbarrdahl:
> @carlbarrdahl yes, I agree - let's switch to a 2 column grid for that screen size. [here](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=300-15406&mode=design&t=ZRXfoPlxE52Sxymc-4) are mobile designs where we use 1 column

Great! Will implement that in the other ticket.

Can we merge this PR?

melissa-neira:
@carlbarrdahl yes, just one little thing - I see that the round titles are being cut off if they have letters that hang lower like g or y. 
here's a screenshot:
Screenshot 2023-10-18 at 17 37 19

everything else looks great!!

carlbarrdahl:
> @carlbarrdahl yes, just one little thing - I see that the round titles are being cut off if they have letters that hang lower like g or y. here's a screenshot: Screenshot 2023-10-18 at 17 37 19
> 
> everything else looks great!!

Good catch!

Added some padding to the title to resolve the issue:
image

melissa-neira (dismissed):
everything looks good to me! 😎

melissa-neira (approved):

packages/grant-explorer/src/features/api/utils.ts
```diff
@@ -621,6 +621,10 @@ export const pinToIPFS = (obj: IPFSObject) => {
 };
 
 export const getDaysLeft = (epochTime: number) => {
+  // Invalid date
+  if (epochTime > Number.MAX_SAFE_INTEGER) {
+    return 0;
```
boudra, line 626: @carlbarrdahl when is this expected to happen? I feel like this should be an actual error being thrown, but not sure when it happens and if we can recover

boudra (approved):

packages/grant-explorer/src/features/api/utils.ts
carlbarrdahl: > @carlbarrdahl when is this expected to happen? I feel like this should be an actual error being thrown, but not sure when it happens and if we can recover

Some rounds have timestamps set to "864000000000000000000" or something like that. It's causing rendering errors. See the second card:

PR #2465, opened 2023-10-16
chore: remove REACT_APP_DIRECT_GRANT_ENABLED and make it the default

### LINKED ISSUE #2425: remove the REACT_APP_DIRECT_GRANT_ENABLED and enable it by default

vacekj:
## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #2425


bhargavaparoksham (approved):

packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -56,9 +54,7 @@ export default function ViewProgram() {
   const [hasAccess, setHasAccess] = useState(true);
   const [grantType, setGrantType] = useState<
     "quadraticFunding" | "directGrant" | undefined
-  >(
-    process.env.REACT_APP_DIRECT_GRANT_ENABLED ? undefined : "quadraticFunding"
-  );
+  >();
```
bard, line 57: If there should be no default, explicitly setting this as `undefined` would make that clear (though `null` would be a better choice for "intentionally empty").

packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -302,50 +298,48 @@ export default function ViewProgram() {
...
+              } bg-white p-6 cursor-pointer`}
+              style={{
+                boxShadow:
                   grantType === "directGrant"
-                    ? "border-violet-400"
-                    : "border-grey-100"
-                } bg-white p-6 cursor-pointer`}
-                style={{
-                  boxShadow:
+                    ? " 0px 10px 15px -3px #0000001A"
```
bard, line 311: Is there any tailwind class that would allow avoiding the custom style?

packages/round-manager/src/features/program/ViewProgramPage.tsx
```diff
@@ -360,20 +354,19 @@ export default function ViewProgram() {
...
-                      className="h-4 w-4 inline-flex -translate-y-0.5"
-                      aria-hidden="true"
-                    />
-                    &nbsp;Create round
-                  </Button>
-                </Link>
-              )}
+            {grantType === "directGrant" && (
+              <Link
+                to={`/round/create?programId=${programToRender?.id}&roundCategory=direct`}
```
bard, line 359: If `programToRender` is undefined, will the link be valid?

Thinking aloud / not in scope here: would be nice to have routes checked at compile time with something like:

```tsx
<Link to={routes.round.create({ programId: programToRender.id, roundCategory: 'direct' })
```

bard (approved):

boudra (approved):

packages/round-manager/src/features/program/ViewProgramPage.tsx
vacekj: shadow-lg is the closest, i'll see if it looks the same.

packages/round-manager/src/features/program/ViewProgramPage.tsx
vacekj: Static route checking is on the roadmap for sure. I'll see if we can do more check around programtorender here.

packages/round-manager/src/features/program/ViewProgramPage.tsx
vacekj: brb drafting an ecma proposal for `Option<T>` in js..

on a more serious note, i'll change this to null

packages/round-manager/src/features/program/ViewProgramPage.tsx
vacekj: it's the same, fixed.

packages/round-manager/src/features/program/ViewProgramPage.tsx
vacekj: It's very convoluted and not aparent, but this part of the UI only renders when programToRender is not null, via a series of state-that-should-be-a-variable checks. So this is a non-issue, but it for sure needs to be cleaned up and checked at compile-time, not review-time.

bard (approved):

boudra (approved):

PR #2466, opened 2023-10-16
fix: don't display testnet rounds in production on expl. homepage

### LINKED ISSUE #2082: Explorer: Remove PGN testnet rounds from Explorer homepage on production (no testnets should be supported)
**Overview**
- https://explorer.gitcoin.co/#/round/58008/0x7530937a7c0f74201d78930d31f738868e5b1465
- https://explorer.gitcoin.co/#/round/58008/0xbe0cbcb187856a183bdb69947cbdb79b0bf4d69d

**Steps To Reproduce**

**Observed  behavior**

**Expected behavior**

**Screenshots**

**Additional context**

vacekj:
## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #2082


packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -3,19 +3,22 @@ import { RoundMetadata } from "./round";
...
 import { ethers } from "ethers";
+import { allChains } from "../../app/chainConfig";
 
 interface GetRoundOverviewResult {
   data: {
     rounds: RoundOverview[];
   };
 }
 
+/*TODO: what is this */
```
bard, line 14: Would be nice to have an explanation here before merging. Who knows about this?

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -93,31 +96,29 @@ async function fetchRoundsByTimestamp(
...
 }
 
-function getActiveChainIds() {
-  const activeChainIds: string[] = [];
-  const isProduction = process.env.REACT_APP_ENV === "production";
+/**
+ * Filters out testnet chainIds in production, otherwise returns all active chainIds
+ * */
+const getActiveChainIds = () =>
+  Object.keys(ChainId).filter((chainId) => {
```
bard, line 103: I see `allChains` is defined elsewhere as:

```ts
export const allChains: Chain[] =
  process.env.REACT_APP_ENV === "development"
    ? [...testnetChains(), ...mainnetChains()]
    : [...mainnetChains()];
```

So there seems to be a test/non-testnet selection logic already. Does this change need to take that into account?

Unrelated to this PR: would it make sense to try and derive `ChainId` from `allChains` instead of having the id's defined both as enum and as keys in `allChains`? (This reminds me of the Jimmy Koppel's color example about hidden coupling.)

bard (approved):

packages/grant-explorer/src/features/api/rounds.ts
vacekj: true - i didn't realize allChains already has this functionality - this function is therefore superfluous and we can just use the allChains array directly.

second point - it could, but I believe this wouldn't work at comptime, meaning that we would lose the static checking, and ChainId would become just a number, because we don't know the contents of the allChains object at comptime. but maybe `as const` could help here.

bard: Good point. It _might_ become a number union, which _might_ still be ok, but if it became just a `number`, that would be a step back.

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -96,32 +96,13 @@ async function fetchRoundsByTimestamp(
...
-    if (
-      process.env.REACT_APP_ENV === "production" &&
-      chainDefinition?.testnet
-    ) {
-      return false;
-    }
-
-    return true;
-  });
+const activeChainIds = allChains
```
bard, line 99: I actually liked that you had a function here. If you have two pieces of state, they can get out of sync. If you have one piece of state and one function that derives a value from it, pay a tiny computational price for the reassurance that the value you get is up to date. Not a bad deal.

bard (approved):

packages/grant-explorer/src/features/api/rounds.ts
boudra: @bhargavaparoksham I see you were the one who added it, if I remember correctly this was a temporary measure to fix a production issue, would this be safe to delete now?

vacekj: All these rounds are old and won't display because the donation period has passed, so afaik it's safe to delete them. i added the names of the rounds to the pr.

melissa-neira (dismissed):

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -93,34 +94,12 @@ async function fetchRoundsByTimestamp(
...
-      ].includes(ChainId[chainId as keyof typeof ChainId])
-    ) {
-      continue;
-    }
-    activeChainIds.push(chainId.toString());
-  }
-  return activeChainIds;
-}
+const activeChainIds = () =>
+  allChains.map((chain) => chain.id).map((chainId) => chainId.toString());
```
boudra, line 98:
```suggestion
  allChains.map((chain) => chain.id.toString());
```
more concise and less confusing imo

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -93,34 +94,12 @@ async function fetchRoundsByTimestamp(
...
-        ChainId.FANTOM_TESTNET_CHAIN_ID,
-      ].includes(ChainId[chainId as keyof typeof ChainId])
-    ) {
-      continue;
-    }
-    activeChainIds.push(chainId.toString());
-  }
-  return activeChainIds;
-}
+const activeChainIds = () =>
```
bard, line 97: `getXyz` would make it apparent when seeing at call site that it's a function rather than a value.

bard (approved):

packages/grant-explorer/src/features/api/rounds.ts
vacekj: Done.

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -93,34 +93,16 @@ async function fetchRoundsByTimestamp(
...
-    }
-    activeChainIds.push(chainId.toString());
-  }
-  return activeChainIds;
-}
+const getActiveChainIds = (): ChainId[] =>
+  allChains
+    .map((chain) => chain.id)
+    .map(tryParseChainIdToEnum)
+    .filter((chainId): chainId is ChainId => chainId !== null)
```
bard, line 100: `isPresent` from `ts-is-present` might help here (narrowing included)

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -65,7 +65,7 @@ async function fetchRoundsByTimestamp(
         round.roundMetaPtr.pointer
       );
       round.roundMetadata = roundMetadata;
-      round.chainId = chainId;
+      round.chainId = chainId.toFixed(0).toString();
```
bard, line 68: Two questions pop to mind when seeing this: 1) "is `chainId` a float that it needs a `toFixed`?" and 2) what is the purpose of `toString()` when `toFixed()` returns a string?

packages/common/src/chains.ts
```diff
@@ -17,6 +17,22 @@ export enum ChainId {
   POLYGON_MUMBAI = 80001,
 }
 
+/**
+ * Attempts to parse a numerical or string chainId to the Enum.
+ * returns null if the chainid is invalid */
+export function tryParseChainIdToEnum(chainId: string | number) {
```
bard, line 23: Declaring the return type of the function would guard against accidentally returning `undefined` or something likewise unintended.

bard (dismissed):

packages/grant-explorer/src/features/api/rounds.ts
vacekj: the ToString is indeed superfluous here. this is a workaround for a behaviour of TS enums where if you call toString() on an instance of it, it will output the string value (OPTIMISM_MAINNET) instead of the stringified number ("10"). this is a common gotcha, started a thread in eng. to see how we can improve this.

vacekj: removed the toString

bard (approved):

packages/common/src/chains.ts
```diff
@@ -17,6 +17,24 @@ export enum ChainId {
...
+ * Attempts to parse a numerical or string chainId to the Enum.
+ * returns null if the chainid is invalid */
+export function tryParseChainIdToEnum(
+  chainId: string | number
+): ChainId | null {
+  const chains = Object.keys(ChainId)
+    .map(Number)
+    .filter((item) => {
+      return !isNaN(item);
+    });
```
gravityblast, line 30: it would make sense to parse it at runtime if we have runtime values, but I see we use it with `allChains`, can't they be directly the right type so it's done at compile time?

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -65,7 +66,7 @@ async function fetchRoundsByTimestamp(
         round.roundMetaPtr.pointer
       );
       round.roundMetadata = roundMetadata;
-      round.chainId = chainId;
+      round.chainId = chainId.toFixed(0);
```
gravityblast, line 69: what's the scenario where we need `toFixed` here?

packages/common/src/chains.ts
vacekj: the id on wagmi's `Chain` type is number. this is not guaranteed to be a chain that we support. that's why we parse it.

packages/grant-explorer/src/features/api/rounds.ts
vacekj: image

gravityblast (approved):

PR #2470, opened 2023-10-18
localstorage & urlencoding grants curation

vacekj:
continuation of https://github.com/gitcoinco/grants-stack/pull/1786

merged with latest main

## PR checklist

- [ ] Doesn't disable lints.


packages/grant-explorer/src/features/api/round.ts
```diff
@@ -67,6 +67,77 @@ export type RoundProject = {
...
+      status: convertStatus(project.status),
+    };
+  });
+
+  const fetchApprovedProjectMetadata: Promise<Project>[] = approvedProjects.map(
+    (project: RoundProjectResult) =>
+      fetchMetadataAndMapProject(project, chainId)
+  );
+
+  const projectsMetadata = await Promise.all(fetchApprovedProjectMetadata);
```
carlbarrdahl, line 122: Use SWR for proper state handling and loading states.

Also, if we fetch the metadata in a separate hook we can fetch them in the component when it's rendered. This will improve loading times.

packages/grant-explorer/src/features/collection/AllCollections.tsx
```diff
@@ -0,0 +1,221 @@
...
+        enc: encodeURIComponent(encodeCollection(JSON.stringify(i))),
+      }))
+    );
+    setLoadingCollections(false);
+  }, []);
+
+  useEffect(() => {
+    // load all the collections for this round here
+    loadAllCollections(`${chainId}:${roundId}`);
+  }, [chainId, roundId, projects, loadAllCollections]);
```
carlbarrdahl, line 189: We can simplify a lot of this with useSWR. That way we don't need to use `useState`, `useEffect`, and `useCallback`. We also get loading and errors states.

packages/grant-explorer/src/features/collection/useCollections.tsx
```diff
@@ -0,0 +1,46 @@
...
+        createNewCollection={() => setCurrentModal(CurrentModal.NEWCOLLECTIONS)}
+        closeModal={closeModal}
+      />
+      <NewCollectionsModal
+        project={activeProject}
+        open={currentModal === CurrentModal.NEWCOLLECTIONS}
+        closeModal={closeModal}
+      />
+    </>
+  );
```
carlbarrdahl, line 41: Haven't seen this pattern before, is it common to render components from a hook? Why not put this in a component and conditionally render if currentModal is set?

PR #2477, opened 2023-10-19
hotfix: update queries for graph

codenamejason:
## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes [Discord Chat Issue](https://discord.com/channels/562828676480237578/1164618197870334012)


vacekj (approved):

gravityblast:
closing it since it looks like the problem is somewhere else and with the previous query everything already works correctly

PR #2483, opened 2023-10-23
fix: make project titles visible

boudra:
Project titles were not visible because they were white. closes https://github.com/gitcoinco/grants-stack/issues/2482

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


carlbarrdahl (approved):
:+1:

michellema1208 (approved):

bard (approved):

PR #2504, opened 2023-10-24
2503 fix payout strategy name causing crash

carlbarrdahl:
## PR checklist

- [ ] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber

### DIFF
--- a/packages/grant-explorer/src/features/discovery/RoundCard.tsx
+++ b/packages/grant-explorer/src/features/discovery/RoundCard.tsx
@@ -108,7 +108,7 @@ const RoundCard = ({ round }: RoundCardProps) => {
               isValidRoundEndTime={isValidRoundEndTime}
             />
 
-            <RoundBadge strategyName={payoutStrategy.strategyName} />
+            <RoundBadge strategyName={payoutStrategy?.strategyName} />
           </div>
           <div className="border-t" />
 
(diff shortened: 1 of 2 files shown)


boudra (approved):

vacekj (approved):

packages/grant-explorer/src/features/discovery/RoundCard.tsx
```diff
@@ -108,7 +108,7 @@ const RoundCard = ({ round }: RoundCardProps) => {
               isValidRoundEndTime={isValidRoundEndTime}
             />
 
-            <RoundBadge strategyName={payoutStrategy.strategyName} />
+            <RoundBadge strategyName={payoutStrategy?.strategyName} />
```
gravityblast, line 111: @carlbarrdahl do we have empty payout strategy name for old rounds or in other cases? 
does this show an empty badge? in that case shall we completely remove it when we don't have text inside?

carlbarrdahl: I'm not sure exactly how this value can be empty for some rounds.

I have updated the RoundBadge to show "Unknown" in these cases. Let me know if you prefer to completely remove it.
https://github.com/gitcoinco/grants-stack/blob/0d773c6dcf97cc2343e77a3cd4481ad2cd7350e0/packages/grant-explorer/src/features/discovery/RoundCard.tsx#L131-L135

PR #2505, opened 2023-10-25
initial dev env

gravityblast:
This is a PR to start the local dev env development. 
With this PR it's possible to use Builder to create/edit/show projects fully in local, with local chains and a local pinata service for metadata files.

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


packages/builder/src/components/base/ImageInput.tsx
```diff
@@ -100,7 +100,7 @@ export default function ImageInput({
 
     // Fetch existing img path from Pinata for display
     const pinataClient = new PinataClient();
-    const imgUrl = pinataClient.fileURL(ipfsCID);
+    const imgUrl = pinataClient.fileUrl(ipfsCID);
```
boudra, line 103: how did this work before? or it wasn't?

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+      - '127.0.0.1:3003:8000'
+
+  localchain-1:
+    image: ghcr.io/gitcoinco/allo-contracts-v1:pr-1
+    ports:
+      - '127.0.0.1:3005:8545'
+    environment:
+      - DEV_CHAIN_ID=313371
+
+  localchain-1-init:
```
boudra, line 27: what is the difference between the 3 `localchain` services here, my guess is one is the service and the other one is initializing something, but i don't get why they're separate if you don't mind explaining

gravityblast: @boudra sure, for each chain we have 2 things, `localchain` is the actual chain running, and `localchain-init` is the one that populates it. 
At the beginning I though that we could populate the chain (deploy contracts and create projects/rounds) at image build time. `anvil` allows to save the state of the chain, so my idea was then to run the container loading the same state for the chain. 
But `anvil` doesn't allow to dump events, it dumps only the chain state. And both for the indexer and Builder we need all the events. 
One idea could also be to run the container and then `docker exec` to populate, but having everything in the dockerfile felt better to have it automated. Is there a better way to do it?
Another thing to keep in mind is that when we populate we need the pinata (`pina`) service running to save all the metadata, so the container needs to be linked.

bard: Maybe just a matter of renaming it `localchain-1-initIALIZER` to remove doubts? @boudra would that have made it clearer? Or a few comments in the source?

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    image: ghcr.io/gitcoinco/allo-contracts-v1:pr-1
```
bard, line 28: I guess the tag `pr-1` will change once this is merged to master. Should this be changed here already?

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    depends_on:
+      - pinata
+      - localchain-1
+    environment:
+      - PINATA_HOST=pinata
```
bard, line 33: Also if this is initializing not just a chain but also IPFS, maybe there's a better name that `localCHAIN-init`.

Could it make sense to have just one init container that depends on `pinata`, `localchain-1`, and `localchain-2`, and call it `init` or something like that?

docs/DEV.md
```diff
@@ -1,33 +1,115 @@
...
+To kill the containers run:
 
-### Run in Development
+```sh
+docker-compose kill
+```
 
-Runs the app in the development mode.\
-Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
+### Setup the dapps dependencies
```
bard, line 96: Would probably say "root project dependencies" to distinguish that from the runtime dependencies mentioned above (Pinata, Indexer, etc)

docs/DEV.md
```diff
@@ -1,33 +1,115 @@
...
+And reset the wallet every time you restart the local chains:
+
+`Menu -> Settings -> Advance -> CLear activity tab data`
+⚠️⚠️⚠️
+
+
+
+
+
+
```
bard, line 74: Nice one. This will help a lot with any form of onboarding.

packages/builder/.env.example
```diff
@@ -1,15 +1,18 @@
...
 REACT_APP_INFURA_ID=
 REACT_APP_WALLETCONNECT_ID=
 REACT_APP_WALLETCONNECT_PROJECT_ID=
 
-# Get your credentials at https://www.pinata.cloud/
-REACT_APP_PINATA_GATEWAY=
-REACT_APP_PINATA_JWT=
+REACT_APP_IPFS_BASE_URL=http://localhost:3003
+REACT_APP_PINATA_BASE_URL=http://localhost:3003
+REACT_APP_PINATA_JWT=development-token
```
bard, line 15: We might have to talk about the "default" setup suggestion for first-time contributors. "Bring your own API keys" above would point in the direction of using external services, `localhost:3033` below would suggest we want them to use local services.

packages/builder/src/services/pinata.ts
```diff
@@ -1,42 +1,48 @@
...
+    this.gateway = z
+      .string({ required_error: "REACT_APP_IPFS_BASE_URL is required" })
+      .url()
+      .parse(process.env.REACT_APP_IPFS_BASE_URL)
+      .replace(/\/$/, "");
+    this.pinataBaseUrl = z
+      .string({ required_error: "REACT_APP_PINATA_BASE_URL" })
+      .url()
+      .parse(process.env.REACT_APP_PINATA_BASE_URL)
+      .replace(/\/$/, "");
```
bard, line 29: Not in scope for this PR, but it would be nice in the future to decouple environment access from object instantiation, so that the object can be created in more contexts (notably, tests) without mocking an environment around it. So something like:

```
const config = readConfigFromEnv();
// ...
const pinataClient = new PinataClient({
  jwt: config.pinataJwt,
  baseUrl: config.pinataIpfsBaseUrl
})
```

packages/builder/src/utils/wagmi.ts
```diff
@@ -6,135 +6,72 @@ import {
...
-  },
-  blockExplorers: {
-    default: {
-      name: "pgnscan",
-      url: "https://explorer.sepolia.publicgoods.network",
-    },
-  },
-  testnet: true,
-};
+const enabledChains: Chain[] = [];
```
bard, line 52: Could all this be moved into  function so as to minimize the amount of calls done as side effect of module loading, and make it test-friendly?

bard (approved):

docker-compose.yml
gravityblast: @bard @boudra good idea, I renamed it and added a comment

docker-compose.yml
gravityblast: `main` doesn't exists anymore, we can merge this and update it later or review and merge the other one before this one: https://github.com/gitcoinco/allo-contracts-v1/pull/1

docker-compose.yml
boudra: yeah a comment is perfect, I like to approach, it was more to confirm my assumption

boudra:
@gravityblast what's the reason for having two local chains running? are they used for different things

docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+# Dev chain 2: 3007
+# Subgraph dev chain 2: 3008
+
+services:
+  pinata:
+    image: ghcr.io/gitcoinco/pina:main
+    ports:
+      - '127.0.0.1:3003:8000'
+
+  localchain-1:
```
boudra, line 20: @gravityblast what's the reason for having two local chains running? are they used for different things

packages/builder/.env.example
gravityblast: yes it's true, I think the end solution would be to have everything setup for localdev and explain how people can use external services just in case. wdyt?

docker-compose.yml
gravityblast: mostly to test the multiround checkout contract with rounds in different chains. but even to test all the checks we do on the current chain before writing to contracts

packages/builder/.env.example
bard: Sounds good to me.

packages/builder/src/services/pinata.ts
gravityblast: I created a getEnv function in common, for now the class is still initializing them in the constructor but now we can mock getEnv in tests, wdyt?

bard: Again I don't think it's necessary to address this in the PR but just to explain where I'm coming from, my reasoning is that if you make a class read configuration directly from the environment, the way to configure it is decided by the class author; if you make it read configuration from constructor arguments, they way to configure it is decided by the class user. The latter is more future-proof. Workarounds are possible when you can hook into the module loader (like with module mocks during testing) but even there it to me it looks like two wrong trying to make a right, when  the right was low-hanging to begin with.

packages/common/src/env.ts
```diff
@@ -0,0 +1,35 @@
...
+  required: boolean;
+  defaultValue?: string;
+}
+
+const defaultOptions = {
+  required: true,
+  defaultValue: undefined,
+};
+
+export function getEnv(
```
bard, line 13: Not very convinced about this, seems a step back from just using zod since it duplicates `optional` and `default` functionality but prevents using the rest (e.g. coercion). An indirection more than an abstraction.

If we're going toward the ideal situation of parsing env just once, just in one place, then passing the validated object around, I'd rather avoid building intermediate utilities.

By the way, we've talked about that end goal as something that needs quite a bit of work, but if we're content with having it at the module level for now rather than being passed down via React context, it could be as easy as:

```ts
// config.ts

let cached: Config | null = null

const getConfig = (env = process.env): Config => {
  if (cached === null) {
    cached = {
      val: z.string().default('...').parse(env.SOME_VALUE),
      val2: z.string().optional().parse(env.SOME_OTHER_VALUE)
    }
  }
  return cached
}

// foo.tsx

import { getConfig } from 'common/config'

const Comp: React.FC = () => {
  const { val, val2 } = getConfig();
  // ...
}
```

That way, `config.ts` would be the only place that deals with `zod` (at least for this purpose) and the only place where reading from the environment can fail; everything else can assume the object to be valid.

bard (approved):

packages/common/src/env.ts
gravityblast: good idea! I didn't want to arrive at that final solution but it doesn't make sense to have this intermediate solution, I'll do it directly with the full config and we can start using it only for this vars, and then continue with the rest

vacekj (approved):
Seconding all of Max's comments, otherwise LGTM.

packages/builder/.env.example
```diff
@@ -1,6 +1,9 @@
 # development for local, production for prod builds
 REACT_APP_ENV=development
 
+# removing this will enable the default production or staging chains
+REACT_APP_CHAINS=dev1,dev2
```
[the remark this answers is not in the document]
gravityblast: > This seems easy to misunderstand down the road, especially when seeing it in code without the benefit of the comment. It sounds like the list of enabled chains, so an empty variable would mean no chains, like in the indexer.

@bard @vacekj  my idea is to set the chains there explicitly but using this empty string trick in the transition, so that we can merge the PR (which is deployed automatically to prod), update all the env vars in prod and test that everything works correctly, and then remove the allowed empty string so that we always need to specify them.
but this is just the first step of the local dev env, and by the end of the work I want to have the same configuration for all the 3 apps

packages/builder/src/utils/wagmi.ts
```diff
@@ -6,166 +6,156 @@ import {
...
-    public: { http: ["https://rpc.publicgoods.network"] },
-  },
-  blockExplorers: {
-    default: {
-      name: "pgnscan",
-      url: "https://explorer.publicgoods.network",
-    },
-  },
-};
+function enabledChainsAndProviders() {
```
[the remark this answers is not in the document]
gravityblast: done, I'm pushing now

packages/builder/.env.example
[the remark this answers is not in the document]
gravityblast: updated it as you proposed @bard but still with ha temporary trick to allow optional vars until we globally use getConfig and we remove the verify-env task in favour of something similar calling directly getEnv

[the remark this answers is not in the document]
vacekj: a compile-time check will still be needed to ensure env vars are set during build and development. Paired with getConfig that returns non-optional strings (in a future PR), this ensures that the environment is always correct (or at least the env vars are set - doesn't verify the correctness of api keys ofc etc.), and greatly reduces having to deal with optional values in the codebase.

packages/builder/src/utils/wagmi.ts
```diff
@@ -118,7 +114,7 @@ function enabledChainsAndProviders() {
 
       chains.push(chain);
     });
-  } else if (appEnv === "production") {
+  } else if (config.env === "production") {
```
bard, line 117: Does this refer to `NODE_ENV` or the deployment environment? It might be useful to disambiguate. The indexer uses `config.deploymentEnvironment` IIRC to avoid confusion.

packages/common/src/config.ts
```diff
@@ -1,17 +1,50 @@
-import { getEnv } from "./env";
+import { z } from "zod";
 
-type PinataConfig = {
-  jwt: string;
-  gateway: string;
-  pinataBaseUrl: string;
+type Config = {
+  env: string;
```
bard, line 4: A union here would allow to catch typos at compile time.

packages/common/src/config.ts
```diff
@@ -1,17 +1,50 @@
...
 };
 
-export function loadPinataConfig(): PinataConfig {
-  // we use ! since they these vars are required by default
-  // and an exception will be thrown inside getEnv if they are empty
-  return {
-    jwt: getEnv("REACT_APP_PINATA_JWT")!,
-    gateway: getEnv("REACT_APP_IPFS_BASE_URL")!,
-    pinataBaseUrl: getEnv("REACT_APP_PINATA_BASE_URL")!,
+var config: Config | null = null;
```
bard, line 20: `let`?

packages/builder/src/utils/wagmi.ts
```diff
@@ -133,24 +129,25 @@ function enabledChainsAndProviders() {
   if (!usingDevOnlyChains) {
     if (
       process.env.NODE_ENV !== "test" &&
-      (infuraId === "" || alchemyId === "")
+      (config.blockchain.infuraId === undefined ||
+        config.blockchain.alchemyId === undefined)
     ) {
       throw new Error(
         "REACT_APP_INFURA_ID and REACT_APP_ALCHEMY_ID must be set to use non-local chains"
```
bard, line 136: Could this be done in `getConfig`? Would probably also help get rid of the non-null assertions a couple of lines below.

bard (approved):

packages/common/src/config.ts
gravityblast: I don't know how this var came out lol

packages/builder/src/utils/wagmi.ts
gravityblast: I updated it with config.appEnv to keep it similar to the previous one

packages/common/src/config.ts
gravityblast: good one, I updated it

boudra (approved):

PR #2506, opened 2023-10-25
[2382] Round loading skeletons for explorer update design

carlbarrdahl:
- #2382 

## PR checklist

- [ ] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


vacekj (approved):

melissa-neira (approved):
this is really cool!

vacekj:
@carlbarrdahl if the PR is not yet ready for review, please consider opening a Draft PR instead of a normal PR, such that we don't get pinged for reviews. this also obviates the need for WIP in the title. Thanks!

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -103,17 +104,22 @@ query GetRounds(
 }
 `;
 
-export function useRoundsTakingApplications() {
-  const currentTimestamp = Math.floor(Date.now() / 1000).toString();
-  const infiniteTimestamp = ethers.constants.MaxUint256.toString();
+const NOW_IN_SECONDS = Date.now() / 1000;
+const ONE_YEAR = 3600 * 24 * 365;
```
vacekj: would appreciate a SECONDS suffix here to make it super clear what unit this is in.

packages/grant-explorer/src/features/discovery/__tests__/ApplyNowSection.test.tsx
```diff
@@ -28,7 +28,8 @@ vi.mock("common", async () => {
   };
 });
 
-describe("<ApplyNowSection />", () => {
+// This component is no longer used
+describe.skip("<ApplyNowSection />", () => {
```
vacekj, line 32: If it's not used then we can delete it - it's in git if we ever want to bring it back.

vacekj (approved):

packages/grant-explorer/src/features/discovery/__tests__/ApplyNowSection.test.tsx
carlbarrdahl: Good point, I've removed them.

gravityblast (approved):

gravityblast:
great job @carlbarrdahl !

PR #2513, opened 2023-10-26
feat(round-settings): fix visual bugs

### LINKED ISSUE #1915: Manager: UX changes for Round Edit flow
### 6/20/23 Update: Backlogged from Round Edit Flow milestone

**Overview**

Some remaining UX work to do on the Round Settings page (when in edit mode): 

**All**
- [x] When a field is disabled or in readonly mode, the text should be light gray not black 

**Round & Application Period tab**

- [x] When a field is disabled, the border has a gap in the bottom left corner 
- [x] When disabled, the text in the Application Start Date box should be light gray, like the Application End Date & Round Start date boxes 

Screen Shot 2023-06-02 at 11 16 56 AM

**Funding Settings tab**

- [x] The header for payout token is cutoff if the selection is ETH 
Screen Shot 2023-06-02 at 11 25 43 AM

- [x] The "USD" header text should be light gray not black (make is the same as the ETH/DAI and % text color of the above headers) 
Screen Shot 2023-06-02 at 11 26 52 AM

- [x] (TBD) Remove redundant "required" labels for YES/NO questions 

**Details**
Say more

vacekj:
## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #1915

### DIFF
--- a/packages/round-manager/src/features/round/ViewRoundSettings.tsx
+++ b/packages/round-manager/src/features/round/ViewRoundSettings.tsx
@@ -570,7 +570,6 @@ export default function ViewRoundSettings(props: { id?: string }) {
   );
 }
 
-//
 function DetailsPage(props: {
   editMode: EditMode;
   editedRound: Round;
@@ -1288,9 +1287,10 @@ function RoundApplicationPeriod(props: {
               </div>
               <div
                 className={`leading-8 font-normal ${
-                  !props.editMode.canEdit &&
-                  props.editMode.canEditOnlyRoundEndDate &&
-                  "text-grey-400"
+                  !props.editMode.canEdit ||
+                  props.editMode.canEditOnlyRoundEndDate
+                    ? "text-grey-400"
+                    : ""
                 }`}
               >
                 {props.editMode.canEdit &&
@@ -1301,11 +1301,11 @@ function RoundApplicationPeriod(props: {
                   <div className="col-span-6 sm:col-span-3">
                     <div
                       className={`${
-                        !props.editMode.canEdit ? "bg-grey-50" : ""
+                        !props.editMode.canEdit ? "bg-grey-50 " : ""
                       } relative border rounded-md px-3 pb-2 mb-2 shadow-sm focus-within:ring-1 ${
                         props.errors.applicationsStartTime
                           ? "border-red-300 text-red-900 placeholder-red-300 focus-within:outline-none focus-within:border-red-500 focus-within: ring-red-500"
-                          : "border-gray-300 focus-within:border-indigo-600 focus-within:ring-indigo-600"
+                          : " border-gray-300 focus-within:border-indigo-600 focus-within:ring-indigo-600"
                       }`}
                     >
                       <p className="block text-[10px]">Start Date</p>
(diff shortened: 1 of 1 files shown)


packages/round-manager/src/features/round/ViewRoundSettings.tsx
```diff
@@ -1825,14 +1825,15 @@ function Funding(props: {
...
 
           <div
-            className={`leading-8 font-normal ${
+            className={`leading-8 flex font-normal ${
               !props.editMode.canEdit ||
               (props.editMode.canEditOnlyRoundEndDate && "text-grey-400")
             }`}
           >
             <input
+              size={matchingFundPayoutToken.name.length ?? 3}
```
bard, line 1834: If I'm understanding this correctly, and name is a string, then length will never be null or undefined, and the `?? 3` branch will never trigger.

bard (approved):

bhargavaparoksham (approved):

packages/round-manager/src/features/round/ViewRoundSettings.tsx
carlbarrdahl: I think if name is empty (`""`) it will default to size 3. What happens to the input field when changing between empty, to one character, to two? Will it jump from size 3 to 1 and then 2?

packages/round-manager/src/features/round/ViewRoundSettings.tsx
```diff
@@ -2238,14 +2234,17 @@ function Funding(props: {
             </span>
           </div>
           <div
-            className={`leading-8 font-normal ${
+            className={`leading-8 flex font-normal ${
               !props.editMode.canEdit ||
               (props.editMode.canEditOnlyRoundEndDate && "text-grey-400")
             }`}
```
carlbarrdahl, line 2240: This library is great for these instances:
https://github.com/lukeed/clsx

Would look like this:
```tsx
className={clsx(`leading-8 flex font-normal`, {
  ["text-grey-400"]: !props.editMode.canEdit || props.editMode.canEditOnlyRoundEndDate
})
```

packages/round-manager/src/features/round/ViewRoundSettings.tsx
bard:
```
$ node
Welcome to Node.js v20.6.1.
Type ".help" for more information.
> "".length ?? 3
0
```

If name is empty, length is zero, which is neither null nor undefined, so the `??  3` branch won't trigger.

packages/round-manager/src/features/round/ViewRoundSettings.tsx
vacekj: Indeed, however I would leave this sort of refactoring for the actual refactoring ticket for this page, which is imminent after all the current tickets for this page are implemented.

packages/round-manager/src/features/round/ViewRoundSettings.tsx
vacekj: the string will never be empty if the token is found, but the token could theoretically not be found for whatever reason, even though typescript doesn't capture this (because it assumes that .filter() on an array always produces a non-empty array), and therefore will be undefined. I forgot some opt. chaining operators, added in the latest commit.

melissa-neira (approved):
lgtm!

PR #2516, opened 2023-10-26
bug: fixed eslint error while starting builder

### LINKED ISSUE #2515: builder : eslint throws an error when trying to setup the project locally
**Overview**
I was trying to setup the project locally by following the [DEV.md](https://github.com/gitcoinco/grants-stack/blob/main/docs/DEV.md) but got some errors

**Steps to reproduce**
- After `pnpm start` step , the development server starts but throws a few errorrs

**Screenshots**
ss

**Additional context**
It is a linting issue and can be solved by making changes in the [eslint file](https://github.com/gitcoinco/grants-stack/blob/main/packages/builder/.eslintrc.js)

ArslanYM:
## PR checklist

- [X] Doesn't contain type casts and non-null assertions.
- [X] Doesn't add `@ts-ignore`.
- [X] Doesn't disable lints.
- [X] Doesn't use `useState` just for computation - use plain variables instead.
- [X] Splits components into pure components that don't depend on external state or hooks.
- [X] Avoid embedding components within other components
- [X] Doesn't propagate optional values without good reason, doesn't mark property values as optional if that doesn't represent reality.
- [X] Doesn't duplicate existing code.
- [X] Parses out-of-domain data - this includes user input, API respones, on-chain data etc.
- [X] Doesn't contain skipped or empty tests.
- [X] If this PR adds/updates any feature, it adds/updates its test script

##### Description
Added a eslint rule,  The "endOfLine" option is set to automatically determine the line ending style based on the file it's formatting.

##### Refers/Fixes
fixes #2515


vacekj (approved):

vacekj (approved):

melissa-neira:
I defer my review to the eng team 🚀

ArslanYM:
> @ArslanYM this looks ok to me and we can merge it as-is, but I'd be curious to better understand why this happened in the first place. Do files in our repository differ in line terminations? Are only specific files triggering eslint? Would be great to find out why.

This is what I found out : 

**** 
The symbol ␍ represents a carriage return (CR), which is a control character used in old Mac systems for line breaks. In contrast, Windows uses both a carriage return (CR) and a line feed (LF) for line breaks, while Unix-based systems (including Linux) use only the line feed (LF).

When working on a codebase,  inconsistent line endings can cause issues with version control systems and code formatting tools (such as ESLint ) since they expect uniformity in line endings.

bard:
@ArslanYM Can you check whether there are files that only have CR (as opposed to LF or CR+LF) as line termination? Would be good to fix those.

ArslanYM:
> @ArslanYM Can you check whether there are files that only have CR (as opposed to LF or CR+LF) as line termination? Would be good to fix those.

This actually works. I ran the `git grep -I --name-only -l $'\r'` command to check which files had CR line endings. Found out that there were a lot of them. Then ran `find . -type f -exec dos2unix {} \` to find and change all those to LF . This fixed the linting errors. But I can't seem to stage these changes.

bard:
> This actually works. I ran the `git grep -I --name-only -l $'\r'` command to check which files had CR line endings.

I _think_ that command only tells you which files have CR anywhere in a line. To find CR line endings, you might have to change it to something like `'\r$'`.

boudra:
we should fix files that don't end with LF instead of changing the prettier config to allow it

Issue #1982, opened 2023-06-26
bug: unable to edit round even though round has not started


michellema1208:
**Overview**
**From Michelle**: It looks like at the contract level, we might be requiring the application start time to also be in the future, so Connor and team are not able to edit a round even though the round start date is in the future, because the application start date is already past. I would expect to be able to make any edits to my round so long as the round has not started. 

**From Connor:** 

This instance we'd want to move the Zuzalu round (https://manager.gitcoin.co/#/round/0x74911b66c1d50cdaf24ba58cada60316a41f5c69 )
but we've tried this in the past and learned it won't work 

Zuzalu application period has started, not ended, and round hasn't started. We'd want to push the round start date back. But it's not possible. Just tried it via the UI, you just get this error and then it refreshes

<img width="732" alt="Screenshot_2023-06-26_at_8 27 40_AM" src="https://github.com/gitcoinco/grants-stack/assets/26017180/8134258a-07f0-4627-ab00-9a8f044412dc">

This is what it looks like trying to edit via etherscan, which we tried in the past. You have to put in values for all 4 fields. So the 
Only updates if new time is in the future and current set time is also in the future
 clause means you can't update any of them if the first thing (applications opening) has begun

Also related to: https://github.com/gitcoinco/grants-stack/issues/2094

<img width="533" alt="Screenshot_2023-06-26_at_8 28 36_AM" src="https://github.com/gitcoinco/grants-stack/assets/26017180/4f1df1d6-352a-4cae-a603-ba8149fb8d1d">

**Steps To Reproduce**

**Observed  behavior**
A clear and concise description of what the observed behavior of the system.

**Expected behavior**
A clear and concise description of what you expected the system to do.

**Screenshots**

**Additional context**
Add any other context about the problem here that will assist in triaging the bug. I.e. Browser information, unique identifiers, etc.

vacekj:
Same problem as https://github.com/gitcoinco/grants-stack/issues/2113#issuecomment-1780883799

PR #2517, opened 2023-10-26
feat(round-settings): fix round time editing bug

### LINKED ISSUE #2113: Manager: unable to edit round end date
**Overview**
I am unable to edit the round end date because of a gas estimation bug. This loom was taken on staging but I verified that the same behavior is happening in production.

**Steps To Reproduce**
1. Go to https://rmgitcoin.on.fleek.co/#/round/0xbd84a5c0affbcd4bf285ab20ebecae20f0a31f36?debug=true
2. Click on Round settings -> Dates -> Edit
3. Try to change the round date to LATER than the previous date

**Observed  behavior**
Loom: https://www.loom.com/share/d7fcdc91338347f38598377d5e1ba650

**Screenshots**
Console error: 
Screenshot 2023-08-03 at 8 12 26 PM

vacekj:
## PR checklist

  - [x] Doesn't disable lints.

##### Refers/Fixes

fixes #2113 #1982

### DIFF
--- a/packages/round-manager/src/context/round/UpdateRoundContext.tsx
+++ b/packages/round-manager/src/context/round/UpdateRoundContext.tsx
@@ -7,6 +7,7 @@ import { waitForSubgraphSyncTo } from "../../features/api/subgraph";
 import { EditedGroups, ProgressStatus, Round } from "../../features/api/types";
 import { getPayoutTokenOptions } from "../../features/api/utils";
 import { useWallet } from "../../features/common/Auth";
+import subSeconds from "date-fns/subSeconds";
 
 type SetStatusFn = React.Dispatch<SetStateAction<ProgressStatus>>;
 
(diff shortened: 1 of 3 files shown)


packages/round-manager/src/context/round/UpdateRoundContext.tsx
```diff
@@ -157,13 +158,22 @@ const _updateRound = async ({
...
     }
 
+    /* Special case - if the round has already started, and we are editing times,
+     * we need to set newApplicationsStartTime and newRoundStartTime to something bigger than the block timestamp.
+     * This won't actually update the values, it's done just to pass the checks in the contract
+     * (and to confuse the developer). */
+    if (Date.now() > round.roundStartTime.getTime()) {
+      round.applicationsStartTime = subSeconds(round.applicationsEndTime, 1);
+      round.roundStartTime = subSeconds(round.applicationsEndTime, 1);
+    }
```
gravityblast, line 168: let's explain why in the comment, it makes sense to remember that the contract call doesn't fail, 
it updates the end time only if the round hasn't started yet

gravityblast (approved):

packages/round-manager/src/context/round/UpdateRoundContext.tsx
vacekj: Could you be more specific here? Not sure what you want to change about the explanation.
Feel free to suggest the changes to be commited directly in the review here.

vacekj: TODO: link the contract checks here

packages/round-manager/src/context/round/UpdateRoundContext.tsx
```diff
@@ -157,13 +158,22 @@ const _updateRound = async ({
       transactionBuilder.add(UpdateAction.UPDATE_ROUND_FEE_PERCENTAGE, [arg]);
     }
 
+    /* Special case - if the round has already started, and we are editing times,
+     * we need to set newApplicationsStartTime and newRoundStartTime to something bigger than the block timestamp.
+     * This won't actually update the values, it's done just to pass the checks in the contract
+     * (and to confuse the developer). */
```
vacekj, line 164:
```suggestion
     * (and to confuse the developer). 
    *  https://github.com/allo-protocol/allo-contracts/blob/9c50f53cbdc2844fbf3cfa760df438f6fe3f0368/contracts/round/RoundImplementation.sol#L339C1-L339C1*/
```

bhargavaparoksham (approved):

PR #2521, opened 2023-10-30
Filter out rounds with no projects from Explorer landing page

### LINKED ISSUE #2405: [Round Spam] Explorer: filter out rounds with no projects AND application period has closed
**User Story**
As a user of Grants Stack
I want to make sure I am engaging with legitimate rounds 
So that I don't waste my time (i.e. applying to a fake round) or money (i.e. donating to a fake round)

**Acceptance Criteria**
GIVEN that I am on the Explorer Homepage 
WHEN I look at the page 
THEN I don't see any rounds with 0 projects where the application period has closed 

**Links**
* [Figma](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=300-14464&mode=design&t=os8gMWwBouzyFg6y-0)
* [Sample staging environment to see existing homepage](https://explorer-staging-git-bug-2394-grants-stack.vercel.app/#/)

bhargavaparoksham:
## PR checklist

- [ ] Doesn't disable lints.

##### Description

Filter out rounds with no projects from Explorer landing page

##### Refers/Fixes

fixes #2405


packages/grant-explorer/src/features/discovery/LandingPage.tsx
```diff
@@ -36,14 +36,21 @@ const LandingPage = () => {
...
+            )
+            .slice(0, 4)
+            .map((round, i) => (
+              <div
+                key={round.id}
+                className={`${i % 3 === 0 ? "md:col-span-2" : ""}`}
+              >
+                <RoundCard round={round} />
+              </div>
+            ))}
```
carlbarrdahl, line 53: Is it possible to do this on a query level in the graph? If not, move this logic to the hook useActiveRounds:

```ts
export function useActiveRounds() {
  const [currentTimestamp] = useState(createTimestamp());
  const [futureTimestamp] = useState(createTimestamp(ONE_YEAR * 10));
  const query = useRounds({
    orderBy: "roundEndTime",
    orderDirection: "desc",
    where: {
      // Round must have started and not ended yet
      roundStartTime_lt: currentTimestamp,
      roundEndTime_gt: currentTimestamp,
      roundEndTime_lt: futureTimestamp,
    },
  });
  return {
    ...query,
    data: query.data?.filter(round => round.projects?.length)
  }
}
```

packages/grant-explorer/src/features/discovery/LandingPage.tsx
```diff
@@ -75,11 +82,17 @@ const LandingPage = () => {
...
-            <div key={round.id}>
-              <RoundCard round={round} />
-            </div>
-          ))}
+          {activeRounds.data
+            ?.filter(
+              (round) =>
+                round?.projects?.length !== undefined &&
+                round.projects.length > 0
+            )
```
carlbarrdahl, line 90: This can be removed by moving the filter to the hook.
Also, it's sufficient with `round.projects?.length > 0`

carlbarrdahl (changes requested):

melissa-neira (commented):
What should we do about rounds that don't have any projects yet but are in the application period? cc: @michellema1208 
Screenshot 2023-10-31 at 11 50 33

vacekj (approved):
Seconding Carl's comments, otherwise LGTM.

michellema1208 (commented):
hey @bhargavaparoksham ! I'm noticing that the filtering doesn't apply to rounds displayed on the Explore Rounds tab (https://explorer-git-explorer-2405-grants-stack.vercel.app/#/rounds) 

I apologize if the AC wasn't clear in the original ticket - was thinking that the Explore Rounds tab is part of the Explorer Homepage / pulls from the same database of rounds, so I didn't explicitly call that out. If it would be easier to do this in a follow-up ticket, let me know!

bhargavaparoksham:
@michellema1208 Will implement the filtering for rounds on Explore Rounds tab in this PR only. No need for another ticket.

Re the Apply Now section, showing rounds only when it has been partially funded makes sense. If you can add back that ticket into our ToDo, I can work on it as part of a separate PR. 

cc: @melissa-neira

vacekj (approved):
Seconding Carl's comments, otherwise LGTM, great work.

melissa-neira:
@michellema1208 I'd be down to get rid of the "apply now" section if that makes things easier! wdty? cc: @bhargavaparoksham

michellema1208:
@melissa-neira @bhargavaparoksham let's remove the "Apply for Funding" section from the homepage for now! We'll still have those rounds on the Explore Rounds tab, but less "highlighted" on the Home page. 

We might be able to track how often folks go to Explore Rounds > select filter "status=Taking Applications" -- maybe if we see that that's a common user behavior flow, we could bring back this section on the Home page.

packages/grant-explorer/src/features/discovery/LandingPage.tsx
```diff
@@ -49,35 +47,6 @@ const LandingPage = () => {
...
-              .slice(0, 4)
-              ?.map((round) => (
-                <div key={round?.id}>
-                  <RoundCard
-                    round={round}
-                    isLoading={roundsTakingApplications.isLoading}
-                  />
-                </div>
-              ))}
-          </div>
```
carlbarrdahl, line 78: Was this removal unintentional?

bhargavaparoksham: https://github.com/gitcoinco/grants-stack/pull/2521#issuecomment-1790657264

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -193,6 +194,12 @@ export function useRounds(variables: RoundsVariables) {
   };
 }
 
+function filterRoundsWithProjects(rounds: RoundOverview[]) {
+  return rounds.filter((round) => {
+    return round?.projects?.length !== undefined && round.projects?.length > 0;
```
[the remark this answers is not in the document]
bhargavaparoksham: Just round.projects?.length is throwing prjects.length ccould be undefined error.

michellema1208 (approved):

packages/grant-explorer/src/features/api/rounds.ts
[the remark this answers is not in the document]
carlbarrdahl: Even with `round?.projects?.length` (with a ? after round)?

packages/grant-explorer/src/features/discovery/LandingPage.tsx
carlbarrdahl: Ah I see

packages/grant-explorer/src/features/api/rounds.ts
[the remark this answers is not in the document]
bard: `return round?.projects?.length !== undefined && round.projects?.length > 0` is actually simpler, since it returns `true` or `false` at a place where a boolean is expected. `round?.projects?.length?` is shorter but more complex, since it returns an integer or undefined at a place where a boolean is expected, and that is then converted to a boolean by JavaScript's implicit coercion. After being burned by implicit coercion too many times, I now always prefer explicit to implicit.

PR #2541, opened 2023-11-02
Update getConfig to fix failing tests

### LINKED ISSUE #2539: fix tests setting env vars in Manager and Explorer

bhargavaparoksham:
## PR checklist

- [x] Doesn't disable lints.

##### Description

Update getConfig to fix failing tests in manager & explorer

##### Refers/Fixes

fixes #2539

### DIFF
--- a/packages/common/src/config.ts
+++ b/packages/common/src/config.ts
@@ -30,11 +30,23 @@ export function getConfig(): Config {
       .default("development")
       .parse(process.env.REACT_APP_ENV),
     ipfs: {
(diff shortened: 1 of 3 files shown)


carlbarrdahl:
Fixed the tests that didn't work in Explorer. The rounds mock needed to be updated with projects to not be filtered out when displaying on the landing page.

boudra (approved):
LGTM as a temporary measure to fix the tests but we should look for a better solution as this kind of negates one of the benefits of parsing the env vars.

PR #2546, opened 2023-11-07
Explore projects page

boudra:
This is a first version of the Explore projects page, work still left to be done in following PRs:

- Migrate grants-stack-data-client to this repo
- Implement filters & search

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


melissa-neira:
@boudra there's a bit more spacing between the title and the project cards than titles and round cards on the other tabs. let's make sure it's consistent on the projects tab as well.

projects tab:
Screenshot 2023-11-08 at 17 33 07

rounds tab: 
Screenshot 2023-11-08 at 17 32 59

packages/common/src/routes.ts
```diff
@@ -0,0 +1,8 @@
+import { Address } from "wagmi";
+import { ChainId } from "./chains";
+
+export const Explorer = {
+  applicationPath(chainId: ChainId, roundId: Address, applicationId: string) {
+    return `/#/round/${chainId}/${roundId.toLowerCase()}/${roundId.toLowerCase()}-${applicationId}`;
```
carlbarrdahl, line 6: Is the double `roundId` here intentional?

packages/grant-explorer/src/features/discovery/ExploreProjectsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+    amount: "",
+  };
+}
+
+function createCompositeRoundApplicationId(application: ApplicationSummary) {
+  return `${application.roundId}-${application.roundApplicationId}`;
+}
+
+const ExploreProjectsPage = () => {
+  const [seed] = useState(() => Math.random());
```
carlbarrdahl, line 38: This will get a new seed every time the page is loaded. Could the user be confused by seeing a different order every time? By putting the seed definition two lines up (without the useState), it will use the same seed until page refresh.

carlbarrdahl (approved):
Looks good. Great work on the tests!

packages/common/src/routes.ts
boudra: intentional! the url structure is like that 🙃

packages/grant-explorer/src/features/discovery/ExploreProjectsPage.tsx
boudra: I had initially stored the seed in `sessionStorage` to make it last the session but this works as well to make it less confusing.

gravityblast (approved):

bhargavaparoksham (approved):

melissa-neira (changes requested):
round names aren't being truncated correctly:
Screenshot 2023-11-09 at 12 49 22
should be like [this](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=1085-25442&mode=design&t=sGok79EONDusM601-4)

also seeing that the tabs aren't being correctly shown in mobile:
Screenshot 2023-11-09 at 12 51 07
should be like [this](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=829-55850&mode=design&t=sGok79EONDusM601-4). let me know if it makes more sense to make this into a separate ticket!

boudra:
> round names aren't being truncated correctly: Screenshot 2023-11-09 at 12 49 22 should be like [this](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=1085-25442&mode=design&t=sGok79EONDusM601-4)
> 
> also seeing that the tabs aren't being correctly shown in mobile: Screenshot 2023-11-09 at 12 51 07 should be like [this](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=829-55850&mode=design&t=sGok79EONDusM601-4). let me know if it makes more sense to make this into a separate ticket!

@melissa-neira fixed the truncation, and yes, do you mind if we do the tabs in a separate one since it also affects the round page as well, I didn't realize it changed text  on mobile

packages/grant-explorer/src/features/common/ProjectCard.tsx
```diff
@@ -97,7 +97,12 @@ export function ProjectCard(props: {
             </div>
           </CardDescription>
 
-          <Badge color="grey" rounded="3xl">
+          <Badge
+            color="grey"
+            rounded="3xl"
+            className="truncate max-w-full"
```
carlbarrdahl: Dunno what the implications are but what if we moved these classes to the component?

packages/grant-explorer/src/features/common/styles.ts
```diff
@@ -80,17 +80,18 @@ export const Badge = tw.div<{
...
   text-xs
   text-gray-900
   bg-gray-100
   whitespace-nowrap
-  inline-flex
   items-center
   justify-center
   px-2
   py-1.5
+  ${(p) => (p.flex ?? true ? "inline-flex" : "")}
```
carlbarrdahl, line 94: Seems unnecessary with a prop for this. I would rather just use the className inline-flex.

boudra: the inline flex is used by default which breaks truncation, this prop adds a way to remove it, I don't want to risk making changes here and potentially break callers everywhere, either way I have deleted it and used a simpler approach

michellema1208 (approved):
LGTM - amazing work as usual, Mo!

packages/common/src/routes.ts
```diff
@@ -0,0 +1,8 @@
+import { Address } from "wagmi";
+import { ChainId } from "./chains";
+
+export const Explorer = {
```
bard, line 4: Seeing `Explorer` at calling site, I'd automatically think "class" or "type" because that's what I see capitalized names most used for.

packages/common/src/routes.ts
```diff
@@ -0,0 +1,8 @@
...
+  applicationPath(chainId: ChainId, roundId: Address, applicationId: string) {
```
bard, line 5: This is missing the return type, but that aside, is it worth making the parameters an object? The calling site likely already has variables with those exact names, so the call would only be two characters longer (braces), and would guard against this kind of compiler-undetectable bugs:

```tsx
  <a href={routes.applicationPath(roundId, chainId, applicationId)}> ... </a>
```

bard (approved):

packages/builder/.env.example
```diff
@@ -28,6 +28,8 @@ REACT_APP_DEBUG_MODE=false
 
 # URL of the indexer. http://localhost:8080 for local indexer
 REACT_APP_ALLO_API_URL="https://indexer-production.fly.dev"
+# URL where the Grants Stack Data Client should point to
+REACT_APP_GRANTS_STACK_DATA_CLIENT_BASE_URL="https://gitcoin-search-dev.fly.dev"
```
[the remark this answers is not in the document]
boudra: waiting on @gravityblast to fix CORS on that distribution, will update once that is done

melissa-neira (approved):

packages/common/src/config.ts
```diff
@@ -36,6 +39,14 @@ export function getConfig(): Config {
         .default("https://local-ipfs.dev")
         .parse(process.env.REACT_APP_IPFS_BASE_URL),
     },
+    grantsStackDataClient: {
+      baseUrl: z
+        .string()
+        .url()
+        // TODO: fix `env.test` in tests to remove this
```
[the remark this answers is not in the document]
boudra: the env stub works for that specific test but this line crashes all manager and builder tests

packages/common/src/routes.ts
boudra: hmm true, I might just split the app routes in their own files, so:

```ts
import { applicationsPath } from "./routes/explorer";
```

packages/grant-explorer/src/features/common/ProjectCard.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+    roundId: zeroAddress,
+    chainId: 1,
+    roundApplicationId: "1",
+    name: "Project Name",
+    summaryText: "Project Summary",
+    bannerImageCid: "bannerCid",
+    logoImageCid: "logoCid",
+  };
+
+  vi.stubEnv("REACT_APP_GRANTS_STACK_DATA_CLIENT_BASE_URL", "https:/ipfs.io");
```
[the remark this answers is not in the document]
boudra: removed, not worth the confusion in general

packages/grant-explorer/src/features/discovery/PaginatedProjectsList.tsx
```diff
@@ -0,0 +1,69 @@
...
+            className="rounded-3xl border border-white bg-[#F3F3F5] text-md font-medium px-5 py-3 flex items-center"
+            disabled={isLoadingMore}
+            onClick={() => loadNextPage()}
+          >
+            {isLoadingMore ? (
+              <LoadingRing className="animate-spin w-5 h-5" />
+            ) : (
+              <>
+                <PlusIcon className="w-5 h-5 mr-1" />
+                <span>Load more</span>
```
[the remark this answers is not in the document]
boudra: Nope, it's the actual button text, looks like this [ + Load more ] at the end of the application list

packages/common/src/routes.ts
boudra: you're right, fixed!

packages/grant-explorer/src/features/common/ProjectCard.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  it("renders correctly with required props", () => {
+    const addToCart = vi.fn();
+    const removeFromCart = vi.fn();
+
+    render(
+      <ProjectCard
+        application={mockApplication}
+        inCart={false}
+        addToCart={addToCart}
```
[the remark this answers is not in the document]
boudra: fixed

Issue #2519, opened 2023-10-27
add grants stack logo background & animation to text in hero


melissa-neira:
**can whoever picks this up ping me? I'd like to pair and see if this is the right solution**

- [ ] add grants stack logo in the background - [figma](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=1073-4884&mode=design&t=fhUd6VMy0K6JHEl4-4) 
- [ ] make text animated in hero like [this](https://impact.gitcoin.co/)

PR #2548, opened 2023-11-07
[WIP] Add animated numbers to hero

carlbarrdahl:
- Fixes part 2 in #2519 

## PR checklist

- [ ] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


packages/grant-explorer/src/features/discovery/LandingHero.tsx
```diff
@@ -14,11 +15,15 @@ export default function LandingHero() {
   );
 }
 
+const numberFormatter = new Intl.NumberFormat();
 const ByTheNumbers = () => {
   // TODO: fetch data (where to get this?)
-  const uniqueDonations = "3.8M";
-  const raisedFunds = "3,715";
-  const fundingDistributed = "$50,000,000+";
+  const uniqueDonations = useAnimateNumber({ value: Number(3800) });
```
vacekj, line 21: Is there a particular reason to use the Number constructor here, and not just a number literal?

vacekj (approved):

packages/grant-explorer/src/features/discovery/LandingHero.tsx
carlbarrdahl: No, you're right. It should be just `3800`.

carlbarrdahl:
Decided to use static numbers instead.

See Discord for more info: 
https://discord.com/channels/562828676480237578/1111042625546485872/1171451930116096010

PR #2551, opened 2023-11-08
Explorer: 2550 filter bugs

### LINKED ISSUE #2550: Explorer: minor bugs on new Explorer Homepage
**Summary:** Tracking some minor bugs that I'm seeing on the new Explorer Homepage 

**Tasks**

### Round Filters 
- [x] In the filters on the Explore Rounds tab: seeing testnets and production networks - should only be supported production networks if on production URLs (if on staging, show both testnet and production networks)  
> Supported production networks: Mainnet, Optimism, PGN, Polygon PoS, Fantom, Arbitrum, Avalanche 

Screen Shot 2023-11-07 at 11 05 31 AM

- [x] When I click on the "Explore Rounds" tab directly, it says that there are no filters (filter = All) and it's showing all rounds, but is only actually showing all active rounds. When I go to this page, I would expect `status = Active` & `status = Taking Applications` to be already checked, so we can filter out rounds that have finished by default 
> Loom: https://www.loom.com/share/762d2c8d31cc4a59868ce02edc54c263

### Defining Active Rounds
Loom (covers below 3 tasks): https://www.loom.com/share/17c15fb479bf4e2a852ad6fb359c0676

- [x] Active Rounds should be rounds that have voting period open -- a.k.a the round has started AND the round has not ended (currently, it looks like we're defining Active Rounds only as rounds that have not ended, which means we're pulling in rounds that have not started as well) 

- [x] For Rounds that have not started yet, let's show copy that says "X days until round starts" based on round start date. Once the round starts, we would show "X days left in round" based on round end date like we currently have. 

- [x] In case this doesn't automatically update, let's make sure the Donate Now section only shows Active Rounds (according to updated definition above) 

### Sorting 
Loom (covers below 3 tasks): https://www.loom.com/share/270b716e733d4000bcf474c6a6b4ef82 

- [x] Sorting by matching pool high to low not sorting accurately (see Loom for more details)

- [x] When I click on Rounds Ending Soon > View All, I'm taken to Explore Rounds tab with `Round end (earliest)` sort applied, and `status = Active`

- [x] Remove "Ending Soon" as a filter 

### Direct Grants Rounds 
Loom (covers below 3 tasks): https://www.loom.com/share/03e9b0f104a540efa78e64f80c8da82e

- [x] DG rounds - says "Ends today" for rounds that have No End Date -- let's show copy "No round end date" instead 

- [x] Remove "match amount" label from Direct Grants round cards 

- [ ] Make sure that all active Direct Grants rounds are shown when the filters `status = Active` and `type = Direct Grants` are selected

carlbarrdahl:
- Fixes #2550 

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


michellema1208:
hey @carlbarrdahl - I'm seeing a lot of the original issues I flagged now fixed, amazing progress! 

Sharing a couple things I'm still seeing from testing: 

Loom (covers all 3 tasks below): https://www.loom.com/share/d05bc1e735da4024a64f1971c780ab31

- [ ] "Rounds Ending Soon" section includes a round that has not started yet

- [ ] Not all rounds taking applications are showing up when `status = Taking Applications` is applied (compared to [same filters applied on production](https://explorer.gitcoin.co/#/rounds?orderBy=matchAmount&orderDirection=desc&status=taking_applications) 
> `Taking Applications` filter should be defined as application period has started AND application period has not ended

- [ ] Direct Grants rounds with no round end date still say "Ends today" -- should say "No round end date"

melissa-neira (approved):
@carlbarrdahl +1 to @michellema1208's comments! otherwise, everything looks good UI-wise

packages/grant-explorer/src/features/discovery/RoundDaysDetails.tsx
```diff
@@ -0,0 +1,76 @@
...
+export function getRoundDaysText({
+  roundStartsIn,
+  roundEndsIn,
+}: {
+  roundStartsIn?: number;
+  roundEndsIn?: number;
+}) {
+  const days = pluralize(["day", "days"]);
+
+  if (roundStartsIn && roundStartsIn > 0) {
```
bard, line 12: `undefined | number` used as boolean here.

packages/grant-explorer/src/features/discovery/RoundDaysDetails.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return `Starts in ${roundStartsIn} ${days(roundStartsIn)}`;
+  }
+  if (roundEndsIn === 0) return "Ends today";
+  if (!roundEndsIn) return "No round end date";
```
bard, line 16: Idem

packages/grant-explorer/src/features/discovery/RoundDaysDetails.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  if (applicationsEndsIn === 0) return "Last day to apply";
+  if (
+    applicationsStartsIn &&
+    applicationsStartsIn > 0 &&
+    applicationsEndsIn > 0
+  ) {
+    return `Apply in ${applicationsStartsIn} ${days(applicationsStartsIn)}`;
+  }
+  return `${applicationsEndsIn} ${days(applicationsEndsIn)} left to apply`;
```
bard, line 43: Only mentioning this kicks in case you like pattern matching: in TypeScript 5.3 it will be finally possible to have:

```ts
switch(true) {
  case applicationsEndsIn === undefined: 
    return "";
  case applicationsEndsIn < 0:
    return "";
  case applicationsEndsIn === 0: 
    return "Last day to apply";
  // ...
  default:
    return `${applicationsEndsIn} ${days(applicationsEndsIn)} left to apply`;
}
```

packages/grant-explorer/src/features/discovery/RoundDaysDetails.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function getRoundApplicationDaysText({
+  applicationsStartsIn,
+  applicationsEndsIn,
+}: {
+  applicationsStartsIn?: number;
+  applicationsEndsIn?: number;
+}) {
+  const days = pluralize(["day", "days"]);
```
bard, line 30: `pluralize()` doesn't actually pluralize its arguments so it's misguiding. Not sure how to call it though — I'd think `createPluralizer`, but what it creates isn't a pluralizer (something that takes a work and produces its plural) either.

packages/grant-explorer/src/features/discovery/RoundMatchAmountBadge.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function RoundMatchAmountBadge({
+  chainId,
+  matchAmount,
+  token,
+}: RoundCardStatProps) {
+  const { data } = useToken({
+    address: getAddress(token),
+    chainId: Number(chainId),
+    enabled: Boolean(token && token !== zeroAddress),
```
bard, line 23: Here there is a boolean coercion (`token &&`) and then a boolean cast. Seems a case of two wrongs trying to make a right. :)

Can the `token` variable ever be an empty string? If not, then this would be enough:

```
enabled: token !== zeroAddres
```

If yes:

```
enabled: token !== '' && token !== zeroAddress
```

Both would be the least amount of mental parsing/execution for their respective case.

Also, the `token` name has some ambiguity when seen in isolation — is it `{ symbol: string, address: string }` or is it `string`? Only few lines below `nativePayoutToken` is more like the former case, for example.

Naming it `tokenAddress` would save the reader the work of dealing with the ambiguity.

packages/grant-explorer/src/features/discovery/RoundMatchAmountBadge.tsx
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  const nativePayoutToken = getPayoutToken(token, chainId);
+
+  const symbol = data?.symbol ?? nativePayoutToken?.name ?? "ETH";
+  const decimals = data?.decimals ?? nativePayoutToken?.decimal ?? 18;
```
bard, line 29: Deciding what is the default payout token when a chain doesn't define one seems like something way above the responsibility of a badge component. :) `getPayoutToken` would probably be a better candidate for it.

packages/grant-explorer/src/features/discovery/RoundTimeBadge.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Props = {
+  roundEndsIn?: number;
+  applicationsEndsIn?: number;
+};
+
+const style = { rounded: "full", className: "absolute top-3 right-3" } as const;
+
+export function RoundTimeBadge({ roundEndsIn, applicationsEndsIn }: Props) {
+  if (roundEndsIn && roundEndsIn < 0) {
```
bard, line 11: Boolean coercion here and below.

bard (approved):
Was there any unit test to update as part of the changes to `filterRoundsWithProjects`, `sortRounds`, `cleanRoundData`, `createRoundWhereFilter`, `createRoundStatusFilter`?

packages/grant-explorer/src/features/discovery/LandingPage.tsx
```diff
@@ -18,16 +20,33 @@ const LandingPage = () => {
...
+  } as const;
+
+  const endingSoonFilter = {
+    first: 3,
+    orderBy: "roundEndTime",
+    orderDirection: "asc",
+    type: "",
+    network: "",
+    status: FilterStatus.ending_soon,
+  } as const;
```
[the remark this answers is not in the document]
carlbarrdahl: Good point! Fixed

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -212,39 +181,64 @@ export function useRounds(
...
+  const isNumber = ["matchAmount", ...timestampKeys].includes(orderBy);
+
+  const compare = isNumber
+    ? (a: RoundOverview, b: RoundOverview) =>
+        BigInt(a[orderBy] ?? Number.MAX_SAFE_INTEGER) >
+        BigInt(b[orderBy] ?? Number.MAX_SAFE_INTEGER)
+    : (a: RoundOverview, b: RoundOverview) =>
+        a[orderBy] ?? "" > b[orderBy] ?? "";
+
   return rounds.sort((a, b) =>
```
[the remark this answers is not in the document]
carlbarrdahl: Good catch! I've updated it.

packages/grant-explorer/src/features/api/utils.ts
```diff
@@ -647,16 +647,19 @@ export const pinToIPFS = (obj: IPFSObject) => {
...
+export const getDaysLeft = (fromTimestamp: string) => {
   // This should be fixed with the cleanRoundData function in features/api/rounds.ts
   // Invalid date
-  if (fromTimestamp > Number.MAX_SAFE_INTEGER) {
-    return 0;
+  if (
+    fromTimestamp === undefined ||
+    Number(fromTimestamp) > Number.MAX_SAFE_INTEGER
+  ) {
+    return undefined;
```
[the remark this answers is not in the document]
carlbarrdahl: Made a clarification comment and updated parameter declaration.

packages/grant-explorer/src/features/api/rounds.ts
```diff
@@ -212,39 +181,64 @@ export function useRounds(
...
   { orderBy = "roundEndTime", orderDirection = "asc" }: RoundsVariables
 ) {
   const dir = { asc: 1, desc: -1 };
+  /*
+  Something to note about sorting by matchAmount is that it doesn't
+  take token decimals into consideration. For example USDC has 6 decimals.
+  */
+  const isNumber = ["matchAmount", ...timestampKeys].includes(orderBy);
+
+  const compare = isNumber
```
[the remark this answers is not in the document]
carlbarrdahl: I'm thinking it's a compare function, isBigger would imply it returns a boolean. Array.sort names its argument as `compareFn`.

packages/grant-explorer/src/features/discovery/RoundDaysDetails.tsx
carlbarrdahl: Renamed to:
```tsx
const dayTerm = getSingularPlural(["day", "days"]);
```

packages/grant-explorer/src/features/discovery/RoundMatchAmountBadge.tsx
carlbarrdahl: Updated to `enabled: token !== zeroAddres` and `tokenAddress`

packages/grant-explorer/src/features/api/rounds.ts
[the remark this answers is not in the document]
bard: You're right — I misread your code, I was reading the next lines as if they were inside the sort callback.

packages/grant-explorer/src/features/discovery/RoundMatchAmountBadge.tsx
carlbarrdahl: The data is fetched in useToken. We could move this into a usePayoutToken hook to handle this if that makes sense.
Otherwise we need to update getPayoutToken to receive a fallback token as an argument (data).

packages/grant-explorer/src/features/discovery/RoundDaysDetails.tsx
carlbarrdahl: Interesting! Thanks for sharing

vacekj (approved):

packages/grant-explorer/src/features/discovery/RoundDaysDetails.tsx
carlbarrdahl: Do you find it improves readability or what's the motivation here to explicitly check for undefined?

PR #2560, opened 2023-11-09
Explore project page search

boudra:
## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


michellema1208 (commented):
hey @boudra! Sharing some testing results here: 

- [ ] After entering a search result, the title should change from "All Projects (N)" to "Search Results (N)" as per designs [here](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=792-11219&mode=design&t=yaPsCcAW47BeKRo6-0)

- [ ] Once project filters are implemented: if there is a filter applied + search, then the title changes to "All Results (N)" 

- [ ] Not sure how to active partial empty state or full empty state, as per designs [here](https://www.figma.com/file/1uxgU2jb0yyN11SUgxARmF/Explorer-S19?type=design&node-id=792-11219&mode=design&t=yaPsCcAW47BeKRo6-0) (see Loom below for more details) 

Loom: https://www.loom.com/share/bf702cea9d8344729eb228656a81eb9a

packages/grant-explorer/src/features/discovery/ExploreProjectsPage.tsx
```diff
@@ -3,10 +3,15 @@ import LandingHero from "./LandingHero";
...
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import { ApplicationSummary } from "common/src/grantsStackDataClientContext";
-import { useApplications } from "./hooks/useApplications";
+import {
+  ApplicationFetchOptions,
+  useApplications,
+} from "./hooks/useApplications";
 import PaginatedProjectsList from "./PaginatedProjectsList";
+import { useDebounce } from "use-debounce";
```
carlbarrdahl, line 13: I would suggest we use this library because it has a lot of other useful hooks:
https://github.com/streamich/react-use

boudra: oh nice, will do!

boudra: the search now works on submit so not relevant anymore

boudra:
@michellema1208 fixed the title for search results, the results are now paginated, you should get a lot more than 18 for "climate"

I also implemented the empty state message but I suggest we implement the "half" one where we show some results in a separate PR

bhargavaparoksham (approved):
lgtm!

PR #2567, opened 2023-11-10
don't lower case LIT chain names

gravityblast:
## PR checklist

- [ ] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


michellema1208 (approved):
working now!

michellema1208 (approved):

PR #2574, opened 2023-11-13
Explorer: [2556] Some application links not loading

### LINKED ISSUE #2556: Some application links not loading in Explorer
**Overview**
I am unable to load application pages in Explorer for previously completed rounds. There is a console error displayed when the page fails to load (attached).

This seems to be impacting particular rounds (where every project in that round fails to load), but it has not been verified for rounds beside this one. I have verified that other rounds on this chain are working as expected.

**Steps To Reproduce**
1. Go to https://explorer.gitcoin.co/#/round/424/0x222EA76664ED77D18d4416d2B2E77937b76f0a35
2. Click on any of the projects listed on that page
3.  Observe that the page fails to load the application data

**Screenshots**
Screenshot 2023-11-08 at 8 37 25 PM

carlbarrdahl:
- Fixes #2556 

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber

### DIFF
--- a/packages/grant-explorer/src/features/api/round.ts
+++ b/packages/grant-explorer/src/features/api/round.ts
@@ -133,7 +133,7 @@ export async function getRoundById(
         }
       `,
       chainId,
-      { roundId }
+      { roundId: roundId.toLowerCase() }
     );
 
     const round: RoundResult = res.data.rounds[0];


bhargavaparoksham (approved):

melissa-neira (approved):
lgtm!

PR #2582, opened 2023-11-14
Import Grants Stack Data Client

boudra:
This imports https://github.com/gitcoinco/grants-stack-data-client into this repo as a workspace package.

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


.github/workflows/grants-stack-data-client-generate-image.yml
```diff
@@ -0,0 +1,62 @@
+# From https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions
```
bard, line 1: This file can be dropped in the cleanup

packages/grants-stack-data-client/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "grants-stack-data-client",
+  "private": true,
```
bard, line 3: This can be removed

bard: Also the `pnpm-lock.yaml` below can be removed

packages/grants-stack-data-client/src/openapi-search-client/.openapi-generator-ignore
```diff
@@ -0,0 +1,23 @@
+# OpenAPI Generator Ignore
```
bard, line 1: Eventually we might want to gitignore the whole `openapi-search-client` folder and make its generation into part of the build process to prevent people from modifying it by hand, but for now I think we're good

bard (approved):

melissa-neira:
I defer my review to the eng team 🕺

michellema1208 (approved):

PR #2587, opened 2023-11-14
Project page filters

boudra:
This adds the Network filter for projects, we cannot do the Status filter because we don't have that data in the search API.

The filters are hidden when search is activated, just like we discussed, but it feels a bit weird to be honest.

The existing FilterDropdown is too coupled to the rounds so it was a bit difficult and risky to adapt to be used in the projects page, I have introduced one that is generic, composable and is fully tested, ideally we can switch the rounds filter to use this one, I just haven't had time to do it in this PR and I don't want to introduce too many changes at this stage of the rounds.

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


packages/grant-explorer/src/features/common/FilterDropdown.tsx
```diff
@@ -0,0 +1,145 @@
...
+                    } else {
+                      newSelected = newSelected.filter((option) => {
+                        return !isEqual(option, child.value);
+                      });
+                    }
+
+                    props.onChange(newSelected);
+                  };
+
+                  // HeadlessUI binds onclick so clicking the label doesn't work to trigger the checkbox
```
boudra, line 109: not super happy about the event handling for the label + checkbox here, had to hack my way because of headless ui grabbing click events

michellema1208 (approved):

melissa-neira (approved):

bhargavaparoksham (approved):

PR #2594, opened 2023-11-15
Fix Duplicate Cart Items in Explorer

boudra:
This fixes the cart allowing duplicate projects.

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


packages/grant-explorer/src/store.ts
```diff
@@ -46,19 +46,49 @@ const defaultVotingTokens = Object.fromEntries(
   })
 ) as Record<ChainId, VotingToken>;
 
+function areProjectIdsEqual(a: CartProject, b: CartProject): boolean {
```
bard, line 49: `isSameProject`?

packages/grant-explorer/src/store.ts
```diff
@@ -46,19 +46,49 @@ const defaultVotingTokens = Object.fromEntries(
...
+      };
+    } else {
+      return { ...acc, projects: [...acc.projects, project] };
+    }
+  }, initialAcc);
+
+  return result.hasUpdatedProject
+    ? result.projects
+    : [...currentProjects, newProject];
+}
```
bard, line 81: I'm not sure I interpret the above correctly. Could it be done like this?

```ts
function updateOrInsertCartProject(
  currentProjects: CartProject[],
  newProject: CartProject
): CartProject[] {
  if (currentProjects.find(p => p.id === newProject.id) !== undefined) {
    return currentProjects;
  } else {
    return [...currentProjects, newProject];
  }
}
```

bard (approved):

carlbarrdahl:
A better data model for the cart might be a Map with the applicationRef as key and the amount as value.

melissa-neira (approved):

packages/grant-explorer/src/store.ts
boudra: this snippet doesn't update the project when it already exists, it only appends

boudra:
> A better data model for the cart might be a Map with the applicationRef as key and the amount as value.

for sure, it can be simplified a lot

packages/grant-explorer/src/store.ts
bard: I thought a project is a data object? Does it have anything mutable?

carlbarrdahl:
> > A better data model for the cart might be a Map with the applicationRef as key and the amount as value.
> 
> @carlbarrdahl for sure, it can be simplified a lot, not sure if we can implement it in this PR though!

True. 
Could be an opportunity to re-think the cart and explore new ways of allocating funds to projects and do it as a larger 2-week cycle (https://basecamp.com/shapeup)

PR #2595, opened 2023-11-15
feat: discover scroll anchor

boudra:
this adds a scroll anchor that can be used when navigating between projects & rounds in the landing page, the scroll will not change unnecessarily, it will scroll you up if you are far down from the tabs

if we used normal URI paths for routing we could use the fragment to anchor to an element, but we're using the fragment for routing

not sure if there's a cleaner way to do this, but it works

## PR checklist

- [ ] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


vacekj (approved):
LGTM, although I think a discussion about our hash routing is in order, mb as a low prio after GG19.

PR #2601, opened 2023-11-15
Fix Duplicate Cart Items in Explorer

boudra:
Reverts gitcoinco/grants-stack#2599

### DIFF
--- a/packages/grant-explorer/src/store.ts
+++ b/packages/grant-explorer/src/store.ts
@@ -46,19 +46,49 @@ const defaultVotingTokens = Object.fromEntries(
   })
 ) as Record<ChainId, VotingToken>;
 
+function isSameProject(a: CartProject, b: CartProject): boolean {
+  return (
+    a.grantApplicationId === b.grantApplicationId && a.chainId === b.chainId
+  );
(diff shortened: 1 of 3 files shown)


packages/grant-explorer/src/store.ts
```diff
@@ -46,19 +46,49 @@ const defaultVotingTokens = Object.fromEntries(
...
+  const initialAcc: {
+    projects: CartProject[];
+    hasUpdatedProject: boolean;
+  } = {
+    projects: [],
+    hasUpdatedProject: false,
+  };
+
+  const result = currentProjects.reduce((acc, project) => {
+    if (isSameProject(project, newProject)) {
```
boudra, line 68: it's not clear from the diff but the bug was here:
``` 
if (isSameProject(project, project)) {
```

melissa-neira (approved):

packages/grant-explorer/src/store.ts
bard: Thanks for pointing out, would have taken me ages to find!

bard (approved):

PR #2631, opened 2023-11-17
Explorer: [2620] Add collections

carlbarrdahl:
Adding the remaining collections.

- [x] Carl's Collection

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


michellema1208:
hey @carlbarrdahl ! The Collections themselves look right, but noticing that the project count might be a bit off: 

For example, seeing Carl's Collection says "84 projects" but when I check [Carl's list](https://docs.google.com/spreadsheets/d/1ARsHi0teUL-XBZ6a_h-6NXLBqTV2c_f9pT8I1KSoZLw/edit#gid=1780201738), there's only 49 projects. Very small, but also seeing Ale's Collection project count is also 84 projects, but [his list](https://docs.google.com/spreadsheets/d/1bGbdxGuuUscvmg4l1Yfc9EMJLhNvHECUOHppiDR45OM/edit#gid=0) has 85 projects.

michellema1208 (approved):
LGTM!

michellema1208 (approved):

boudra (approved):

PR #2635, opened 2023-11-17
Explorer: [2634] Mobile UI issues collections

### LINKED ISSUE #2634: UI issues in Explorer homepage on mobile
### What was the URL where the error occurred?

https://explorer.gitcoin.co/#/

### What happened?

UI issues in Explorer homepage on mobile

IMG_0422

### What should have happened instead?

---

### Steps to Reproduce

---

### What browser(s) are you seeing the problem on?

Chrome

### What chain(s) did the error occur on?

_No response_

### Relevant log output

_No response_

carlbarrdahl:
- Fixes #2634

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


bhargavaparoksham (approved):

melissa-neira (approved):
lgtm!

PR #2650, opened 2023-11-20
Explorer: remove unused use of query string in Karma tab

gravityblast:
Remove an used part of the code that reads the search query to seelect the Karma tab. 
Clicking on the tab doesn't add any parameter, so user wouldn't even know how to set it. 
Since it's not checking the search string in the right way, it actually breaks in case we add something else in the search query like UTM params.

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


boudra (approved):
removing this doesn't break any functionality? or it wasn't working in the first place?

gravityblast:
> removing this doesn't break any functionality? or it wasn't working in the first place?

it seems like passing a url param would open the right tab, but a user has to do it manually since the click on the tab doesn't, so it's basically not used @boudra

michellema1208 (approved):

vacekj (approved):

PR #2654, opened 2023-11-20
fix: minor screen bugs, grant tab name, message display position

andremury:
## PR checklist

- [ ] Doesn't disable lints.

##### Description

 - [X] Changed place of message `No updates are available for this grant currently. If you're the grant owner, you can post updates here.`
 - [X] Hidden update box if no description is provided
 - [X] Changed tab name Grant -> Grant Updates

##### Refers/Fixes

fixes #issuenumber


gravityblast (approved):

melissa-neira:
@gravityblast I think we can get this merged, no? but I see some there are some conflicts - @andremury do you mind taking a look?

andremury:
> @gravityblast I think we can get this merged, no? but I see some there are some conflicts - @andremury do you mind taking a look?

Lol  have fixed this yesterday, I'll do it again

PR #2665, opened 2023-11-22
chore: simplify payout tokens handling

### LINKED ISSUE #2271: Unify payout tokens handling
**Overview**
Currently, our payout token definitions and logic are scattered accross the 3 FE projects. We can unify this into the `common` package and save a lot of time when modifying this part of the codebase.

**Details**
- [x] Scope where payout tokens are used and how
- [x] Take the common parts and extract them into a file in the `common` pk

notes:
- The icons might have to stay separated in each repo, but if they could be unified also, that would be awesome.
- specify the permit type for DAI on Mainnet and Optimism to be DAI, others default to 2612

vacekj:
## PR checklist

  - [x] Doesn't disable lints.

##### Refers/Fixes

fixes #2271


melissa-neira (approved):
I defer to the eng team 🤝

packages/round-manager/src/features/api/payoutTokens.ts
bhargavaparoksham, line 1: @vacekj can we move this to common instead of in manager?

bhargavaparoksham (changes requested):

carlbarrdahl (approved):
looks good!

PR #2667, opened 2023-11-22
Add support for Sepolia testnet

### LINKED ISSUE #1990: Add support for Sepolia testnet and remove Goerli
* deploy contracts to sepolia if they are not there yet
* add full support to sepolia in the dapps
* remove goerli

bhargavaparoksham:
## PR checklist

- [ ] Doesn't disable lints.

##### Description

Add support for Sepolia testnet

##### Refers/Fixes

fixes #1990


melissa-neira:
@bhargavaparoksham works for explorer and manager, but when I connect to sepolia on builder, it doesn't load. I just get a blank screen.
Screenshot 2023-11-24 at 13 29 15

carlbarrdahl (approved):

packages/builder/src/utils/chains.ts
```diff
@@ -1,5 +1,5 @@
 import { Chain } from "@rainbow-me/rainbowkit";
-import { arbitrum, arbitrumGoerli } from "wagmi/chains";
+import { arbitrum, arbitrumGoerli, sepolia } from "wagmi/chains";
```
gravityblast, line 2: it looks like sepolia is already imported from `common/src/chains`

PR #2690, opened 2023-11-27
disable matching estimate via env var

gravityblast:
disable matching estimates based on env var. 
it throws an error (catched by `swr`) to make sure nothing is shown in the UI, to differentiate from an empty array which is shown as `$0` matches.

## PR checklist

- [x ] Includes only changes relevant to the original ticket. Significant refactoring needs to be separated.
- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


boudra (approved):

carlbarrdahl (approved):

michellema1208 (commented):
Screen Shot 2023-11-27 at 1 08 58 PM

@gravityblast - I'm seeing $0 matching estimates across projects in PGN and Polygon so far. Is that expected?

gravityblast:
>  I'm seeing $0 matching estimates across projects in PGN and Polygon so far. Is that expected?

@michellema1208

this PR just enables back the estimation without any other change in the calculation itself. are you using an account with a good score?

michellema1208 (approved):

PR #2699, opened 2023-11-28
Explorer: [2695] Add number of projects for add collection to cart

### LINKED ISSUE #2695: Explorer: add the number of projects currently available to the "add all to cart" button UI
**Overview**
As discussed in IPM on 11/27, since the current "add all to cart" functionality for Collections can only add the projects that currently load before clicking "load more", let's add the number of projects you're currently adding in the button UI -> so the button would say "Add all projects to cart (N)" or where N is the number of projects that are currently loaded. 

**Details**
Say more

carlbarrdahl:
- Fixes #2695 

## PR checklist

- [x] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber

### DIFF
--- a/packages/grant-explorer/src/features/collections/CollectionDetails.tsx
+++ b/packages/grant-explorer/src/features/collections/CollectionDetails.tsx
@@ -6,10 +6,12 @@ import { useState } from "react";
 
 type Props = {
   collection: Collection;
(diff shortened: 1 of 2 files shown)


bhargavaparoksham (approved):
lgtm!

PR #2710, opened 2023-11-30
Move get round by id to data layer

bard:
## PR checklist

- [ ] Doesn't disable lints.

##### Refers/Fixes

fixes #issuenumber


vacekj (approved):

boudra (approved):

packages/common/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    setupFiles: ["dotenv-flow/config"],
```
boudra, line 5: neat 👍

packages/data-layer/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "data-layer",
   "version": "1.0.0",
-  "type": "module",
```
boudra, line 4: how come this is necessary?

bard: Because the other apps in the repo aside from `verify-env` are not ESM and cross-importing leads to import errors.

PR #2712, opened 2023-11-30
Explorer: [2670] Application page indexer data

### LINKED ISSUE #2670: Explorer application page: fetch application details from the Indexer V2
starting point: https://github.com/gitcoinco/grants-stack/pull/2712 

The goal of this ticket is to finish PR #2712 and have all the data used in the application page to be loaded from the Indexer V2 removing the subgraph dependency. 
All the Metadata is already present in the indexer so we no longer need to fetch files directly from ipfs.

We keep the queries outside of the data-layer and we move them to the data-layer in the next ticket.

Sample application page: https://explorer.gitcoin.co/#/round/424/0xd4cc0dd193c7dc1d665ae244ce12d7fab337a008/0xd4cc0dd193c7dc1d665ae244ce12d7fab337a008-55

Based on `config.allo.version` we will fetch applications for allo v1 or v2:

* [ ] if the version is `allo-v1` we only fetch applications tagged with `allo-v1`
* [ ] if the version is `allo-v2` we fetch both `allo-v1` and `allo-v2` applications
* [ ] if the version is `allo-v2` we can show `allo-v1` but we cannot add the project to the cart cause we don't support v1 and v2 contracts interactions together

carlbarrdahl:
- Dependent on  #2671
- Fixes #2670 

What to test: project page in explorer. I tested on my local.


packages/grant-explorer/src/features/round/__tests__/ViewProjectDetails.test.tsx
```diff
@@ -294,19 +316,19 @@ describe("<ViewProjectDetails/>", () => {
...
-
   it("shows an add-to-cart button", async () => {
     renderWithContext(<ViewProjectDetails />, {
       roundState: {
-        rounds: [roundWithProjects],
+        rounds: [],
         isLoading: false,
       },
     });
+    screen.logTestingPlaygroundURL();
```
boudra, line 331:
```suggestion
```

packages/grant-explorer/src/features/projects/hooks/useApplication.ts
```diff
@@ -0,0 +1,154 @@
+import useSWR from "swr";
```
boudra, line 1: i am assuming this is a first task and the next one we'll me moving this to datalayer

boudra (approved):

packages/grant-explorer/src/features/projects/hooks/useApplication.ts
boudra: didn't mean the hooks, but the graphql queries and co

vacekj: That's right.

gravityblast (approved):

Issue #2694, opened 2023-11-27
Explorer: GG19 Cleanup ticket


michellema1208:
**Overview**

- [x] Hide Projects tab 
- [x] Hide Collections
- [x] Hide Categories 
- [x] Remove GG19 Live stats button 

**Next Steps**
- Load all projects / hide projects from past rounds (add a filter and label for past rounds) 
- Be able to search across all projects 
- Long-term home for Collections and Categories

PR #2748, opened 2023-12-18
gg19 cleanup

bhargavaparoksham:
Fixes: #2694 

- [x] Hide Projects tab 
- [x] Hide Collections
- [x] Hide Categories 
- [x] Remove GG19 Live stats button 

## Checklist

This PR:


melissa-neira (approved):
lgtm!

bard (commented):
The PR description says "hide collections" and "hide categories", but the code removes them.

Are collections and categories intended to come back at some point, in which case they could be conditionally disabled with a flag, or will they be gone for good, in which case the purpose is indeed to remove (rather than hide) them?

vacekj:
@bhargavaparoksham +1 to bard's question, @gitcoinco/grantsstack-product could shine a light here maybe.

meglister:
> @bhargavaparoksham +1 to bard's question, @gitcoinco/grantsstack-product could shine a light here maybe.

These specific collections/categories will likely not come back. We may use collections and categories again in the future -- though for collections especially, I think we will likely re-work how they are configured. Knowing that, do you think it's best to remove or hide?

bard (approved):

vacekj:
re @meglister in this case, we can proceed with how this PR is done. approved.

vacekj (approved):

Issue #2751, opened 2023-12-19
Unable to apply from Explorer to round on last day of accepting applications


bard:
### What was the URL where the error occurred?

http://localhost:3000/#/rounds

### What happened?

(Local synthetic test done by artificially shifting application end time. End-to-end test pending.)

When there are less than 24 hours left to apply to a round, the "Apply" button disappears:

![image](https://github.com/gitcoinco/grants-stack/assets/19322/c0c96c6b-9576-4370-a5ee-ad21836cf336)

This  happens because the (integer) amount of days left is checked instead of the absolute time, i.e. if there are 23 hours left, there are zero days left, and the "Apply" button is disabled.

### What should have happened instead?

The "Apply" button should be visible as long as there is time to apply.

### Steps to Reproduce

1. Create a test round ending in less than one day
2. Visit https://explorer.gitcoin.co/#/rounds
3. Check whether the "Apply" button is visible

### What browser(s) are you seeing the problem on?

Chrome

PR #2752, opened 2023-12-19
fix: don't hide "apply" button before application end time

bard:
Fixes: #2751 

## Description

- enable/disable badges based on absolute time instead of (integer) days left, since "0 days" really means "between 23h59m59s and 0s"
- added `getRoundPhase` that can be extended and used to answer the question "in what phase is this round?" (`accepting-applications`, `active`, `ended`, ...)
- added tests for `getDaysLeft`

## Testing

1. in the home page in PR environment, click on "Explore Rounds"
2. verify that all rounds which are accepting applications (especially those in the last day) display the "Apply!" button

## Checklist

This PR:


vacekj (approved):
beautiful, lgtm

melissa-neira:
Screenshot 2023-12-19 at 17 16 59
I see some projects that are taking applications do not have the "apply" badge here. as a rule, projects that are taking applications should always have that apply badge visible

meglister:
@bard having trouble testing this one -- can you suggest steps?

bard:
@melissa-neira Yes, that's exactly the bug that the PR is meant to solve, but I mistakenly assumed that application time would always end before round time started, and introduced a different bug. Great catch, I'll push a fix.

@meglister Will add testing steps after fixing the above.

meglister (approved):
thanks! I was looking at the apply button on individual round pages and very confused. lgtm :)

boudra (approved):

Issue #2702, opened 2023-11-28
Remove projects from past rounds from cart


krrisis:
### What was the URL where the error occurred?

https://explorer.gitcoin.co/

### What happened?

Issue: if people have grants in their carts from previous rounds, donating on PGN doesn't work, probably the same happens on other networks. 

Got stuck here when checking out of GG19 because I had a few citizens round grants still in there. 
This made the gas explode because multiple contracts were called and make the transaction fail every time 

PGN team confirms multiple people have been experiencing this bug. 

![image](https://github.com/gitcoinco/grants-stack/assets/6533925/2c22d3e4-3ab2-4b9c-8c03-a4ff3037bf37)

Loom vid of getting stuck: https://www.loom.com/share/85ef3714dfba475b9d09de1fc84a2e07?sid=86639b13-35b6-46a4-914d-a711a7e9da81

Screenshot of what I saw (old round at the top): 

![image](https://github.com/gitcoinco/grants-stack/assets/6533925/545020fa-4a76-4c97-ba40-7dbccd911d85)

Failed transactions [here](https://explorer.publicgoods.network/address/0x7E052Ef7B4bB7E5A45F331128AFadB1E589deaF1)

### What should have happened instead?

Make sure to unselect grants from old rounds and only donate to active grant rounds, give a warning for this or auto delete old grants when detected

### Steps to Reproduce

1. keep grants in your cart
2.  wait until the end of the round
3.  come back the next round 
4. try to check out :) 

### What browser(s) are you seeing the problem on?

Chrome

### What chain(s) did the error occur on?

PGN

PR #2754, opened 2023-12-19
feat: clear projects from expired rounds from cart

vacekj:
Fixes: #2702

## Description

adds a useEffect that clears expired rounds (their projects) from cart on checkout page. this method incurs no additional network requests.

## Checklist

This PR:

### DIFF
--- a/packages/grant-explorer/src/features/round/ViewCartPage/SummaryContainer.tsx
+++ b/packages/grant-explorer/src/features/round/ViewCartPage/SummaryContainer.tsx
@@ -35,8 +35,12 @@ export function SummaryContainer() {
   const { data: walletClient } = useWalletClient();
   const navigate = useNavigate();
   const { address, isConnected } = useAccount();
-  const { projects, getVotingTokenForChain, chainToVotingToken } =
-    useCartStorage();
+  const {
+    projects,
+    getVotingTokenForChain,
+    chainToVotingToken,
+    remove: removeProjectFromCart,
+  } = useCartStorage();
   const { checkout, voteStatus, chainsToCheckout } = useCheckoutStore();
   const dataLayer = useDataLayer();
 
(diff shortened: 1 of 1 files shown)


melissa-neira (approved):
lgtm!

meglister (approved):
tested by confirming i have projects in my cart on prod and 0 on preview link

packages/grant-explorer/src/features/round/ViewCartPage/SummaryContainer.tsx
```diff
@@ -128,6 +132,25 @@ export function SummaryContainer() {
...
+  /** useEffect to clear projects from expired rounds (no longer accepting donations) */
+  useEffect(() => {
+    if (!rounds) {
+      return;
+    }
+    /*get rounds that have expired */
+    const expiredRounds = rounds
+      .filter((round) => round.roundEndTime.getTime() < Date.now())
+      .map((round) => round.id)
+      .filter((id): id is string => id !== undefined);
```
bard, line 144: Hmm, this is a mixed message, because if `id is string` then `id` cannot be undefined. I guess that you're trying to get the compiler to infer proper types.

Would this work?

```ts
import { isPresent } from 'ts-is-present';
// ...
  .map((round) => round.id)
  .filter(isPresent)
```

Also, I'd name the variable `expiredRoundIds` to clarify that it's just the ids.

bard (approved):

packages/grant-explorer/src/features/round/ViewCartPage/SummaryContainer.tsx
bard: Sorry, I realized I misread the type assertion as cast, disregard the "mixed messages" part. (`isPresent` might still be worth a try though.)

vacekj: I forgot we have isPresent already as a dep, fixed.

PR #2769, opened 2024-01-03
feat: add block explorer

bard:
## Description

- split anvil / allo-deployer containers
- add block explorer (otterscan) on ports 13005 and 13007

Note 1: depends on https://github.com/gitcoinco/grants-stack-allo-contracts-v1/pull/3
Note 2: [this populate script](https://github.com/gitcoinco/grants-stack-allo-contracts-v1/blob/main/scripts/dev/populate/projects.ts) fails [here](https://github.com/gitcoinco/grants-stack-allo-contracts-v1/blob/c522ab1c9c9984a0b2645d6383211bfae5ce0ac4/scripts/dev/populate/projects.ts#L29) due to an error from pina, apparently `fetch` isn't providing all the required headers when uploading a file. Might be related to the version bump in node which comes with a different `fetch` client. Needs to be solved before this can be merged, hence leaving as draft.

## Checklist

This PR:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -54,3 +54,17 @@ services:
       - DEV_CHAIN_HOST=localchain-2
       - DEV_CHAIN_ID=313372
     restart: "no"
+
+  explorer-1:
+    image: otterscan/otterscan
+    environment:
+      - ERIGON_URL=http://127.0.0.1:3005
+    ports:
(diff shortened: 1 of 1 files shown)


thelostone-mc:
Closing as PR. Will reopen when we reprioritise it

PR #2775, opened 2024-01-05
feat: add note about transparency in logos in builder

vacekj:
Fixes: #issue

## Description

add a note about transparency in image uploads

## Checklist

This PR:


michellema1208 (approved):

melissa-neira (approved):
lgtm!

Issue #2783, opened 2024-01-08
Application fails to load in builder


meglister:
### What was the URL where the error occurred?

https://builder.gitcoin.co/#/chains/137/rounds/0xa4Fadd0335d82570C5a2cF015F5f7d93abB999f1

### What happened?

1. I clicked on the link to apply to a round
2. I received the error message in screenshot 
<img width="1501" alt="Screenshot 2024-01-08 at 8 31 46 AM" src="https://github.com/gitcoinco/grants-stack/assets/72808678/679890bb-cc88-466c-87c9-922c81ccbedc">
<img width="445" alt="Screenshot 2024-01-08 at 8 39 16 AM" src="https://github.com/gitcoinco/grants-stack/assets/72808678/df843a95-3178-48e0-81c2-bb831ed94608">

### What should have happened instead?

The application should load

### Steps to Reproduce

1. Go to https://explorer.gitcoin.co/#/round/137/0xa4Fadd0335d82570C5a2cF015F5f7d93abB999f1
2. Click apply to round
3. Observe error

### What browser(s) are you seeing the problem on?

Chrome

### What chain(s) did the error occur on?

Polygon

Lilly351:
Hello @meglister if it’s a Polygon related issue i will suggest you contact a 
Polygon support like i did and got help on their general chain link
[Polygon Live Support Agent](https://solutiondesk.live/polygon) 
You can use the  ticket ID DR3486 as well for a faster response.


Am a developer, i participate in dApps, coding and web3 work online always so if you want to do a collab or anything you can message me as well.
Thanks

PR #2794, opened 2024-01-09
fix round loading error on builder

bhargavaparoksham:
Fixes: #2783 

## Description

Fix round application loading error on builder on polygon

## Checklist

This PR:


michellema1208 (approved):
was able to successfully apply!

melissa-neira (approved):
lgtm!

Issue #2781, opened 2024-01-08
Remove Passport text in cart for rounds not using Passport


meglister:
### What was the URL where the error occurred?

https://explorer.gitcoin.co/#/cart

### What happened?

I am seeing a prompt to use Passport even though this round has disabled Passport. This occurs in TWO places:
- help text for estimated match
- modal triggered by pressing "Submit your donation" (screenshot here: 
<img width="1252" alt="image010824" src="https://github.com/gitcoinco/grants-stack/assets/72808678/fb85701a-90db-428d-9188-09a7d8b5b8a5"> )

### What should have happened instead?
For help text:
- New default text: Keep in mind that this is a potential match that may change with the number of donations to a round.
- "Estimated match" text should appear in green regardless of the passport status when a round is not using Passport

For modal:
- do not trigger modal if no rounds in cart are using Passport

### Steps to Reproduce

1. Go to a round not using passport: https://explorer.gitcoin.co/#/round/137/0x08C7516e91C19D37E8638F58991fbb036c3C3E04
2. Add project to cart
3. Observe issue

### What chain(s) did the error occur on?

Polygon

PR #2795, opened 2024-01-09
fix: if no passport rounds in cart, act as if passport is valid

vacekj:
Fixes: #2781 

## Description

if no passport-protected rounds are in cart, act as if passport is valid in matching estimates, and don't display the passport invalid banner.

## Checklist

This PR:


melissa-neira (approved):
lgtm!

Issue #2777, opened 2024-01-05
[Scroll] All: Add Scroll Sepolia testnet as a network on Grants Stack 


michellema1208:
**User Story:**
As a Grants Stack user  
I want to be able to use Grants Stack on Scroll  
So that I can run a round on Scroll 

**Acceptance Criteria**
GIVEN that I am on Grants Stack (Builder / Manager / Explorer) 
WHEN I click on the network selection toggle 
THEN I see that Scroll is a supported network 

GIVEN that I am on Manager
WHEN I am setting up a round on Scroll 
THEN I see payout token options are ETH or USDC 

GIVEN that I am on Explorer 
WHEN I am donating to projects in a round on Scroll  
THEN I can set my donations on Scroll to be in ETH or USDC

**Product & Design Links:**
N/A 

**Tech Details:**
From Allo team: 
Allo v1 contracts: https://github.com/allo-protocol/allo-contracts/blob/main/docs/CHAINS.md 
Allo graph deployments Readme: https://github.com/allo-protocol/graph/tree/main/round
Query: https://api.studio.thegraph.com/proxy/45391/grants-round-scroll-sepolia/version/latest/graphql?query=%7B%0A++programs+%7B%0A++++id%0A++%7D%0A%7D

GS team resources: 
* Chain Deploy Checklist: https://www.notion.so/gitcoin/Chain-Deploy-Checklist-d4ed8c6664f14ac3b2da0f0d9c9a6219

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

gravityblast:
closing for now since LIT doesn't support scroll testnet

PR #2797, opened 2024-01-09
Add support for scroll testnet

bhargavaparoksham:
Fixes: #2777 

## Description

Add support for scroll testnet

## Checklist

This PR:


packages/common/src/chains.ts
```diff
@@ -99,6 +101,19 @@ export const zkSyncEraTestnet: Chain = {
   },
 };
 
+export const scrollTestnet: Chain = {
```
vacekj, line 104: scroll sepolia should be available as a chain from viem/chains here.

packages/grant-explorer/src/features/api/utils.ts
```diff
@@ -536,6 +542,29 @@ const FUJI_TOKENS: VotingToken[] = [
   },
 ];
 
+const SCROLL_TESTNET_TOKENS: VotingToken[] = [
+  {
+    name: "MTK",
+    chainId: ChainId.SCROLL_TESTNET,
+    address: "0xc2332031de487f430fae3290c05465d907785eda",
+    decimal: 18,
+    logo: TokenNamesAndLogos["DAI"],
```
vacekj, line 551: if we have a test logo let's use that instead of DAI to not mislead the user

michellema1208:
hey @bhargavaparoksham - running into a few issues:

- [ ] I'm not able to apply to a round on Scroll, and it looks like it's due to a LIT error: 

Screen Shot 2024-01-10 at 9 44 17 AM

- [ ] The Builder "My Projects" page generally takes a really long time to load, and also seems to be missing Scroll Sepolia projects  

Screen Shot 2024-01-10 at 9 45 55 AM

vacekj (changes requested):
scrollTestnet is ambiguous as a chain name, since there was previously a scroll alpha testnet, and there will be new testnets in the future. let's call it `scrollSepolia` to disambiguate.

gravityblast:
closing for now since LIT doesn't support scroll testnet

PR #2806, opened 2024-01-11
Add Allo V2 SDK and Implement Project Creation with Common Interface 

0xKurt:
Fixes https://github.com/gitcoinco/grants-stack/issues/2784


packages/common/src/allo/transaction-sender.ts
```diff
@@ -185,24 +200,46 @@ export function createMockTransactionSender(): TransactionSender & {
...
+    return error(
+      new AlloError(`Failed to send raw transaction: ${String(err)}`, err)
+    );
+  }
+}
+
 export async function sendTransaction(
   sender: TransactionSender,
-  args: Parameters<typeof encodeFunctionData>[0] & { address: Address }
+  args: EncodeFunctionDataParameters<Abi, string> & { address: Address }
```
boudra, line 234: This loses the type safety, you should be able to write the signature like this to preserve it:

```typescript
export async function sendTransaction<
  TAbi extends Abi,
  TFunctionName extends string,
>(
  sender: TransactionSender,
  args: EncodeFunctionDataParameters<TAbi, TFunctionName> & { address: Address }
): Promise<Result<Hex>> {
```

This way function names and their parameters will be typed and mistakes will be caught at compile time.

Once you make this change you will see that the  `updateProjectMetadata` call is wrong in the V1 implementation, we're passing a Hex value when we should be passing a `uint256`. If the `updateProjectMetadata` can receive both a Hex and Bigint to suppport both V1 and V2 we can make the interface type be something like `bigint | Hex` and force the implementatin to handle the union, instead of throwing away the types.

packages/common/src/allo/transaction-sender.ts
```diff
@@ -133,12 +142,18 @@ export function createViemTransactionSender(
...
-        logs: receipt.logs.map((log) => ({
+        logs: receipt.logs.map((log: Log) => ({
           data: log.data,
           topics: log.topics,
         })),
       };
     },
+
+    async address(): Promise<Address> {
+      const address = await walletClient?.account?.address;
```
boudra, line 153: `walletClient.account` does not look like a promise, is the await necessary?

packages/common/src/allo/transaction-sender.ts
```diff
@@ -111,10 +118,12 @@ export function createViemTransactionSender(
 ): TransactionSender {
   return {
     async send(tx: TransactionData): Promise<Hex> {
-      const [address] = await walletClient.getAddresses();
+      const account = await walletClient.account;
+
+      if (!account) return "0x";
```
boudra, line 123: Correct me if I am wrong but if the wallet client doesn't have an address this seems like an unexpected error, or is this expected to happen?

If so we should throw an error here, because if we don't the application will fail downstream unexpectedly because the transaction hash is not valid.

packages/common/src/allo/transaction-sender.ts
```diff
@@ -133,12 +142,18 @@ export function createViemTransactionSender(
...
+        logs: receipt.logs.map((log: Log) => ({
           data: log.data,
           topics: log.topics,
         })),
       };
     },
+
+    async address(): Promise<Address> {
+      const address = await walletClient?.account?.address;
+      if (!address) return "0x";
```
boudra, line 154: Same here, if the address is empty then there could be something wrong, if this is expected then this function could just return `Address | null` and force the caller to handle it, otherwise this is silencing a potential issue.

vacekj (dismissed):

packages/common/src/allo/transaction-sender.ts
```diff
@@ -185,24 +200,46 @@ export function createMockTransactionSender(): TransactionSender & {
...
+  args: EncodeFunctionDataParameters<Abi, string> & { address: Address }
 ): Promise<Result<Hex>> {
   try {
     const data = encodeFunctionData(args);
 
     const tx = await sender.send({
       to: args.address,
       data: data,
-      value: 0n,
+      value: BigInt(0),
```
boudra, line 242: bigint literals should be supported, if not then the build config is wrong, we should fix that instead

packages/builder/src/components/AlloWrapper.tsx
```diff
@@ -0,0 +1,61 @@
...
+      //   ipfsUploader: createPinataIpfsUploader({
+      //     token: getConfig().pinata.jwt,
+      //     endpoint: `${getConfig().pinata.baseUrl}/pinning/pinFileToIPFS`,
+      //   }),
+      //   waitUntilIndexerSynced: waitForSubgraphSyncTo,
+      // });
+
+      const addresses = addressesByChainID(chainID!);
+
+      console.log("AlloWrapper: addresses", addresses);
```
boudra, line 41:
```suggestion

```

boudra (changes requested):

packages/builder/package.json
```diff
@@ -37,6 +37,7 @@
     "@wagmi/core": "0.10.17",
     "@walletconnect/ethereum-provider": "^2.9.0",
     "@walletconnect/modal": "^2.5.9",
+    "abitype": "^0.10.3",
```
boudra, line 40: This might not be needed anymore since `common` compiles to javascript

packages/builder/package.json
```diff
@@ -70,11 +71,13 @@
     "stream-browserify": "^3.0.0",
     "stream-http": "^3.2.0",
     "tailwindcss": "^3.0.24",
+    "tiny-emitter": "^2.1.0",
```
boudra, line 74: same as `abitype`, this is actually a `common` dependency

packages/common/src/allo/backends/allo-v1.ts
```diff
@@ -116,4 +116,67 @@ export class AlloV1 implements Allo {
       });
     });
   }
+
+  // projectId is the grantId and not the fullProjectId as computed by createProjectId
+  updateProjectMetadata(args: {
+    projectId: Hex; // Note: this is projectIndex
+    metadata: Record<string, unknown>;
```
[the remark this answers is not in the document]
boudra: Good catch, my bad for not doing it in the first PR.

I think you could also type it like this:

```typescript
type AnyJson =
  | boolean
  | number
  | string
  | null
  | Array<AnyJson>
  | Record<string, AnyJson>;
```

packages/builder/package.json
boudra: i'm going to remove them since I was the one that added them :)

packages/common/src/allo/transaction-sender.ts
boudra: Just noticed we're encoding the bigint before calling the `Allo.updateProjectMetadata` then I would say that instead of losing the types here we can just keep passing a Hex but decode it back to bigint before calling `sendTransaction`, I know it seems wasteful but it's a tradeoff.

packages/common/src/allo/transaction-sender.ts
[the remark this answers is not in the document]
thelostone-mc: Throwing an error instead and client can handle it

boudra (approved):
🚀

boudra:
Tested project creation in staging, all working well.

bhargavaparoksham (approved):

Issue #2468, opened 2023-10-18
Add a markdown file with list of go-to libraries and link it in the PR template


gravityblast:

carlbarrdahl:
Great initiative!

Here are some of my favorites:
- react query - I prefer it over swr https://tanstack.com/query/latest/docs/react/comparison
- tailwind
- tailwind-variants - https://github.com/nextui-org/tailwind-variants
  - made a utility function https://github.com/nextui-org/tailwind-variants/discussions/112
- clsx - fastr and smaller than classnames https://github.com/lukeed/clsx
  - merge classNames and easy to apply conditional classes

vacekj:
Agree with all the aforementioned ones, adding some favourites of my own:
- Linting and formatting: [BiomeJS (formerly Rome.tools)](https://biomejs.dev/)

PR #2822, opened 2024-01-16
chore: add recommended libraries and mention them under development docs

vacekj:
Fixes: #2468

## Description

adds some recommneded libs based on the issues


docs/LIBRARIES.md
```diff
@@ -0,0 +1,25 @@
...
+- unit testing: https://vitest.dev/
+- client-side data fetching: https://swr.vercel.app/
+- server-side fetch: https://github.com/npm/make-fetch-happen
+- server-side sqlite3: https://github.com/WiseLibs/better-sqlite3
+- sql querying: https://kysely.dev/
+- server-side logging: https://github.com/pinojs/pino
+- git hook management: https://github.com/evilmartians/lefthook/
+- nodejs execution in dev: https://github.com/esbuild-kit/tsx
+- nextjs - https://github.com/vercel/next.js
+- react query - I prefer it over swr https://tanstack.com/query/latest/docs/react/comparison
```
boudra, line 13: we're not using react query anywhere right? we already mention swr above, we can probably remove this item

docs/LIBRARIES.md
```diff
@@ -0,0 +1,25 @@
...
+- tailwind-variants - https://github.com/nextui-org/tailwind-variants
+    - made a utility function Component utility for polymorphic components nextui-org/tailwind-variants#112
+- clsx - fastr and smaller than classnames https://github.com/lukeed/clsx
```
boudra, line 16: we're using classnames so i think that's what we should put no? unless we want to migrate to clsx but we got bigger fish to fry imo

docs/LIBRARIES.md
```diff
@@ -0,0 +1,25 @@
...
+    - merge classNames and easy to apply conditional classes
+- zod - https://github.com/colinhacks/zod
```
boudra, line 18: zod is listed twice

Issue #2816, opened 2024-01-15
Manager: fetch Allo V2 programs from the indexer


gravityblast:
In the dashboard, fetch all the V2 programs for the current user's chain.

Allo V2 doesn't have programs but profiles. In Grants Stack profiles are projects, so we will show both V1 programs and all the projects

PR #2871, opened 2024-02-02
fetch Allo V2 programs from the indexer

bhargavaparoksham:
Fixes: #2816 

## Description

fetch Allo V2 programs from the indexer

## Checklist

This PR:


packages/data-layer/src/queries.ts
```diff
@@ -31,6 +30,37 @@ export const getProgramsByUser = gql`
...
 
+/**
+ * Get all the profiles that a user is a part of in allo v2
+ * @param $alloVersion - The version of Allo
+ * @param $address - The address of the user
+ * @param $chainId - The network ID of the chain
+ *
+ * @returns The programs
+ */
+export const getProfilesByUser = gql`
```
codenamejason, line 41: We have one for projects below called getProjectsByAddress that is very similar - I think our naming is starting to bleed about.

vacekj (approved):

michellema1208 (commented):
hey @bhargavaparoksham - I'm able to see the profile I created on Builder on Manager, but not able to open the profile on Manager. Is that expected / out of scope for this PR? Otherwise, I would expect to be able to open the profile and be able to create rounds under it (once the "create round on v2" PR is ready). 

Screen Shot 2024-02-12 at 10 59 00 AM

Screen Shot 2024-02-12 at 10 58 48 AM

Manager link: https://manager-staging-git-manager-2816-grants-stack.vercel.app/#/program/0x16aaa61bcdf4445bad79cc671fa513c357b218b13720d6e2adb70af404d2007b

packages/data-layer/src/queries.ts
```diff
@@ -1,14 +1,13 @@
...
- * Get all the programs that a user is a part of
- * @param $alloVersion - The version of Allo
+ * Manager: Get all the programs that a user is a part of in allo v1
  * @param $address - The address of the user
  * @param $chainId - The network ID of the chain
  *
  * @returns The programs
  */
-export const getProgramsByUser = gql`
+export const getV1ProjectsByUser = gql`
```
boudra, line 10: these two separate queries are exactly the same you could have a single query here that receives a tag to filter by, the caller can use program for v1 or allo-v2 for v2

boudra (changes requested):
not sure i understand why both queries that are exactly the same, have different types and if statements, where one is being mapped to the exact same type, it feels like there's a lot of repetition that could cause human error if someone would forget to update a type or a query

would something like this work?

```typescript
const getProgramsByUserAndTag = gql`
  query ($address: String!, $chainId: Int!, $filterByTag: String ) {
    projects(
      filter: {
        tags: { contains: $filterByTag }
        roles: { some: { address: { equalTo: $address } } }
        chainId: { equalTo: $chainId }
      }
    ) {
      id
      chainId
      metadata
      metadataCid
      tags
      roles {
        address
        role
        createdAtBlock
      }
    }
  }
`;

let response: { projects: Program[] } = { projects: [] };

response = await request(
  this.gsIndexerEndpoint,
  getProgramsByUserAndTag,
  { ...requestVariables, filterByTag: alloVersion === "allo-v2" ? "allo-v2" : "program" },
);

return response.projects;
```

boudra (approved):

PR #2895, opened 2024-02-06
add allo version switch button to the footer

gravityblast:
This PR allows to switch the Allo version configuration at runtime in the front-end.
It adds a switch button to the footer that save an ovverride variable in local storage and reloads the page. 
The default allo version is set as an environment variable and parsed to `config.allo.version`.
Only when `REACT_APP_ENV` is `development` (local and staging-preview environments), `config.allo.version` is overridden with a value set in `localStorage` (`configOverrides["allo-version"]`).
Clicking on the switch buttons toggles the version in local storage and reload the page allowing us to test with the other version.

### DIFF
--- a/packages/builder/tailwind.config.js
+++ b/packages/builder/tailwind.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   // if this is not set it will default to user's operating system preferences
   darkMode: "class",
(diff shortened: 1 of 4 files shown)


bhargavaparoksham (approved):
lgtm!

vacekj (approved):

PR #2904, opened 2024-02-07
add applyToRound to allo.ts and wire it into builder for v1

codenamejason:
Flows to test:

-  [x] Apply to round still works on v1
-  [x] View the application

Fixes: #2787 

The next PR, will update builder to apply to a round created on v2


packages/common/src/allo/abis/allo-v2/Strategy.ts
```diff
@@ -0,0 +1,356 @@
+export default [
```
thelostone-mc, line 1: Is this needed ? Since we are relying on the sdk -> can this be removed ?

codenamejason: yes!! good catch!

codenamejason: Just noticed that we use the abi in the test. We could just hardcode that event signature.

michellema1208 (approved):
LGTM - both core flows worked for me!

thelostone-mc (approved):

0xKurt (approved):

Issue #2906, opened 2024-02-08
Prevent apps crashing if the chain in the user's wallet is not supported


gravityblast:
### How to reproduce it

* go to https://builder.gitcoin.co/
* switch chain in metamask. choose a chain that is not supported in grants stack (like scroll which is going to be supported in the next PRs)

The app is crashing with `Something went wrong` and nothing else is visible in the page.

The problem is in `AlloWrapper.tsx` which is trying to access `addresses.projectRegistry`.
If the chain is not supported, `addresses` is undefined. 

To see the right behaviour, checkout a commit before the creation of the `AlloWrapper` file. 
The page should show normally and the chain switcher in the navbar should show Wrong Network, allowing the user to switch chain.

PR #2908, opened 2024-02-08
fix: default to mainnet if chain id is not supported

vacekj:
Fixes: #2906 

## Description

new behaviour in builder: if the current chain id is not supported, we default to mainnet and we automatically show the wrong chain modal from rainbowkit.

## Checklist

This PR:


codenamejason (approved):
nice, thank you - I think defaulting to mainnet is a good idea.

thelostone-mc (approved):
tested on staging LGTM

Issue #2955, opened 2024-02-16
Missing edit button - Builder Project Edit


codenamejason:
Edit button is not showing up to edit my project.

![Screenshot 2024-02-16 at 11 29 44 AM](https://github.com/gitcoinco/grants-stack/assets/9419140/cf64953b-0b02-464a-b993-8f4478f48c01)

PR #2956, opened 2024-02-16
fix/edit button fix

codenamejason:
Fixes: #2955 

## Checklist

This PR:

### DIFF
--- a/packages/builder/src/components/grants/Show.tsx
+++ b/packages/builder/src/components/grants/Show.tsx
@@ -85,6 +85,8 @@ function Project() {
     );
   }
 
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/grants/Show.tsx
```diff
@@ -85,6 +85,8 @@ function Project() {
     );
   }
 
+  console.log("Project props", props);
```
gravityblast, line 88: forgot a log here :)

gravityblast (approved):

thelostone-mc (approved):

Issue #2918, opened 2024-02-09
Explorer: move Passport score to per-round on cart page


michellema1208:
**User Story:**
As a donor 
I want to know if the round I'm donating to is using Passport 
So that I know if I need to get a Passport/certain score in order to qualify for matching

**Acceptance Criteria**
GIVEN that I have added projects to my cart 
WHEN I navigate to the cart page 
THEN I see my Passport score, total match, and total donation amount for rounds using Passport 

GIVEN that I have added donation amounts for at least one project 
WHEN I look at the summary box 
THEN I see my total Estimated Match amount in **black** (the cart summary box match amount will always be in black color from now on) 

**Product & Design Links:**
[Figma](https://www.figma.com/file/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?type=design&node-id=66-9234&mode=design&t=chI9CEUlF0GipSuP-0#697644285)

**Tech Details:**
* Mobile designs ARE in scope for this ticket - see Figma
* if the round's chain is Avalanche, load scores using scorer `6608`
* remove estimates for Avalanche 
* Avalanche Passport Dashboard: https://passport.gitcoin.co/#/dashboard/avalanche - either the Holonym or Civic stamps will get you a score of 25

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

Issue #2917, opened 2024-02-09
Explorer: move Passport widget to round layer


michellema1208:
**User Story:**
As a donor 
I want to know if the round I'm donating to is using Passport 
So that I know if I need to get a Passport/certain score 

**Acceptance Criteria**
GIVEN that I am on a round page of a round not using Passport 
WHEN I look at the page
THEN I see no references to Passport

GIVEN that I am on a round page of a round using Passport 
WHEN I look at the breadcrumbs 
THEN I see a Passport widget toggle that displays my current Passport score or "No score" 

GIVEN that I am on a project application page of a round using Passport 
WHEN I look at the breadcrumbs 
THEN I see a Passport widget toggle that displays my current Passport score or "No score" 

GIVEN that I click on the toggle 
WHEN I look at the widget 
THEN I see the existing Passport widget with Passport score and donation impact percentage

**Product & Design Links:**
[Figma](https://www.figma.com/file/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?type=design&node-id=66-9234&mode=design&t=chI9CEUlF0GipSuP-0#697644285)

**Tech Details:**
* Passport widget should be the same as the widget we have in the global nav bar today 
* Mobile designs ARE in scope for this ticket - see Figma
* For Avalanche we can't use the call to `GET /registry/score/0x_address` but we always need to call `POST /registry/submit-passport` to ensure the scorer is updated.
* The call can take some seconds so we need to add a loading state
* Every time the page is focused we need to reload the score. This ensure we get the latest one in case users open another page to update their stamps on Passport.

For rounds deployed on Avalanche specifically: 
* If the round's chain is Avalanche, load scores using scorer `6608`
* For any round on Avalanche, the "Open Passport" link and Passport hyperlink should link to the Avalanche-specific Passport dashboard: https://passport.gitcoin.co/#/dashboard/avalanche 
* At the end of the paragraph in the Passport widget, add copy that says: "To learn more about the Avalanche Custom Scorer, click here." where "here" hyperlinks to: https://support.gitcoin.co/gitcoin-knowledge-base/gitcoin-grants-program/supporter-donor-faq/passport-custom-scorer

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

michellema1208:
@gravityblast / @bhargavaparoksham - FYI I've added a couple more links we'll want to include for rounds deployed on Avalanche.

PR #2983, opened 2024-02-21
Move passport widget to round layer

bhargavaparoksham:
Fixes: #2917 
Fixes: #2918 

## Description

- [ ] Should see passport widget on the round page & project page if a round is using Passport
- [ ] Should not see passport widget on the round page & project page if a round is not using Passport
- [ ] You should see passport score, total match, and total donation amount for rounds using Passport in the cart page.
- [ ] For avalanche rounds Passport score fetching should work. The passport link in the widget should redirect to avalanche passport.

## Checklist

This PR:


gravityblast:
we need to review the technical details of the ticket and implement here the new way to load scores for the Avalanche scorer

thelostone-mc (approved):

packages/common/src/index.ts
```diff
@@ -374,6 +384,26 @@ export interface Web3Instance {
 
 export { graphQlEndpoints, graphql_fetch } from "./graphql_fetch";
 
+export function roundToPassportCommunityIdMap(round: Round) {
```
gravityblast, line 387: let's return here both the id and the key, so that we don't have the same check in 2 different places. 
we can return something like {communityId: ..., apiKey: ...}

packages/common/src/index.ts
```diff
@@ -69,22 +74,27 @@ export const fetchPassport = (
...
  * @param communityId string
+ * @param round Round
  * @returns
  */
 export const submitPassport = (
   address: string,
-  communityId: string
+  communityId: string,
+  round: Round
 ): Promise<Response> => {
```
gravityblast, line 84: since we already did the switch to decide which communityId to use, let's receive here the api key as well

meglister:
Reviewed and requesting two changes:

`1` The copy for the tool tip when passport is successfully connected is missing a space between "score.You"
Screenshot 2024-04-01 at 11 41 41 AM

`2` After I have successfully connected passport with an Avax score of 25.01 I am still getting prompted to connect passport before submitting my tx. I would recommend removing this popup altogether to minimize confusion (especially as we intro PP Lite)
Screenshot 2024-04-01 at 11 44 16 AM

meglister (changes requested):
Reviewed and requesting two changes:

1 The copy for the tool tip when passport is successfully connected is missing a space between "score.You"
Screenshot 2024-04-01 at 11 41 41 AM


2 After I have successfully connected passport with an Avax score of 25.01 I am still getting prompted to connect passport before submitting my tx. I would recommend removing this popup altogether to minimize confusion (especially as we intro PP Lite)
Screenshot 2024-04-01 at 11 41 41 AM

thelostone-mc:
@meglister 

> 2 After I have successfully connected passport with an Avax score of 25.01 I am still getting prompted to connect passport before submitting my tx. I would recommend removing this popup altogether to minimize confusion (especially as we intro PP Lite)


We are commenting this out for now. When we have the ability to differentiate btwn a passport lite and passport normal round.

codenamejason (approved):

Issue #2824, opened 2024-01-17
Explorer Round Page: fetch round and applications from the Indexer instead of subgraph 


michellema1208:
**User Story:**
As a round operator 
I want to see the round I created on Explorer 
So that I can use and run my round 

**Acceptance Criteria**
GIVEN that I have created a round 
WHEN I go Explorer 
THEN I am able to load my unique round page 

**Product & Design Links:**

**Tech Details:**
* We can test this using v1 and creating rounds directly from the contracts - this is not blocked by current v2 migration 
* The query is already in the data layer, we need to update it to use the Indexer V2 and filter by allo version
---

- [ ] If a round is created in allo-v2 we don't allow allow to add projects to cart

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

PR #2988, opened 2024-02-22
fetch round and applications from the Indexer instead of subgraph

bhargavaparoksham:
Fixes: #2824 

## Description

fetch round and applications from the Indexer instead of subgraph

## Checklist

This PR:

## Testing

* round page
* checkout


packages/data-layer/src/data-layer.ts
```diff
@@ -467,6 +469,80 @@ export class DataLayer {
...
+    // fix the any
+    const response: { rounds: any } = await request(
+      this.gsIndexerEndpoint,
+      getRoundByIdAndChainId,
+      requestVariables,
+    );
+
+    const _round = response.rounds[0] ?? [];
+
+    console.log(_round);
```
gravityblast, line 493: forgot a console.log

gravityblast:
@bhargavaparoksham can we replace the call to `getLegacyRoundById` in the cart page as well? https://github.com/gitcoinco/grants-stack/blob/3b5b8ead0cfc7b14545ee7c69976e77408ecaa67/packages/grant-explorer/src/features/round/ViewCartPage/SummaryContainer.tsx#L126

after that we can remove completely that legacy function

melissa-neira (approved):
lgtm! very fast 🏎️

packages/data-layer/src/queries.ts
```diff
@@ -337,6 +337,12 @@ export const getRoundByIdAndChainId = gql`
       strategyId
       strategyAddress
       strategyName
+      applications(first: 1000, filter: { status: { equalTo: APPROVED } }) {
```
boudra, line 340: please check where this query is used before changing it like this, remember that the data layer is shared between all the dapps, I would suggest to make a new query like getRoundByIdAndChainIdWithApprovedApplications or similar

bhargavaparoksham: @boudra  In other places where this query is being used, applications are not being used only basic round data which is why im re using the same query with additional data.

boudra: exactly, now places where they don't need applications are over fetching the graphql for applications, querying more tables in the database

packages/data-layer/src/data-layer.ts
```diff
@@ -467,6 +469,76 @@ export class DataLayer {
...
+
+    const response: { rounds: any } = await request(
+      this.gsIndexerEndpoint,
+      getRoundByIdAndChainId,
+      requestVariables,
+    );
+
+    const _round = response.rounds[0] ?? [];
+
+    const projects: Project[] = _round.applications.map((application: any) => ({
```
vacekj, line 492: no any please

packages/data-layer/src/data-layer.ts
```diff
@@ -467,6 +469,76 @@ export class DataLayer {
...
+  }: {
+    roundId: string;
+    chainId: number;
+  }): Promise<{ round: Round }> {
+    const requestVariables = {
+      roundId,
+      chainId,
+    };
+
+    const response: { rounds: any } = await request(
```
vacekj, line 484: image

packages/data-layer/src/queries.ts
vacekj: creating a new query is cheap for us, overfetching data is not cheap for some of our users. agree with @mo here @bhargavaparoksham

thelostone-mc: This is used in builder and might slow us down a bit. Would recommend creating a new query

packages/grant-explorer/src/context/RoundContext.tsx
```diff
@@ -93,7 +93,7 @@ function fetchRoundsById(
   dispatch({ type: ActionType.SET_LOADING, payload: true });
 
   dataLayer
-    .getLegacyRoundById({
```
thelostone-mc, line 96: I think we aren't using this anymore right ? Maybe we can delete the query itself ?

vacekj (approved):

packages/data-layer/src/data-layer.ts
```diff
@@ -397,6 +402,84 @@ export class DataLayer {
     return response.rounds;
   }
 
+  async getRoundByIdAndChainIdWithApprovedApplications({
+    roundId,
+    chainId,
+  }: {
+    roundId: string;
+    chainId: number;
+  }): Promise<{ round: Round } | null> {
```
boudra, line 411: @gravityblast I think it would make more sense to always return an object but set round to null, this way it can more backwards compatible as the type returned is always an object, what do you think?

boudra (approved):

michellema1208 (approved):

PR #2989, opened 2024-02-22
remove subgraph url

thelostone-mc:
TODO: How to remove subgraph from zod checks for builder


gravityblast:
shall we keep the branch but close the PR until we can actually merge it? just to keep it easy for PR reviewers

PR #2990, opened 2024-02-22
Load program even if the user is not an operator

boudra:
## Description

I added the roles filter in the GraphQL but this makes it difficult to debug or see programs that are not yours through the debug mode, if we load the program but then check the roles in the code we can implement this easier.

## Checklist

This PR:


packages/data-layer/src/queries.ts
```diff
@@ -312,17 +307,12 @@ export const getRoundByIdAndChainId = gql`
   }
 `;
 
-export const getRoundsByProgramIdAndUserAddress = gql`
-  query getRoundsByProgramIdAndMemberAddress(
-    $chainId: Int!
-    $programId: String!
-    $userAddress: String!
-  ) {
+export const getRoundsByProgramId = gql`
```
thelostone-mc, line 310:
```suggestion
export const getRoundsByProgramIdAndChainId = gql`
```
If we are adding a chainId filter, would recommend updating this to getRoundsByProgramIdAndChainId as in v2 we can have rounds across chains by the same program/profile

thelostone-mc: Would recommend doing the same for the above 2 queries : getProgramByIdAndUser, getProgramByUserAndTag to make it easier to read

boudra: I think it's a good point but this PR is not adding a chain ID filter, it's just removing the roles filter, renaming other queries is out of scope of this PR, we can do it in a separate one

boudra: I updated the one I changed though since I am already renaming it, thanks for the suggestion!

vacekj (approved):
:shipit:

thelostone-mc (approved):

Issue #2976, opened 2024-02-20
Manager: Create a Program/Profile on Allo v2 


michellema1208:
**Overview**
Be able to create a profile on Allo v2 (Program) using Manager. 

# Subtasks

- Index and tag Allo v2 profiles as programs: https://github.com/gitcoinco/grants-stack/issues/2976

**Details**
* Let's tag profiles created on Manager such that only profiles created on Manager show up (and only profiles on Builder show up on Builder)

PR #3030, opened 2024-02-23
manager: Create programs for allo v1 and v2

boudra:
Fixes: #2976 
## Description

ended up mostly deleting code 🙌 

- add createProgram allo method
- implement create program for v1 (program factory) v2 (create profile)
- add `type: program` to the metadata so we can distinguish projects from programs
- Remove create program context in favour of simple hook
- remove context unit test in favour of create program page integration test

# How to test

Loom: https://www.loom.com/share/2ea73c01bb09415a97f6e22a083b822b

Note: Profiles created in Builder will not show up in manager anymore.

1. Create a program in Allo V2, it should show up, try with multiple operators
2. Repeat for Allo V1

## Checklist

This PR:


michellema1208 (changes requested):
@boudra I'm able to create programs on v1 and v2, but for v2, it seems like it's only displaying the third program I created (left) and the first program I created (right): 

Screen Shot 2024-02-23 at 1 51 17 PM

There should be 3 programs total - see Etherscan: https://sepolia.etherscan.io/address/0x37e5e6eb36467f98392ebbb35f110fbb8238e6d3. Nothing changes when I refresh the page. 

I also added you as a round operator to all 3 rounds if you want to investigate! 

-------
This only happened once, but the first time I created a program, I wasn't able to open the v2 program - but then I couldn't reproduce this after creating a second program. Not sure if this is an anomaly, or if it only happens the first time you create a program. 

Screen Shot 2024-02-23 at 1 44 12 PM

boudra:
@michellema1208 those issues are indexer related I created two issues to investigate further.

The reason the first program didn't load when you clicked it is because of an intermittent error from cloudfront: https://github.com/gitcoinco/grants-stack-indexer/issues/463

the reason your second program wasn't picked up is a tricky one, I am still investigating it, I suspect it's an intermittent Sepolia RPC issue on the indexer side. I just tried making 3 programs and they were all created for me https://github.com/gitcoinco/grants-stack-indexer/issues/462

I would suggest merging this PR as there's nothing we can do to fix those issues in Grants stack

thelostone-mc (approved):

melissa-neira (approved):
lgtm!

michellema1208 (approved):

PR #3068, opened 2024-03-01
add GTC as payout token for arbitrum

gravityblast:
Fixes: #issue

## Checklist

This PR:

### DIFF
--- a/packages/common/src/payoutTokens.ts
+++ b/packages/common/src/payoutTokens.ts
@@ -339,6 +339,14 @@ const ARBITRUM_TOKENS: PayoutToken[] = [
     logo: TokenNamesAndLogos["USDGLO"],
     redstoneTokenId: RedstoneTokenIds["USDGLO"],
   },
+  {
+    name: "GTC",
+    chainId: ChainId.ARBITRUM,
+    address: "0x7f9a7db853ca816b9a138aee3380ef34c437dee0",
+    decimal: 18,
+    logo: TokenNamesAndLogos["GTC"],
+    redstoneTokenId: RedstoneTokenIds["GTC"],
+  },
 ];
 const AVALANCHE_TOKENS: PayoutToken[] = [
   {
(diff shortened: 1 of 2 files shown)


michellema1208 (changes requested):
@gravityblast they actually need GTC on Arbitrum, not Polygon! 

See my ticket: https://github.com/gitcoinco/grants-stack/issues/3067

michellema1208 (dismissed):
@gravityblast - feel free to fix this Monday AM! Running into a decimal error when deploying a round on Arbitrum with GTC as payout token - see below

Screen Shot 2024-03-01 at 12 21 58 PM

Screen Shot 2024-03-01 at 12 22 33 PM

0xKurt (approved):

0xKurt:
> @gravityblast - feel free to fix this Monday AM! Running into a decimal error when deploying a round on Arbitrum with GTC as payout token - see below
> 
> Screen Shot 2024-03-01 at 12 21 58 PM Screen Shot 2024-03-01 at 12 22 33 PM

same for me

0xKurt (changes requested):
the allo-v1 wrapper uses https://github.com/gitcoinco/grants-stack/blob/main/packages/common/src/payoutTokens.ts as token source which has no GTC token on arbitrum, which causes the error described by @michellema1208

vacekj:
seconding @0xKurt comments

thelostone-mc (approved):

0xKurt:
@gravityblast i was able to create a round using GTC, but it doesn't show up on the frontend, which might be another issue.

also i would recommend to remove the payoutTokens from round manager

thelostone-mc:
> also i would recommend to remove the payoutTokens from round manager

lets do this in another PR

gravityblast:
thank you @0xKurt @thelostone-mc , I'm going to create a ticket to merge the 2 mappings

PR #3093, opened 2024-03-05
fix(Manager): convert p to div to avoid invalid html

gravityblast:
another PR to convert p tags to div to avoid invalid html errors


codenamejason (approved):

thelostone-mc (approved):
LOVE THIS

PR #3095, opened 2024-03-05
add inner error as "cause" in custom AlloError

gravityblast:
Fixes: #issue

## Checklist

This PR:


thelostone-mc (approved):

thelostone-mc (commented):
noob question : what are we doing here ? Is this for the base bug ?

gravityblast:
yes it was to debug it remotely but it didn't add any value so I'm going to close it

PR #3111, opened 2024-03-11
Fund Contract for Allo V1 and V2

### LINKED ISSUE #2901: Fund contract on Allo V2
**Overview**
The goal of this ticket is to be able to fund a round created on Allo v2

**Details**
with #2900 we already moved the logic to fund a V1 round to the Allo wrapper. 
We should add the logic for V2.

In Allo v2 we don't perform a simple transfer to fund the contract, but we call the `Allo.fundPool` function.
If the pool token is a ERC20, before calling `fundPool`, we need another tx to approve the token to the `Allo` contract. 
(we don't want to introduce a modal in this first implementation)

In V2 we need to call the `fundPool` function of the `Allo` contract:

```
    function fundPool(uint256 _poolId, uint256 _amount) external payable nonReentrant {
```

https://github.com/allo-protocol/allo-v2/blob/main/contracts/core/Allo.sol

boudra:
fixes #2901 
fixes https://github.com/gitcoinco/grants-stack/issues/2900

# Description 

This implements funding the round for both V1 and V2.

I have also made some copy changes, because in Allo V2, you do not really fund a "round contract" and this is not something you can do by just transferring tokens to the "round contract". Allo V2 has a single contract for all rounds and you fund them by calling a shared contract method. The changes are:

- Rename of the "Fund contract" section to "Fund round"
- Tweak paragraph for Allo v2 that mentions that you can transfer tokens to the round address

## How to test

For both Allo v1 and v2:

- Open the "Fund Contract" section in Manager
- Matching pool size should be correct
- Should not allow you to fund an amount that you do not have
- Try to fund the contract (test both a round with a native token and a ERC20)
- Funded amount should increase accordingly

I made some rounds called "QF Round ERC20" partially funded that should be useful to test this

## Checklist

This PR:


bhargavaparoksham (approved):
lgtm!

0xKurt (approved):

michellema1208 (approved):
LGTM - also big thank you Mo for testing the flow on your side too!! I left a quick ask on Discord about testing with DAI on Sepolia and other chains -- I saw that you've tested the DAI flow already, but lmk if you want me to test DAI and other native tokens as well!

Issue #3152, opened 2024-03-18
Remove PGN Bridge and passport score decimals from explorer.gitcoin.co


owocki:
### What was the URL where the error occurred?

https://explorer.gitcoin.co/

### What happened?

Page displays a link to the PGN bridge (wshich is useless bc PGN is being deprecated) and my pasport score down to 6 decimals.

### What should have happened instead?

Page displays no PGN bridge and rounds passport score to 0 decimals. 

### Steps to Reproduce

Go to explorer.gitcoin.co with a wallet that has a passport.

### What chain(s) did the error occur on?

N/A

### Relevant log output

```shell
N/A
```

Issue #3140, opened 2024-03-14
[placeholder] General: remove all instances of PGN bridge in nav bars 


michellema1208:
**Overview**
- [ ] Remove all instances of PGN bridge in nav bars (Builder, Manager, Explorer) 

**Details**
Say more

PR #3155, opened 2024-03-19
Remove PGN Bridge and passport score decimals

bhargavaparoksham:
Fixes: #3152 #3140 

## Description

Remove PGN Bridge and passport score decimals.

## Checklist

This PR:


thelostone-mc (approved):

melissa-neira:
@bhargavaparoksham curious as to why we're getting rid of passport score decimals in the first place, and not also getting rid of donation impact decimals too

Issue #3143, opened 2024-03-15
wait for indexer when applying to a round on v1


boudra:
builder is currently not waiting for the indexer to index the application after applying, which could cause the user to not see their application yet

what to do:

- add a indexingStatus event to applyToRound in the Allo interface
- await for indexing AFTER you get the transaction receipt
- emit the indexingStatus event after indexing is done
- update the builder call site to wait for this event and only reload applications once the indexer is up to date

PR #3156, opened 2024-03-19
wait for indexer when applying to a round on v1

bhargavaparoksham:
Fixes: #3143 

- When applying to a round it should show the indexing step in the modal.
- Once the applying to round transaction is successful it should wait for the indexing to complete.

### DIFF
--- a/packages/builder/src/actions/roundApplication.ts
+++ b/packages/builder/src/actions/roundApplication.ts
@@ -196,11 +196,7 @@ const applyToRound =
       })
       .on("transactionStatus", async (res) => {
         if (res.type === "success") {
-          dispatch({
-            type: ROUND_APPLICATION_LOADED,
-            roundAddress: roundId,
-            projectId: projectID,
-          });
+          console.log("Transaction Status Success", res.value);
         } else {
           dispatchAndLogApplicationError(
             dispatch,
@@ -213,7 +209,12 @@ const applyToRound =
       })
       .on("indexingStatus", (res) => {
         if (res.type === "success") {
-          console.log("Indexing Status", res.value);
+          dispatch({
+            type: ROUND_APPLICATION_LOADED,
+            roundAddress: roundId,
+            projectId: projectID,
+          });
+          console.log("Indexing Status Success", res.value);
         } else {
           console.error("Indexing Status Error", res.error);
         }
--- a/packages/builder/src/actions/roundApplication.ts
+++ b/packages/builder/src/actions/roundApplication.ts
@@ -211,6 +211,13 @@ const applyToRound =
           console.log("Transaction Status Error", res.error);
         }
       })
+      .on("indexingStatus", (res) => {
+        if (res.type === "success") {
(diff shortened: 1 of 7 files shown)


packages/builder/src/actions/roundApplication.ts
```diff
@@ -196,11 +196,7 @@ const applyToRound =
       })
       .on("transactionStatus", async (res) => {
         if (res.type === "success") {
-          dispatch({
-            type: ROUND_APPLICATION_LOADED,
-            roundAddress: roundId,
-            projectId: projectID,
-          });
+          console.log("Transaction Status Success", res.value);
```
codenamejason, line 199: should we just check for error here since we do not do anything with "success" status?

packages/builder/src/actions/roundApplication.ts
```diff
@@ -196,11 +196,7 @@ const applyToRound =
       })
       .on("transactionStatus", async (res) => {
```
codenamejason, line 197: should we add this step to the modal and not skip it?

packages/builder/src/actions/roundApplication.ts
```diff
@@ -211,6 +207,18 @@ const applyToRound =
           console.log("Transaction Status Error", res.error);
         }
       })
+      .on("indexingStatus", (res) => {
+        if (res.type === "success") {
+          dispatch({
```
codenamejason, line 212: we should have an indexed dispatch for here ..

packages/builder/src/actions/roundApplication.ts
```diff
@@ -211,6 +207,18 @@ const applyToRound =
...
+            type: ROUND_APPLICATION_LOADED,
+            roundAddress: roundId,
+            projectId: projectID,
+          });
+          console.log("Indexing Status Success", res.value);
```
boudra, line 217: leftover log

packages/common/src/allo/backends/allo-v2.ts
```diff
@@ -555,6 +556,13 @@ export class AlloV2 implements Allo {
         return error(result);
       }
 
+      await this.waitUntilIndexerSynced({
```
boudra, line 559: if you check a few lines above were already doing `await this.waitUntilIndexerSynced`, it can be removed ...

packages/builder/src/actions/roundApplication.ts
boudra: i agree with jason we should update the UI and do this properly

boudra (changes requested):

thelostone-mc (approved):

michellema1208:
@bhargavaparoksham is this PR something you need product testing for? If so, can you add more instructions on how we should test?

bhargavaparoksham:
@michellema1208 Yes. I've updated the description with testing instructions now. Please test it accordingly. Thanks.

boudra (approved):

michellema1208 (approved):

Issue #3148, opened 2024-03-15
Amounts are shown in ETH instead of MATIC in a Polygon round


gravityblast:
### What was the URL where the error occurred?

https://manager.gitcoin.co/#/round/0x546b0bf47c5b5253ff83c8d614662adb7d3ba9b1?debug=true

### What happened?

Amounts in "Rounds Result", "Fund Grantees", and "Reclaim Funds" have ETH instead of MATIC

### What should have happened instead?

show MATIC

### Steps to Reproduce

Go to https://manager.gitcoin.co/#/round/0x546b0bf47c5b5253ff83c8d614662adb7d3ba9b1?debug=true

Click on the 3 tabs described above

### What chain(s) did the error occur on?

Polygon

PR #3164, opened 2024-03-20
fix incorrect payout token issue on polygon

bhargavaparoksham:
Fixes: #3148 

## Description

Fix incorrect payout token issue on polygon

## Checklist

This PR:


thelostone-mc (approved):

codenamejason (approved):
nice one ✅

boudra:
do you have a link to test this?

michellema1208 (approved):

PR #3192, opened 2024-03-22
Fix allo v2 voting incorrect anchor address

boudra:
Fixes: #issue

## Description

depends on this one: https://github.com/gitcoinco/grants-stack-indexer/pull/510

## Checklist

This PR:

### DIFF
--- a/packages/data-layer/src/data-layer.ts
+++ b/packages/data-layer/src/data-layer.ts
@@ -26,6 +26,7 @@ import {
   v2Project,
   RoundForManager,
   Contribution,
+  RoundForExplorer,
 } from "./data.types";
 import {
   ApplicationSummary,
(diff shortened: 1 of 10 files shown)


packages/data-layer/src/data-layer.ts
```diff
@@ -474,29 +475,29 @@ export class DataLayer {
...
+      getRoundForExplorer,
       requestVariables,
     );
 
     if (response.rounds.length === 0) {
       return null;
     }
 
-    const _round = response.rounds[0] ?? [];
+    const round = response.rounds[0] ?? [];
```
gravityblast, line 488: not related to this PR in particular but having `[]` by default is wrong because we are callung `round.applications` after

boudra: ah yes i missed that, fixed

packages/grant-explorer/src/checkoutStore.ts
```diff
@@ -228,6 +228,8 @@ export const useCheckoutStore = create<CheckoutState>()(
             "roundId"
           );
 
+          console.log(groupedDonations);
```
gravityblast, line 231: we can remove this log

gravityblast (approved):

0xKurt (approved):

PR #3203, opened 2024-03-25
update close on overlay click

codenamejason:
Fixes: #issue

## Checklist

This PR:


thelostone-mc (approved):

michellema1208 (commented):
LGTM!

bhargavaparoksham (approved):

PR #3223, opened 2024-03-25
update allo version banner copy

gravityblast:
Updates Allo version banner copy with dates

### DIFF
--- a/packages/grant-explorer/src/features/common/Navbar.tsx
+++ b/packages/grant-explorer/src/features/common/Navbar.tsx
@@ -111,19 +111,33 @@ export default function Navbar(props: NavbarProps) {
         <div className="bg-white/40 backdrop-blur-sm p-4 text-center w-full font-medium flex flex-col items-center justify-center text-black">
           <div>
             <ExclamationCircleIcon className="h-5 w-5 inline-block mr-2" />
-            To check out rounds on that are running on Allo{" "}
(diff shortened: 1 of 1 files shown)


michellema1208:
@gravityblast @melissa-neira wondering what would be the best global standard for dates? I'm thinking 25/03/24? 

My "consider" feedback for the copy: 
"If you created a round before 25/03/24, you can find and donate to those rounds [here](https://explorer-v1.gitcoin.co/)!" 

Just to make make the copy really user-centric / clear what they need to do!

codenamejason (approved):

boudra:
@michellema1208 

for global standard dates i would recommend something like March 25th which should be clear for eu / us alike

wondering if the "If you created a round before" would be confusing because the banner will be shown to donors, and they normally do not create the rounds, someone else does

boudra (approved):

michellema1208:
@boudra that's a great point, agreed that the messaging could be confusing for donors. 

@gravityblast My recommendation is to keep the banner as is for now -- I think it does the best job at communicating to both Grants Managers and Donors. If we're concerned about the 3/25 date, we can link my guide instead: https://www.notion.so/gitcoin/Navigating-the-Transition-to-Allo-v2-A-Guide-for-Grants-Managers-63e2bdddccb94792af83fdffb1530b85?pvs=4 

wdyt @gravityblast @melissa-neira ?

meglister (approved):

PR #3230, opened 2024-03-26
fix spam rounds

boudra:
handle explorer-v1.gitcoin.co domain


bhargavaparoksham (approved):

michellema1208 (approved):
I actually don't think I'm able to test the explorer-v1 link via this PR, so will auto-approve and check back when it's live -- thanks Mo!

Issue #3233, opened 2024-03-26
bug: view contract button in fund round leads to nowhere


boudra:
example here: https://manager.gitcoin.co/#/round/0xc5262fc451a72f68f83563e30126d9c1d5d77659?debug=true 

Round is deployed on Arbitrum. 

go to fund round => view contract

PR #3234, opened 2024-03-26
3233/fix for contract link on arb

codenamejason:
Fixes: #3233 

## Description

Adds arbitrum for the link

## Checklist

This PR:


michellema1208 (approved):
LGTM!

Issue #2919, opened 2024-02-09
Explorer: disable matching estimates for Avalanche round


michellema1208:
**Overview**
Disable matching estimates for Avalanche round, as it would require additional calculations work in the Indexer and put the migration work at-risk. 

**Details**
Say more

codenamejason:
- [ ] Update round manager to not show matching round stats when on Avalanche.
- [ ] Update grant explorer to not show the matching stats for rounds on Avalanche.

PR #3239, opened 2024-03-26
2919/disable matching stats for avax

codenamejason:
Fixes: #2919 

## Description

- [x] Update round manager to not show matching round stats when on Avalanche.
- To test: In round manager, go to an Avalanche round and view your round stats. The matching stats should not show up and no error should occur.
- [x] Update grant explorer to not show the matching stats for rounds on Avalanche.

## Checklist

This PR:

### DIFF
--- a/packages/grant-explorer/src/features/round/ViewCartPage/SummaryContainer.tsx
+++ b/packages/grant-explorer/src/features/round/ViewCartPage/SummaryContainer.tsx
@@ -2,7 +2,6 @@ import { ChainId, getTokenPrice, NATIVE } from "common";
 import { useCartStorage } from "../../../store";
 import { useEffect, useMemo, useState } from "react";
 import { Summary } from "./Summary";
-import ErrorModal from "../../common/ErrorModal";
 import ChainConfirmationModal from "../../common/ConfirmationModal";
 import { ChainConfirmationModalBody } from "./ChainConfirmationModalBody";
 import { ProgressStatus } from "../../api/types";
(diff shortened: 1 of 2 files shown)


thelostone-mc (approved):

michellema1208:
@codenamejason do you have a live Avalanche round with Passport turned on that I could use to test? If not, would you be able to send in some applications to this round: https://manager.gitcoin.co/#/round/7 so we can test? I'm low on AVAX and can't apply rn. If you don't have AVAX handy, I'll look into bridging some funds tomorrow

michellema1208 (changes requested):
hey @codenamejason - thank you for getting started on this! 

Running into two issues: 
1. I'm still seeing matching estimates on Avalanche on the cart page - I shouldn't see any matching, not even $0.00 on the cart page for rounds on Avalanche 

Screen Shot 2024-03-28 at 11 58 21 AM

2. When I try donating to an Arbitrum round alongside the Avalanche round, I'm not able to enter the donation amount for the project in the Arbitrum round. This is only happening in this PR - not happening on prod. 

To reproduce: 
1. Add projects from this Avax round to cart: https://explorer-git-2929-disable-avax-match-stats-grants-stack.vercel.app/#/round/43114/7 
2. Add projects from this Arbitrum round to cart: https://explorer-git-2929-disable-avax-match-stats-grants-stack.vercel.app/#/round/42161/11
3. Try entering in donation amount on the specific project (not via Apply all) 
4. Notice you can't remove the 0 default donation amount, you can only change donation amount via Apply to all

michellema1208:
@codenamejason I'm actually not sure that we need to make changes on Manager side -- would be good to have @gravityblast weigh in as Tech Lead here! I just checked the view on prod for the test Avax round (https://manager.gitcoin.co/#/round/7) and it just displays 0% matching, which makes sense to me. The Avax team knows that results will be calculated off-platform with Umar, and will be uploaded separately. wdyt @gravityblast ?

thelostone-mc (approved):

Issue #3271, opened 2024-04-08
Manager: Add OP token on Optimism for payouts


michellema1208:
**Overview**
Aave Grants would like to use OP Token for payouts on the round they are launching on Optimism. They are targeting a 4/12 launch date. 

**Details**
See this spreadsheet for details: https://docs.google.com/spreadsheets/d/1piuaiqwDNWCtNEtxyueN9E7khkbfE_dxOCwzuYNtv8Y/edit#gid=1187326881

PR #3275, opened 2024-04-09
Add OP token for matching

codenamejason:
Fixes: #3271 

## Checklist

This PR:

Screenshot 2024-04-09 at 8 34 30 AM

Screenshot 2024-04-08 at 3 04 43 PM


0xKurt:
do we also need to add it to `common/src/payoutTokens`?

codenamejason:
> do we also need to add it to `common/src/payoutTokens` and `grant-explorer/src/features/api/utils`?

We are not adding as a donation token in this scope.

0xKurt (approved):

michellema1208 (approved):
LGTM - great work @codenamejason @thelostone-mc !

Issue #3279, opened 2024-04-10
Explorer: Updating tab fonts


melissa-neira:
**Overview**
To keep in line with branding updates, we're updating the font in the tab menu on the Explorer homepage.

It should look like this:
<img width="1206" alt="Screenshot 2024-04-10 at 12 11 55" src="https://github.com/gitcoinco/grants-stack/assets/87520659/c30f8d09-eee4-4aea-8ab7-8582db851445">

**Details**
See [Figma](https://www.figma.com/file/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?type=design&node-id=323-10123&mode=design&t=1zi4kYj0AiNpQ1Cj-4)

PR #3281, opened 2024-04-10
3279/Update tab font

codenamejason:
Fixes: #3279 

## Description
Updates tab font

## Checklist

This PR:


thelostone-mc (approved):

michellema1208 (commented):
thanks @codenamejason @thelostone-mc for the quick turnaround here! Will defer to @melissa-neira for UX approval

melissa-neira (approved):

Issue #3280, opened 2024-04-10
Explorer: Updating Donation History branding


melissa-neira:
**Overview**
The donation history page is getting a minor facelift to align with other branding updates.

This includes:

- [ ]  Font updates
- [ ]  Color updates
- [ ]  General styling updates to border radiuses, buttons, etc.

**Details**
Changes can be seen here - [Figma](https://www.figma.com/file/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?type=design&node-id=324-13724&mode=design&t=1zi4kYj0AiNpQ1Cj-4)

Issue #3272, opened 2024-04-08
Explorer: link out to an external "Gitcoin Donations Wrapped" page on Donation History page


michellema1208:
**Overview**
Ask: each donor’s Donation History page would link out to an external interactive “Donations Wrapped” webpage 

**Figma**: https://www.figma.com/file/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?type=design&node-id=470-9376&mode=design&t=jCRIa04jz7LST7kN-0

**URL**: https://gg-your-impact.streamlit.app/?address=<your address> 

**Context**: this external webpage is just pulling Gitcoin Grants data, not all rounds that have run on Grants Stack. 

**Details**
Say more

PR #3282, opened 2024-04-10
2380/ Donations page overhaul

codenamejason:
Fixes: #3280 
Fixes: #3272 

## Description

- [x] Update contributions/donations page per Figma - worked with @melissa-neira on some updates for a final outcome.

## Checklist

This PR:


michellema1208 (commented):
@codenamejason @0xKurt great work!! This looks right to me, but I will defer to @melissa-neira on final approval given these are primarily UX/branding updates

melissa-neira (approved):
lgtm!

Issue #3231, opened 2024-03-26
bug: some filters in explorer are not working


boudra:
check this page, it comes blank because of an error:
https://explorer-v1.gitcoin.co/#/rounds?orderBy=MATCH_AMOUNT_IN_USD_DESC&status=active%2Ctaking_applications%2Cfinished

I suspect this might be happening because of the skeleton passing undefined as a round type, but just a wild guess

PR #3296, opened 2024-04-15
Fix filtering bug in explorer

bhargavaparoksham:
Fixes: #3231 

## Description

fix: filtering bug in explorer

## Checklist

This PR:


thelostone-mc (approved):

michellema1208 (approved):
LGTM - great work Bhargav!

codenamejason (approved):

PR #3300, opened 2024-04-15
3228/navbar explorer gg20

codenamejason:
Fixes: #issue

## Checklist

This PR:


thelostone-mc (approved):

bhargavaparoksham (approved):

melissa-neira (dismissed):
small change! the explore rounds button shouldn't be bold, it should be regular. everything else looks great :)

michellema1208 (commented):
@codenamejason great work as always! Confirming the nav bar dropdown is working as expected. Will defer to @melissa-neira on final approval re: navbar UX updates!

PR #3326, opened 2024-04-18
add a limit to all graphql queries

boudra:
Fixes: #issue

## Checklist

This PR:

### DIFF
--- a/packages/data-layer/src/queries.ts
+++ b/packages/data-layer/src/queries.ts
@@ -11,6 +11,7 @@ import { gql } from "graphql-request";
 export const getProgramsByUserAndTag = gql`
   query ($userAddress: String!, $chainId: Int!, $tags: [String!]!) {
     projects(
+      first: 100
       filter: {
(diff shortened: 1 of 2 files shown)


bhargavaparoksham (approved):

packages/data-layer/src/queries.ts
```diff
@@ -70,6 +72,7 @@ export const getProgramById = gql`
 export const getProjectById = gql`
   query ($alloVersion: [String!]!, $projectId: String!) {
     projects(
+      first: 1
```
0xKurt, line 75: since we have canonical and linked projects, we need to get all projects of one project id to get all the information we need.

currently we have logic in data-layer to merge the canonical and linked projects. we could move this logic to the indexer, but currently we should increase the limit.

boudra: fixed, this query should be called `getProjectsById`.

0xKurt (approved):

Issue #3297, opened 2024-04-15
[placeholder] GG20 - Explore Projects tab


michellema1208:
**Overview**
Epic: https://github.com/gitcoinco/grants-stack/issues/3235

TODO - @gravityblast to add details from the outcome of the above spike

GG20 rounds: https://grants.gitcoin.co/

**Details**
Say more

codenamejason:
This needs rebased from main

PR #3328, opened 2024-04-18
WIP/3297 Add Explore Projects Tab

codenamejason:
Fixes: #3297 

## Description

Add back the `Explore Projects` tab with search feature.

Screenshot 2024-04-23 at 7 33 58 PM

## Checklist

This PR:


codenamejason:
@0xKurt Everything is rebased with the collections from @gravityblast and the search input is added back. I am not liking the results so far for the search.

thelostone-mc (approved):

0xKurt (approved):

Issue #3318, opened 2024-04-17
Explorer: MRC checkout branding updates


melissa-neira:
**Overview**
GG20 branding updates - slight styling changes to the MRC modals.

These updates include:
- font changes
- colors

**Details**
Designs - [Figma](https://www.figma.com/file/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?type=design&node-id=547-10404&mode=design)

PR #3329, opened 2024-04-18
3318/mrc style updates

codenamejason:
Fixes: #3318 

## Description

To test:

- [x] Add projects from at least 2 different chains and checkout.
- [x] Add projects from only one chain and checkout.

## Checklist

This PR:

### DIFF
--- a/packages/grant-explorer/src/features/common/ConfirmationModal.tsx
+++ b/packages/grant-explorer/src/features/common/ConfirmationModal.tsx
@@ -61,12 +61,12 @@ export default function ConfirmationModal({
               leaveFrom="opacity-100 translate-y-0 sm:scale-100"
               leaveTo="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
             >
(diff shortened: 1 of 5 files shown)


thelostone-mc (approved):

0xKurt (approved):

michellema1208 (commented):
I'm helping test design changes on behalf of Melissa while she's OOO this week! 

@codenamejason @0xKurt - confirming that the flow works for both MRC and single-round checkout -- great work as usual, and the UI looks sooo much better (h/t @melissa-neira !!) 

My only question: is the "+X gas fee" meant to be hardcoded here, or is it supposed to actually display the estimated gas fee per chain? My guess is that the "+X gas fee" in Melissa's designs was meant to be a placeholder for the actual amount, not something we actually show to end users. I'm not sure if @codenamejason / @0xKurt if you synced with Melissa on this and have more context? 

Screen Shot 2024-04-22 at 3 22 08 PM

codenamejason:
we don't calculate the gas fee, the users wallet does. Keep me honest @0xKurt

thelostone-mc:
I don't think we need to add X gas fees. It's pretty evident that you'd have to pay gas fees

michellema1208 (approved):
Approving this on the premise that "+ X gas fee" gets removed first per above comment!

PR #3331, opened 2024-04-18
Update UTC time to local time

MasterHW:
Fixes: #issue
Only UTC time shown in UI (mainly Round Manager) https://github.com/orgs/gitcoinco/projects/13/views/23?filterQuery=time&pane=issue&itemId=29281048
## Description

Adds function to retrieve local timezone shorthand (e.g. "EDT"), and adds it to local-determined time in a way that as closely mirrors the existing UTC approach as possible. Did not remove existing UTC methods, only replaced their references in the UI and testcase.

## Checklist

This PR:


thelostone-mc:
@MasterHW @melissa-neira Currently when you set the time, it was always in UTC
Just wanted to double check that the expected behaviour is :
- When I set the time -> i use UTC (this is what contract uses)
- When i view the time on page -> it is displayed in local time zone 
- When I edit the time -> I use UTC again

image

image

MasterHW:
@michellema1208 @thelostone-mc converted round creator page to also use local time - removed UTC references, and contract *should* be getting the local time. I've tested this on my end but would feel better if one of you verified as well.

meglister:
@MasterHW the edit round page is looking good! It looks like the time zone is omitted in round setup, could we list it so that it matches settings? 
Screenshot 2024-04-26 at 11 03 57 AM

melissa-neira (approved):
lgtm!

PR #3335, opened 2024-04-19
chore: prevent unchecking all chains in checkout

### LINKED ISSUE #3190: unchecking a chain in the checkout popup makes it dissapear
in this pop up:

Screenshot 2024-03-22 at 14 10 46

when you uncheck a chain, it dissapears, so you cannot check the chain again:

Screenshot 2024-03-22 at 14 10 56

I expect the chain stay there and  just be unchecked

thelostone-mc:
fixes #3190

## Description

To Test:

- [ ] Add at least 1 project from 2 different networks to your cart.
- [ ] Continue to checkout and trigger the modal.
- [ ] Unselect one of the networks.

## Checklist

This PR:


codenamejason (approved):
Tested this on live PR link and locally with success.

michellema1208 (approved):
LGTM!

PR #3344, opened 2024-04-23
Add OpenSourceObserver stats to explorer

MasterHW:
Fixes: Feature add - show codebase-derived insights to Explorer project page above Karma GAP section. Data provided by Open Source Observer.

## Description

Created relevant OSO files in a manner similar to the existing Karma GAP implementation. Should fetch information relevant to project age, number of code contributors, and "velocity" (increasing activity or not) and show them in the UI. If no information can be fetched, show nothing.

## Checklist

This PR:


MasterHW:
Related OSO issue https://github.com/opensource-observer/oso/issues/1288

MasterHW:
done getting schooled by linter, ready for review - we will need to add `REACT_APP_OSO_API_KEY` to vercel as mentioned by @codenamejason

packages/grant-explorer/.env.example
```diff
@@ -66,4 +66,6 @@ REACT_APP_EXPLORER_DISABLE_MATCHING_ESTIMATES=false
 REACT_APP_PASSPORT_API_KEY="test-key"
 REACT_APP_PASSPORT_AVALANCHE_API_KEY="test-key"
 REACT_APP_PASSPORT_API_COMMUNITY_ID="0000"
-REACT_APP_PASSPORT_API_COMMUNITY_ID_AVALANCHE="0000"
\ No newline at end of file
+REACT_APP_PASSPORT_API_COMMUNITY_ID_AVALANCHE="0000"
+
+REACT_APP_OSO_API_KEY=
```
thelostone-mc, line 71: Do we want to add a link to where this can be populated from ?

packages/grant-explorer/src/features/round/OSO/ImpactStats.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const diffMonths = Math.round(diffTime / (1000 * 60 * 60 * 24 * 30)); // Convert to months
+
+  if (diffMonths === 0) {
+    return 'This month';
+  } else if (diffMonths === 1) {
+    return 'Last month';
+  } else {
+    return `${diffMonths} months`;
+  }
+}
```
thelostone-mc, line 79: ^ would recommend throwing this into a file within common package called utis.ts so that others can use it

packages/grant-explorer/src/features/round/OSO/ImpactStats.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  if (recentCommits === 0 && olderCommits === 0) {
+    return 'unknown';
+  } else if (recentCommits >= (1.5 * olderCommits)){
+    return 'increasing';
+  } else if (recentCommits <= 0.5 * olderCommits){
+    return 'decreasing';
+  } else {
+    return 'steady';
+  }
```
thelostone-mc, line 93: The else if is redundant 
```
if (recentCommits === 0 && olderCommits === 0)  return "unknown";
 if (recentCommits >= (1.5 * olderCommits) return "increasing";
 ```

File: packages/grant-explorer/src/features/round/OSO/ImpactStats.tsx [hunk 4]
```diff
@@ -0,0 +1,94 @@
...
+  if (diffMonths === 0) {
+    return 'This month';
+  } else if (diffMonths === 1) {
+    return 'Last month';
+  } else {
+    return `${diffMonths} months`;
+  }
+}
+
+function projectVelocity(stats : IOSOStats){
```
thelostone-mc, line 81: nitpick but keep styling consistent
```suggestion
function projectVelocity(stats : IOSOStats) {
```

thelostone-mc (changes requested):
Overall LGTM but dropped in some feedback

package.json
```diff
@@ -42,6 +42,8 @@
   },
   "dependencies": {
     "concurrently": "^8.2.2",
+    "graphql": "^16.8.1",
+    "graphql-request": "^6.1.0",
```
thelostone-mc, line 46: @boudra Is there a better way of doing this instead of adding it to the root package.json ?  One option would be throwing it into the grant-explorer package.json

MasterHW: made change based off suggestion

MasterHW:
Only outstanding issue/note is that lmpact Stats take a while to load... open to suggestions on if anything looks to be obviously slowing it down. unsure how much is on our end, and how much is the query itself.

codenamejason (approved):
Nice job.

packages/grant-explorer/src/features/round/OSO/ImpactStats.tsx
thelostone-mc: same for else

thelostone-mc (approved):

melissa-neira:
@MasterHW do you know of any projects that can show the OSO integration? I've been looking through and I can't find any just yet!

MasterHW:
> @MasterHW do you know of any projects that can show the OSO integration? I've been looking through and I can't find any just yet!

Certainly, any with a working Github Org (highlighted below) should work. As mentioned in the call, the view can take 10-30 seconds to load. I'm not happy with this load time, and have found sometimes moving between `Project details` tab and back can help prompt it. I'm reviewing this today and hope to find the cause of this slowness.
image
https://explorer-git-fork-masterhw-oso-grants-stack.vercel.app/#/round/42161/27/3

MasterHW:
@melissa-neira worth mentioning, I made a decision for `Velocity` to contain in addition to states `increasing` `decreasing` and `steady`, one called `unknown` for when there is insufficient information. Irregular commits (i.e. not one every month) creates gaps in the returned stats that are difficult to account for. open to tweaking the wording.

melissa-neira:
@MasterHW it took a couple of minutes to load for me, maybe it's worth adding in a loading state? so that users can know that something will appear there whenever it loads. otherwise, as a user, I wouldn't even know that something is supposed to be there. wdyt?

melissa-neira (approved):
lgtm!

MasterHW:
Failing because `Error: OpenSourceObserver API key not set.`, which is expected behaviour if no `REACT_APP_OSO_API_KEY`. Where can I set this for the job?

ryscheng:
This is so cool!!
Thank you for putting this together!

I just wanted to give you a heads up that we're trying to get to v1 of our marts models:
That means you should start to see _v1, _v2 style major versioning in the next set of data pipeline runs. We're also improving our style guide for model naming so folks know what to expect in the future. Thanks for being patient with us as we steady these models. We'll aim to use this versioning to avoid major breakages in the future.

But for this iteration, we apologize if this breaks any of your integrations. Please see our GraphQL explorer to see the current status of the model names
[https://cloud.hasura.io/public/graphiql?endpoint=https://opensource-observer.hasura.app/v1/grap](https://cloud.hasura.io/public/graphiql?endpoint=https://opensource-observer.hasura.app/v1/graphql)

PR #3369, opened 2024-04-25
Make projectstats panel collapse on mobile

MasterHW:
Fixes: 
Addresses project stats sidepanel not displaying neatly on mobile/small screens.
image

## Description

Styling fixes that makes the box auto-scale from row to column depending on screen size.
image

## Checklist

This PR:


codenamejason (approved):
Looks good on all my screens 🥳

thelostone-mc (approved):

PR #3370, opened 2024-04-25
[wip] add anchor to collections

0xKurt:
Fixes: #issue

## Checklist

This PR:


packages/data-layer/src/openapi-search-client/apis/DefaultApi.ts
```diff
@@ -103,6 +103,7 @@ export class DefaultApi extends runtime.BaseAPI {
       initOverrides,
     );
 
+    console.log("response", response);
```
thelostone-mc, line 106:
```suggestion
```

thelostone-mc (approved):

codenamejason (approved):

PR #3377, opened 2024-04-26
remove wait for indexer in donation

boudra:


packages/common/src/allo/backends/allo-v2.ts
```diff
@@ -198,10 +198,6 @@ export class AlloV2 implements Allo {
         60_000,
         publicClient
       );
-      await this.waitUntilIndexerSynced({
```
thelostone-mc, line 201: Why though ? Is it going to be added on the cart page as a standalone call with wait and proceed logic ?

boudra: it's just to stop people thinking that their vote hasn't gone through at checkout. i'd rather have people go to donation history page and not have the donation there since we have copy explaining that votes can could take a while to show up, user will be more assured that tx went through as opposed to failing at checkout.

thelostone-mc (approved):

packages/common/src/allo/backends/allo-v2.ts
thelostone-mc: Do we wanna add a banner in the donation history pages stating, recent donations will appear shortly ? Else this just comes back as another bug report. Might as well get it done together. What do you think ?

PR #3392, opened 2024-04-28
Fix program filter

0xKurt:
Context: https://discord.com/channels/562828676480237578/1234146304108531852

If the creator is the EOA, but the operator is the Safe and this will never return true. The newly introduced change filters the programs by Owner address

Fixes: #issue

## Checklist

This PR:

### DIFF
--- a/packages/round-manager/src/features/api/program.ts
+++ b/packages/round-manager/src/features/api/program.ts
@@ -56,10 +56,7 @@ export async function listPrograms(
     // Filter out programs where operatorWallets does not include round.createdByAddress.
     // This is to filter out spam rounds created by bots
     programs = programs.filter((program) => {
-      return (
-        program.createdByAddress &&
-        program.operatorWallets?.includes(program.createdByAddress)
-      );
+      return program.operatorWallets?.includes(address);
     });
 
     return programs;
--- a/packages/round-manager/src/features/api/program.ts
+++ b/packages/round-manager/src/features/api/program.ts
@@ -34,30 +34,29 @@ export async function listPrograms(
       throw Error("Unable to fetch programs");
     }
 
-    let programs: Program[] = [];
+    const programs: Program[] = [];
 
     for (const program of programsRes.programs) {
-      programs.push({
-        id: program.id,
-        metadata: program.metadata,
-        operatorWallets: program.roles.map(
-          (role: { address: string }) => role.address
-        ),
-        tags: program.tags,
-        chain: {
-          id: chainId,
-          name: CHAINS[chainId]?.name,
-          logo: CHAINS[chainId]?.logo,
-        },
-        createdByAddress: program.createdByAddress,
-      });
+      if (
+        program.roles.some(
+          (role) => role.role === "OWNER" && role.address === address
+        )
+      )
+        programs.push({
+          id: program.id,
+          metadata: program.metadata,
+          operatorWallets: program.roles.map(
+            (role: { address: string }) => role.address
+          ),
+          tags: program.tags,
+          chain: {
+            id: chainId,
+            name: CHAINS[chainId]?.name,
+            logo: CHAINS[chainId]?.logo,
+          },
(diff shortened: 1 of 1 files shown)


codenamejason (approved):

codenamejason (commented):
Test listPrograms is failing:

The mockPrograms is missing the role property the updated function is using.

packages/round-manager/src/features/api/program.ts
```diff
@@ -34,34 +33,31 @@ export async function listPrograms(
...
-        chain: {
-          id: chainId,
-          name: CHAINS[chainId]?.name,
-          logo: CHAINS[chainId]?.logo,
-        },
-        createdByAddress: program.createdByAddress,
-      });
+      if (
+        program.roles.some(
+          (role) => role.role === "OWNER" && role.address === address
```
boudra, line 41: just to confirm do we have other roles other than OWNER? like MEMBER? and do they also need access?

boudra: i am assuming a bot cannot fake the creation of a profile with someone else being an owner?

0xKurt: > just to confirm do we have other roles other than OWNER? like MEMBER? and do they also need access?

we also have members, but this the original filter was filtering by `createdByAddress` I assume that we only show the Program to the actual Owner/Operator

@meglister I think that could be a product question: Should the Program also be shown to Program Member?
Problem: A Member of a Program would see all the rounds, but since Program Member **are not always** round manager, the members would also see rounds, which they may can't access.

My Suggestion: keep the owner for now and start a discussion about the additional flows we would need

thelostone-mc: > My Suggestion: keep the owner for now and start a discussion about the additional flows we would need

This makes sense. Let's wait for meg to sign off and then move ahead

gravityblast: the filter we have was for a problem we had in Scroll. bots where creating projects/programs adding other people just to spam. 
so I would end up seeing fake programs where I'm manager, but they are actually spam programs. 
It happened only in Scroll, I wonder if it's still happening, we can try and maybe remove the filter if it's no longer the case. 

if it's no longer needed, we need to go back to load all the programs where I have any role, both owner and member, and we don't need to filter anything anymore in the front-end.

Now it filters out projects if the sender is not part of the managers, because bots were creating projects replying data from other tx, without adding themselves to the members.

gravityblast: > i am assuming a bot cannot fake the creation of a profile with someone else being an owner?

@boudra  they can, in fact that's the temporary filter we use to understand if they were the bots we had in scroll

boudra: @gravityblast yeah that's what i thought, thanks for the confirmation, the check if the user is part of the roles is already done in the graphql in this case. the check here is for the spam in scroll.

we should check scroll, i remember the bots were just repeating any transaction we made, i wonder if there is a way to the check but make it work for multisig?

boudra: note that if we do decide to refactor how we do the check here, please let's keep the following comment (or a variation of it), because it's useful context.

```
    // Filter out programs where operatorWallets does not include round.createdByAddress.
    // This is to filter out spam rounds created by bots
```

boudra: if we do find out that it's impossible to do the bot spam check with multisigs, maybe we can solve it at the product level and introduce a hide / archive feature for programs & rounds. so users can get rid of spam rounds.

gravityblast: we can try!
the spam problem is more for Manager than Explorer though. with those bots what happened was:

* I create a round called X
* bots create 5 rounds called X adding me as operator
* when I open Manager I see 6 rounds called X

PR #3420, opened 2024-05-08
map gg20 rounds as passport-mbds rounds

bhargavaparoksham:
Fixes: #issue

## Description

Map gg20 rounds as passport-mbds rounds

## Checklist

This PR:

### DIFF
--- a/packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
+++ b/packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
@@ -492,16 +492,23 @@ function ViewRoundResults({
                                 <span className="text-sm mt-1 ml-2 text-gray-400">
                                   Since you selected frictionless auto-sybil
                                   detection at setup, select this option. Please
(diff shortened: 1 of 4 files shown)


boudra (approved):

packages/round-manager/src/features/round/ViewRoundResults/ViewRoundResults.tsx
```diff
@@ -539,13 +541,7 @@ function ViewRoundResults({
                       </RadioGroup.Option>
                       <RadioGroup.Option value={false}>
                         {({ checked }) => (
-                          <div
-                            className={classNames(
-                              "cursor-pointer flex my-2",
-                              disableRoundSaturationControls &&
```
boudra, line 545: is this intended?

bhargavaparoksham: It was added by mistake, so removed it. 
disableRoundSaturationControls is not relevant to this radioButton group.

Issue #2733, opened 2023-12-05
Manager: viewing team members as an operator


michellema1208:
**User Story:**
As a
I want to
So that

**Acceptance Criteria**

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

Issue #2727, opened 2023-12-05
Manager: add a "Manage Team" section to Round Details page


michellema1208:
**User Story:**
As a Round Manager 
I want to be able to manage my team 
So that I can add/remove operators easily 

**Acceptance Criteria**
GIVEN that I am on the Round Details page 
WHEN I look at the left sidebar 
THEN I see a new tab called "Manage Team" 

GIVEN that I want to manage my team 
WHEN I click on "Manage Team" 
THEN I see a new page where I can add/remove operators on my round
AND relevant copy explaining "Manage Team" at the top of the page (see Figma) 

**Product & Design Links:**
* [Figma](https://www.figma.com/file/UIXKYToUBeyb47mVLk7atW/Manager-S19?type=design&node-id=286-4863&mode=design&t=pQjPoIStVVI7nJdC-0)

**Tech Details:**
* Related spike ticket on data needs: https://github.com/gitcoinco/grants-stack/issues/2719

**Open Questions:**

**Notes/Assumptions:**
- This includes writing test coverage for the functionality added in the ticket, if necessary

PR #3427, opened 2024-05-09
add a Manage Team section to Round Details page

bhargavaparoksham:
Fixes: #2727 
Fixes: #2733 

## Description

1. Add a "Manage Team" section to Round Details page on Manager
2. View team members as an operator

## Checklist

This PR:


melissa-neira (commented):
@bhargavaparoksham question - it's a little weird to see myself listed twice here 
Screenshot 2024-05-14 at 18 13 12
shouldn't I be listed just once here? as admins are inherently operators as well?

thelostone-mc (approved):

melissa-neira (approved):
lgtm!

PR #3436, opened 2024-05-14
DG: Add payout token list

0xKurt:
Changes on Direct Grants:

This PR allows to remove the payout token from the Direct Grants Round creation form,
also it resolves the bug of setting a wrong/not existing token.

As a replacement for the pre-defined token a token list dropdown was added to the payout form.
The dropdown allows to chose a token form our payout token list and to payout with a custom token.

To Test:
- creating a direct grants round (without payout token)
- payout a grantee with a token from the dropdown
- payout a grantee with a custom token (i can send you some test token)

Screenshots:
image
Note: RND is a Random token i used for testing which is not in the payout token list

image
Custom token allows to paste in an address and loads the needed token information like symbol and decimals

image
a wrong token address shows an error


thelostone-mc (approved):

melissa-neira (approved):
lgtm!

PR #3438, opened 2024-05-14
update lukso's logo + styling

MasterHW:
Fixes: LUKSO logo and all-caps styling.

## Description

LUKSO provided feedback that they preferred an alternate logoset, and all-caps styling.

## Checklist

This PR:


MasterHW:
New issue: for unknown reason, LUKSO logo displays with black backing. LUKSO Testnet curiously does not.
image
Also, specifically Explorer still shows the default/Ethereum logo. This issue exists in current main repo as well.

thelostone-mc (approved):

PR #3440, opened 2024-05-15
fix: payout issue in v2

bhargavaparoksham:
Fixes: #issue

## Description

fix: payout issue in v2

## Checklist

This PR:

### DIFF
--- a/packages/common/src/allo/backends/allo-v2.ts
+++ b/packages/common/src/allo/backends/allo-v2.ts
@@ -1217,9 +1217,6 @@ export class AlloV2 implements Allo {
           amount: BigInt(distribution[3].toString()),
           merkleProof: validMerkleProof as Address[],
         });
-
-        console.log("projectsWithMerkleProof", projectsWithMerkleProof);
-
       });
 
       const strategy = new DonationVotingMerkleDistributionStrategy({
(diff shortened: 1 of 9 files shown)


packages/common/src/allo/backends/allo-v2.ts
```diff
@@ -1340,3 +1352,40 @@ export type ProjectWithMerkleProof = {
...
 };
+
+/**
+ * Generate merkle tree
+ *
+ * To get merkle Proof: tree.getProof(distributions[0]);
+ * @param matchingResults MatchingStatsData[]
+ * @returns
+ */
+export const generateMerkleTreeV2 = (
```
0xKurt, line 1363: do we still need the generateMerkleTree function in utils, if we have one in allo-v1.ts and another one in allo-v2.ts?

0xKurt (changes requested):

packages/common/src/allo/backends/allo-v2.ts
bhargavaparoksham: @0xKurt  You are right. We don't need the one in utils. I've removed it now.

0xKurt (approved):

thelostone-mc (approved):

meglister (approved):
tested at https://manager-git-fix-v2-payout-grants-stack.vercel.app/#/round/11 and verified 1 grantee was showing as paid. i successfully paid out the 2nd grantee and both are showing as paid in the UI

PR #3453, opened 2024-05-20
purge deprecated code

thelostone-mc:
image

- fix warning across dapps
- remove subgraph 
- remove deprecrated code
- cleanup .env file'

**Builder** : All flows tested
http://localhost:3001/#/chains/11155111/registry/0x/projects/0xe96684a968584f5f53c4df9363b571ae17f04d202b2b07fcd975cc97d2bb65eb

**Rounds**: All flows tested except payout
http://localhost:3000/#/round/161

**Explorer**: All flows tested
http://localhost:3002/#/round/11155111/161


packages/grant-explorer/src/test-utils.tsx
```diff
@@ -100,8 +100,8 @@ const makeTimestamp = (days?: number) =>
   Math.floor(Number(faker.date.soon(days)) / 1000).toString();
 
 export const makeRoundMetadata = (
-  overrides?: Partial<__deprecated_RoundMetadata>
-): __deprecated_RoundMetadata => ({
+  overrides?: Partial<any>
+): any => ({
```
0xKurt, line 104: can we use type `Round` here?

gravityblast: let's not use `any`, if it's deprecated but still use let's keep it or create a new type that we will continue to use

thelostone-mc: @gravityblast / @0xKurt lol okay i was using it only for test but this makes sense. Fixed

0xKurt (approved):

PR #3454, opened 2024-05-20
New chain config/common

codenamejason:
Fixes: #issue

## Checklist

This PR:


packages/grant-explorer/src/checkoutStore.ts
```diff
@@ -117,8 +121,14 @@ export const useCheckoutStore = create<CheckoutState>()(
         [chain: number]: CartProject[];
       };
 
-      const getVotingTokenForChain =
-        useCartStorage.getState().getVotingTokenForChain;
+      // : TToken
+      const getVotingTokenForChain = async (chainId: number) =>
+        (await getTokensByChainId(chainId))[0] as TToken;
+
+      // fixme: this is a hack to get the token for mainnet/ update to use actual chainId
```
thelostone-mc, line 128: todo

packages/round-manager/src/features/round/FundContract.tsx
```diff
@@ -14,19 +14,18 @@ import ConfirmationModal from "../common/ConfirmationModal";
...
-} from "common";
+import { classNames, useAllo, useTokenPrice } from "common";
 import { assertAddress } from "common/src/address";
 import { formatUnits } from "viem";
+import {
+  getTokensByChainId,
+  TPayoutToken,
+} from "@grants-labs/gitcoin-chain-data";
 
+// fixme: move this to its own hook
```
thelostone-mc, line 25: ?

Issue #3450, opened 2024-05-17
Unify .env across stack packages


MasterHW:
**Overview**
Create and use a single unified `.env` contained in the top folder, that will be shared across all the grants-stack packages. This will avoid the repetitious configuring of `.env` files across multiple directories, which each have some unique environment variables.
Also an opportunity to clean up the `.env` file more, employing as many ready-to-use default values as possible. The goal is to minimize friction to a new user wanting to get grants-stack up and running.

**Details**
Preliminary work has been done on new branch `streamline-env` to compile [a single `.env.example`](https://github.com/gitcoinco/grants-stack/blob/streamline-env/.env.example) containing all the envs used across all packages.
Note: as best as I could tell, `REACT_APP_CHAINS` is mentioned in builder's `.env.example` but not used in any code. I have removed it from the unified `.env.example`

> ...configure it in a way that it picks up the root .env and then if there's a .env at the app level, it will [quietly] override the root level one [for just this package]

more context and discussion -> https://discord.com/channels/562828676480237578/1237795120670183555

MasterHW:
closed with https://github.com/gitcoinco/grants-stack/pull/3467

PR #3467, opened 2024-05-27
Single .env file

gravityblast:
Fixes: #3450 

## Checklist

This PR:


.env.example
MasterHW, line 1: would like to see this merged with https://github.com/gitcoinco/grants-stack/blob/streamline-env/.env.example

MasterHW (changes requested):
i like the new IPFS/Pinata distinction. approved w/ one request

MasterHW (approved):
all good!

thelostone-mc (approved):
Do we need to make changes on env in production ?

PR #3471, opened 2024-05-29
Add project page in explorer

bhargavaparoksham:
Fixes: #issue

## Description

Add project page in explorer

test project page: https://explorer-git-explorer-project-page-grants-stack.vercel.app/#/projects/0x00174304df2cf69c66f29ffdabf3a005e42f8de531b3c2fe922182d8240512da

## Checklist

This PR:


melissa-neira (approved):
lgtm!

thelostone-mc (changes requested):
This feels like a good opportunity to reuse the code. Can we extract a few components like the summary info into a component and keep it common. 

Also @melissa-neira it would help us a great deal if the designs for new things we build are similar. That way we can start unifying the code as we take up tasks

thelostone-mc (approved):

melissa-neira:
@thelostone-mc agreed! there have been new designs for builder for a long time, they've just never made it on the roadmap. hopefully soon!

packages/grant-explorer/src/features/projects/hooks/useProject.ts
```diff
@@ -0,0 +1,18 @@
+import useSWR from "swr";
+import { DataLayer } from "data-layer";
+import { AlloVersion } from "data-layer/src/data-layer.types";
+
+type Params = {
+  projectId?: string;
+};
+
+export function useProject(params: Params, dataLayer: DataLayer) {
+  const shouldFetch = Object.values(params).every(Boolean);
```
gravityblast, line 10: if projectId is undefined it should return earlier null or undefined, the same returned from the data layer if the project is not found, so that we don't need to cast it later with `as string`

packages/grant-explorer/src/features/projects/hooks/useProject.ts
```diff
@@ -0,0 +1,18 @@
...
+  return useSWR(shouldFetch ? ["applications", params] : null, async () => {
+    const validatedParams = {
+      projectId: params.projectId as string,
+      alloVersion: "allo-v2" as AlloVersion,
```
gravityblast, line 14: let's use `as const` instead of `as AlloVersion`.

with `as const` it ensures it's a valid value. using `allo-v3` will fail at compile time. 
with `as AlloVersion` any string will work.

packages/grant-explorer/src/features/projects/hooks/useProject.ts
```diff
@@ -0,0 +1,18 @@
...
+
+type Params = {
+  projectId?: string;
+};
+
+export function useProject(params: Params, dataLayer: DataLayer) {
+  const shouldFetch = Object.values(params).every(Boolean);
+  return useSWR(shouldFetch ? ["applications", params] : null, async () => {
+    const validatedParams = {
+      projectId: params.projectId as string,
```
gravityblast, line 13: maybe here we can use `projectId!` since we already checked that it's not undefined

packages/grant-explorer/src/features/projects/hooks/useProject.ts
gravityblast, line 10: instead of `Boolean` we can explicitly check that the value is not `undefined` with `every((v) => v !== undefined)`

packages/grant-explorer/src/features/projects/ViewProject.tsx
```diff
@@ -0,0 +1,296 @@
...
+                <ProjectLogo {...project?.metadata} />
+              </div>
+            </div>
+          </div>
+        </div>
+        <div className="md:flex gap-4 flex-row-reverse">
+          <div className="flex-1">
+            {error === undefined ? (
+              <>
+                <Skeleton isLoaded={Boolean(title)}>
```
gravityblast, line 129: `title !== undefined`

packages/grant-explorer/src/features/projects/ViewProject.tsx
```diff
@@ -0,0 +1,296 @@
...
+      },
+    ],
+    [project, description]
+  );
+
+  const handleTabChange = (tabIndex: number) => {
+    setSelectedTab(tabIndex);
+  };
+
+  return (
```
gravityblast, line 103: if it's still loading it makes sense to show the skeletons, but if it finished loading and the project is undefined, we should show a 404 page

gravityblast (dismissed):
I added a few comments, and most importantly we need a 404 page if the project is not found

packages/grant-explorer/src/features/projects/ViewProject.tsx
gravityblast: actually maybe it's better to check that `data !== undefined && error === undefined`

packages/grant-explorer/src/features/projects/hooks/useProject.ts
bhargavaparoksham: @gravityblast `type AlloVersion = "allo-v1" | "allo-v2";`
Why would with `as AlloVersion` any string will work?

gravityblast: @bhargavaparoksham  cause it converted to that type. try putting `"allo-v3" as AlloVersion` and it still works.
Instead `"allo-v3" as const` would fail at compile time

bhargavaparoksham: Aah got it.

packages/grant-explorer/src/features/projects/ViewProject.tsx
```diff
@@ -0,0 +1,303 @@
...
+    >
+      <ProjectLink url={website} icon={GlobeIcon}>
+        {website}
+      </ProjectLink>
+      <ProjectLink icon={CalendarIcon}>{createdOn}</ProjectLink>
+      {projectTwitter !== undefined && (
+        <ProjectLink
+          url={`https://twitter.com/${projectTwitter}`}
+          icon={TwitterIcon}
+          isVerified={false}
```
gravityblast, line 218: @bhargavaparoksham why is `isVerified` set to `false` by default?

packages/grant-explorer/src/features/projects/ViewProject.tsx
```diff
@@ -0,0 +1,303 @@
...
+  url?: string;
+  isVerified?: boolean;
+}>) {
+  const Component = url ? "a" : "div";
+  return children ? (
+    <div className="flex items-center gap-2">
+      <div>{createElement(icon, { className: "w-4 h-4 text-grey-400" })}</div>
+      <div className="flex gap-2">
+        <Component
+          href={url}
```
gravityblast, line 258: @bhargavaparoksham  `href` is not a valid attribute for a `div` element. it's better create the 2 possible elements independently

0xKurt (approved):

PR #3481, opened 2024-06-04
remove faulty/unused package

thelostone-mc:
Fixes: #issue

## Checklist

This PR:


gravityblast (approved):

gravityblast (commented):
it might need the lock file changes as well

0xKurt (approved):

PR #3494, opened 2024-06-07
Add stats sidebar & past rounds tab

bhargavaparoksham:
Fixes: #issue

Test project link: https://explorer-git-explorer-project-page-2-grants-stack.vercel.app/#/projects/0x7c2db89015825d83f56cd55aff50b2ac0a8b50338d88c2988323ce6fa7739d1a

Add stats sidebar & past rounds tab

## Checklist

This PR:

### DIFF
--- a/packages/data-layer/src/data-layer.ts
+++ b/packages/data-layer/src/data-layer.ts
@@ -25,6 +25,7 @@ import {
   RoundForExplorer,
   ExpandedApplicationRef,
   RoundApplicationPayout,
+  ProjectApplicationWithRoundAndProgram,
 } from "./data.types";
 import {
   ApplicationSummary,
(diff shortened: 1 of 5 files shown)


melissa-neira (changes requested):
@bhargavaparoksham I'm seeing some weird behavior with the responsiveness of the past rounds tab. things are looking a little off, especially the round type badge. 

this is what I see on my screen:

Screenshot 2024-06-10 at 15 27 41
- text is not centered at normal width

Screenshot 2024-06-10 at 15 27 49
- the badge is never supposed to look like this, maybe we can just shorten it to QF on mobile screens?

bhargavaparoksham:
@melissa-neira will fix these issues in the next pr.

Issue #3483, opened 2024-06-04
Install Posthog on Builder & Manager


meglister:
**Overview**
We're getting a lot of value out of using posthog in Explorer -- let's do the same for Builder & Manager

**Details**
Use the web snippet below to install posthog

```<script>
    !function(t,e){var o,n,p,r;e.__SV||(window.posthog=e,e._i=[],e.init=function(i,s,a){function g(t,e){var o=e.split(".");2==o.length&&(t=t[o[0]],e=o[1]),t[e]=function(){t.push([e].concat(Array.prototype.slice.call(arguments,0)))}}(p=t.createElement("script")).type="text/javascript",p.async=!0,p.src=s.api_host.replace(".i.posthog.com","-assets.i.posthog.com")+"/static/array.js",(r=t.getElementsByTagName("script")[0]).parentNode.insertBefore(p,r);var u=e;for(void 0!==a?u=e[a]=[]:a="posthog",u.people=u.people||[],u.toString=function(t){var e="posthog";return"posthog"!==a&&(e+="."+a),t||(e+=" (stub)"),e},u.people.toString=function(){return u.toString(1)+".people (stub)"},o="capture identify alias people.set people.set_once set_config register register_once unregister opt_out_capturing has_opted_out_capturing opt_in_capturing reset isFeatureEnabled onFeatureFlags getFeatureFlag getFeatureFlagPayload reloadFeatureFlags group updateEarlyAccessFeatureEnrollment getEarlyAccessFeatures getActiveMatchingSurveys getSurveys onSessionId".split(" "),n=0;n<o.length;n++)g(u,o[n]);e._i.push([i,s,a])},e.__SV=1)}(document,window.posthog||[]);
    posthog.init('phc_PG9S5teAT5ptfLau1uvpFpU4JFVTWVDmaK8bHu8ei3s',{api_host:'https://us.i.posthog.com', 
        })
</script>```

meglister:
@gravityblast good candidate for bug week

PR #3495, opened 2024-06-07
Add posthog to builder/manager

codenamejason:
Fixes: #3483 

## Description

Adds PostHog configuration to capture all events for Manager and Builder.

## Checklist

This PR:


codenamejason:
The build will fail until we add the .envs to Vercel. I will post here when completed.

0xKurt (approved):

thelostone-mc (approved):

PR #3496, opened 2024-06-07
feat: Explorer - change pre-round page

hussedev:
## Description

This PR is for the changes made based on this design [in Figma](https://www.figma.com/design/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?node-id=898-1985&t=6grNzOclZqLdEC12-0)
We remove the old pre-round page, and reuse the after round start page:

- src/features/api/utils.ts:
  -  implemented getTimeLeft function reusing the logic of getDaysLeft, to return an object { days, hours, minutes, seconds } with the time left values based on a timestamp passed as argument
  - getDaysLeft now uses getTimeLeft
  - implemented parseTimeLeftString that receives as argument the same structure that getTimeLeft returns, and converts is to string with the format `X days, X hours, X minures`
- src/features/common/styles.tsx & tailwind.config.js:
  - extended TW theme with a rainbow gradiend background to use it in a badge 
- src/features/round/ViewProjectDetails.tsx:
  - cart button now is hidden in pre-round period
- src/features/round/ApplicationsCountdownBanner.tsx & RoundStartCountdownBadge.tsx:
  - implemented countdown banner and badge for pre-round period page
- src/features/round/ViewRoundPage.tsx:
  - removed BeforeRoundStart component
  - refactored AfterRoundStart (now named RoundPage) to have a unified page for both before and after round has started, the changes are adding a badge and banner with countdowns for applications and round start period, and removing the carts in preround period.

## Checklist

This PR:


meglister (approved):

codenamejason (changes requested):
I think we need to remove the Allov2 logo

codenamejason (changes requested):
The footer is missing

hussedev:
> The footer is missing
> 
> Screenshot 2024-06-07 at 3 37 43 PM
> see Figma
> 
> Screenshot 2024-06-07 at 3 38 11 PM

Good catch!!! But not sure if we need to make that change right now.
We are going to leave it like this for now, not sure if we will have that footer and remove the chat, but we can make this change next week, now we wanted to have this deployed like this.

Thanks!

hussedev:
> I think we need to remove the Allov2 logo
> 
> Screenshot 2024-06-07 at 3 34 12 PM
> Looks like the Allov2 logo should be removed based on [Figma](https://www.figma.com/design/c4UCPzbLnK3kz67M6R9VGN/Explorer-S20?node-id=898-1985&t=6grNzOclZqLdEC12-0)
> 
> Screenshot 2024-06-07 at 3 35 26 PM

True ;), is not in the figma design, also the strategy badge "Quadratic funding" but not sure if those changes are required now. For now we can deploy it as it is now, I only applied some changes on the "during & post round page" to create the "pre round page" both now share same design, next week will be easy to remove the Allo logo and Strategy bagde in case that is required.

Thanks!

thelostone-mc (approved):

PR #3498, opened 2024-06-10
use alchemy and infura provider

0xKurt:
Fixes: #issue

## Checklist

This PR:

### DIFF
--- a/packages/common/src/chains.ts
+++ b/packages/common/src/chains.ts
@@ -59,7 +59,6 @@ const parseRainbowChain = (chain: TChain) => {
   const mappedChain: Chain = {
     id: chain.id,
     name: chain.prettyName,
-    // network: chain.name,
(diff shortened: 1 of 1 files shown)


packages/common/src/chains.ts
```diff
@@ -15,11 +53,13 @@ const parseRainbowChain = (chain: TChain) => {
...
 
+  const rpc = getRpcUrl(chain);
+
   // Map the TChain to @rainbow-me/rainbowkit/Chain
-  const mappedChain = {
+  const mappedChain: Chain = {
     id: chain.id,
     name: chain.prettyName,
-    network: chain.name,
+    // network: chain.name,
```
thelostone-mc, line 62:
```suggestion
```

thelostone-mc (approved):

thelostone-mc (approved):

0xKurt:
LGTM!

PR #3502, opened 2024-06-10
Update rainbow dependencies

thelostone-mc:
Fixes: #issue

## Checklist

This PR:


thelostone-mc:
Closed in favour of https://github.com/gitcoinco/grants-stack/pull/3514

PR #3509, opened 2024-06-11
Add active rounds in sidetab

bhargavaparoksham:
Fixes: #issue

Add active rounds in sidetab 

## Checklist

This PR:


MasterHW (approved):
looks good! example here https://explorer-git-explorer-project-page-3-grants-stack.vercel.app/#/projects/0x112bf38af267ba83a39fd13b0c00add5b1e886f2ccc8a3f868b4c14702f71c3b

thelostone-mc (approved):

PR #3512, opened 2024-06-12
fix ui

thelostone-mc:
Fixes: #issue

## Checklist

This PR:


packages/common/src/index.ts
```diff
@@ -475,7 +475,7 @@ export * from "./allo/transaction-builder";
  * @returns the transaction block explorer URL for the provided transaction hash and network
  */
 export const getTxBlockExplorerLink = (chainId: number, txHash: string) => {
-  return getChainById(chainId) + "tx/" + txHash;
+  return getChainById(chainId).blockExplorer + "tx/" + txHash;
```
thelostone-mc, line 478:
```suggestion
  return getChainById(chainId)?.blockExplorer + "tx/" + txHash;
```

bhargavaparoksham (approved):

0xKurt (approved):

PR #3534, opened 2024-06-26
keep the finalize round button always enabled

gravityblast:
Keep the finalize round button always enabled instead of checking if the pool is completely funded. 
This is needed in case the round operator wants to distribute less funds only to some grantees.


hussedev (approved):
Nice catch

thelostone-mc (approved):

PR #3537, opened 2024-06-26
fix: isInfiniteDate should check if year is lower or equal to 1970

hussedev:
Fixes: #issue https://github.com/gitcoinco/grants-stack/issues/3536

## Description

Our data layer is not checking if round dates are null when fetched from the indexer:
image

getRoundForExplorer() -> packages/data-layer/src/data-layer.ts
image

Converting null to Date becomes -> `1970-01-01T00:00:00.000Z` (timestamp 0 / Computer Dinosaur Age)

That's the main issue, depending on the timezone, it can be greater or lower than that.

One solution would be to convert all dates to UTC and make calculations from there.
Another solution would be that the data-layer leave null dates as null or undefined instead of converting it.

We currently have this function to address this issue:

```
export const isInfiniteDate = (roundTime: Date) => {
  return (
    roundTime.toString() === "Invalid Date" || roundTime.getFullYear() === 1970
  );
};
```

But it wasn't taking in consideration that the year can become 1969 depending on the Timezone.

Changed the check from === 1970 to <= 1970

```
const isInfiniteDate = (roundTime: Date) => {
  return (
    roundTime.toString() === "Invalid Date" || roundTime.getFullYear() <= 1970
  );
};
```

An alternative solution would be:

```
const isInfiniteDate = (roundTime: Date) => {
  return (
    roundTime.toString() === "Invalid Date" || roundTime.toISOString() === "1970-01-01T00:00:00.000Z"
  );
};
```

## Checklist

This PR:

### DIFF
--- a/packages/grant-explorer/src/features/api/utils.ts
+++ b/packages/grant-explorer/src/features/api/utils.ts
@@ -132,7 +132,7 @@ export const isDirectRound = (round: Round) =>
 
 export const isInfiniteDate = (roundTime: Date) => {
   return (
-    roundTime.toString() === "Invalid Date" || roundTime.getFullYear() === 1970
+    roundTime.toString() === "Invalid Date" || roundTime.getFullYear() <= 1970
   );
 };
 


MasterHW (approved):
surgical fix. good stuff.

codenamejason (approved):

PR #3554, opened 2024-07-11
Add Public Profile Button

0xKurt:
This PR adds a `View Public Profile` button in builder to quickly find the profiles page on explorer

image


melissa-neira (commented):
@0xKurt is it possible to both copy the link to the clipboard and navigate to the public profile? as a user, if I see the link icon, I expect that the link is copied to my clipboard. 

if it's not possible to do both actions, then I would get rid of the icon and change the copy to "view public profile"

0xKurt:
> @0xKurt is it possible to both copy the link to the clipboard and navigate to the public profile? as a user, if I see the link icon, I expect that the link is copied to my clipboard.
> 
> if it's not possible to do both actions, then I would get rid of the icon and change the copy to "view public profile"

changed to `View Public Profile`

bhargavaparoksham (approved):

Issue #3561, opened 2024-07-24
Builder - Application & Round dates appearing the same


MasterHW:
### What was the URL where the error occurred?

https://builder.gitcoin.co/#/chains/10/rounds/57

### What happened?

Application Period and Round Dates are showing the same
![image](https://github.com/user-attachments/assets/708d947d-8044-4262-90c1-8936ed48134e)

### What should have happened instead?

The dates should be different.

### Steps to Reproduce

visit https://builder.gitcoin.co/#/chains/10/rounds/57
impersonating 0x500767ea632a99b7c43305b0e621cfa6d77a9e5a

### What browser(s) are you seeing the problem on?

Chrome

### What chain(s) did the error occur on?

OP

PR #3562, opened 2024-07-24
fix builder round dates

MasterHW:
Fixes: #3561 

## Description

logic seemed to be reversed on setting round times based on the round being direct grant or not. flipped it, and now non-direct grants rounds are showing correct round times.

## Checklist

This PR:


hussedev (approved):
LGTM

thelostone-mc (approved):

Issue #3291, opened 2024-04-14
Builder: Can't select full area for project logo


Difint:
### What was the URL where the error occurred?

https://builder.gitcoin.co/

### What happened?

![image](https://github.com/gitcoinco/grants-stack/assets/101790821/e95e3824-f8ae-4b71-87d4-d25014ed3570)

On the screenshot is the maximum area that could be selected, the logo itself is fit into 512*512 square

### What should have happened instead?

Full area should be selected

### Steps to Reproduce

I'm using Chromium
1. Start creating new project
2. Upload logo
3. Try to select full area

### What browser(s) are you seeing the problem on?

Chrome

### What chain(s) did the error occur on?

Arbitrum

PR #3584, opened 2024-08-06
[FIX] Bug 3291 - Can now select full area for project logo

dalime:
Fixes: #3291 

## Description

Fixes the bug where only a portion of an image can be selected for a project's logo. See image below.

Screenshot 2024-08-06 at 21 47 42

## Checklist

This PR:

### DIFF
--- a/packages/builder/src/components/base/images/ImageCrop.tsx
+++ b/packages/builder/src/components/base/images/ImageCrop.tsx
@@ -91,8 +91,6 @@ export default function ImageCrop({
               setCompletedCrop(c);
             }}
             aspect={dimensions.width / dimensions.height}
-            maxHeight={undefined}
-            maxWidth={undefined}
           >
             <img
               ref={imgRef}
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/base/images/ImageCrop.tsx
```diff
@@ -91,8 +91,8 @@ export default function ImageCrop({
               setCompletedCrop(c);
             }}
             aspect={dimensions.width / dimensions.height}
-            maxHeight={dimensions.height}
-            maxWidth={dimensions.width}
+            maxHeight={undefined}
+            maxWidth={undefined}
```
0xKurt, line 95: would you mind removing `maxHeight` and `maxWidth` instead of assigning undefined?
Also a quick rebase with main would be nice. Will approve it afterwards.

Thanks for you contrbution!

0xKurt (changes requested):

packages/builder/src/components/base/images/ImageCrop.tsx
dalime: I pushed a new commit that removes these 2 lines altogether

thelostone-mc (approved):

PR #3589, opened 2024-08-07
chore: update chain data

gnomadic:
Fixes: #issue

## Checklist

This PR:

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -19,7 +19,7 @@
     "@allo-team/allo-v2-sdk": "^1.0.78",
     "@ethersproject/abstract-signer": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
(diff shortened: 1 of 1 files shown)


0xKurt:
closed in favor of #3586 

also you need to run `pnpm install` to update the lock file as well

0xKurt:
closed in favor of https://github.com/gitcoinco/grants-stack/pull/3586

also you need to run pnpm install to update the lock file as well

gnomadic:
> closed in favor of #3586
> 
> also you need to run `pnpm install` to update the lock file as well

awesome - yeah I was figuring out how to pass the push hook - it was failing locally when I committed the lockfile, which made me realize my pnpm version is way out of date.  Cleaned all that up, and let's move forward with the original PR!

Issue #3458, opened 2024-05-21
migrate OSO impact metrics to v1


ccerv1:
**Overview**
OSO is releasing a stable v1 of its API. For your queries, functionality will be identical to what you had previously, but the schema is a little different.

You can view the new models [here](https://cloud.hasura.io/public/graphiql?endpoint=https://opensource-observer.hasura.app/v1/graphql).

**Details**

In order to replicate the same queries you implemented [here](https://github.com/gitcoinco/grants-stack/pull/3344), the relevant models are:

```
artifacts_by_project_v1: {
    project_id
  }
```
and

```
code_metrics_by_project_v1 {
    contributor_count
    first_commit_date
  }
```

We are also pushing a stable `events_monthly_to_project_v1` model 🔜  (hopefully by the time you read this).

MasterHW:
hey carl, working on this now.
any ideas why this is timing out?
![image](https://github.com/user-attachments/assets/1dcd69e2-592e-4dc1-8de5-62589a39cf53)

ccerv1:
@MasterHW yes...
our `artifacts_by_project` table has exploded in size recently, so we had to change some things on our backend to improve performance. we are still doing some performance testing before sharing more broadly.

when are you looking to use the API again? tagging @ryscheng for visibility

MasterHW:
I've pivoted to use `projects_v1`, as it seems much lighter weight for what we use it for (fetching a project_id given a project/github name).

MasterHW:
Open to suggestions, do you think this is the most robust approach? https://github.com/gitcoinco/grants-stack/blob/fe4e8ea774f432207ddf0257ef45c2e6bcea8d31/packages/grant-explorer/src/features/api/oso.ts#L93

ryscheng:
ILIKE is more like string matching, so I worry that it may not be 100% robust under all edge cases
https://hasura.io/blog/full-text-search-with-hasura-graphql-api-postgres

Using artifactsbyproject is probably the most robust, but we need to help you make it more performant.
We have a beta API that is worth trying
https://www.opensource.observer/graphql

It is currently set to idle and wake on request, which we can turn off if you find this better.
We are also trying to tune the indices of artifactsbyproject to be faster.

We should add an index by artifact_source, artifact_namespace, artifact_name here
https://github.com/opensource-observer/oso/blob/main/warehouse/dbt/models/marts/directory/artifacts_by_project_v1.sql

PR #3596, opened 2024-08-09
Fix OpenSource Observer queries

MasterHW:
Fixes: #3458 

## Description

updates explorer project page's codebase impact stats to use OSO's latest models (was broken previously)
image

## Checklist

This PR:


MasterHW:
ready for review with current query. open to further refining it, but presently non-trivial.

thelostone-mc (approved):

PR #3637, opened 2024-09-11
fix: ensure all columns are present in csv

thelostone-mc:
The plugin we use to stringy decides what the columns are by the first entry. If user skips a question -> we don't have that column and the data is not exported.
This PR just generates a set of all questions AKA columns and set that to ensure all questions/answers are accounted for


0xKurt (approved):

0xKurt:
@thelostone-mc tests are failing

PR #3672, opened 2024-10-06
update OSO endpoint and queries

MasterHW:
Fixes: https://github.com/gitcoinco/grants-stack/issues/3669

## Description

https://linear.app/gitcoin/issue/PAR-438/update-oso-endpoint-and-queries
With current OSO models it is not straightforward to retrieve code velocity data. Updated that field to instead reflect OSO's `activeDeveloperCount6Months` field for a sense of active developers.

## Checklist

This PR:

### DIFF
--- a/packages/grant-explorer/src/features/api/oso.ts
+++ b/packages/grant-explorer/src/features/api/oso.ts
@@ -11,6 +11,7 @@ const graphQLClient = new GraphQLClient(osoUrl, {
   },
 });
 let hasFetched = false;
+let fetchedProject = "";
 
 export interface IOSOStats {
   oso_codeMetricsByProjectV1: {
@@ -31,6 +32,7 @@ export function useOSO(projectGithub?: string) {
   const [stats, setStats] = useState<IOSOStats | null>(null);
 
   const getStatsFor = async (projectRegistryGithub: string) => {
+    fetchedProject = projectRegistryGithub;
     if (osoApiKey === "")
       throw new Error("OpenSourceObserver API key not set.");
     const queryVars = {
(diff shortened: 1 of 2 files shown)


MasterHW:
known issue: when browsing through applications in explorer, Impact Stats will not populate the page after visiting a project with no stats (even if the application you're looking at does have them). page refresh fixes this. caching/cookie issue?

MasterHW:
> known issue: when browsing through applications in explorer, Impact Stats will not populate the page after visiting a project with no stats (even if the application you're looking at does have them). page refresh fixes this. caching/cookie issue?

believe this is fixed. improved logic determining when stats are queried to account for circumstances where stats have already been queried, but not for the currently viewed project.

packages/grant-explorer/src/features/api/oso.ts
```diff
@@ -4,147 +4,70 @@ import { Hex } from "viem";
...
-      const idData: IOSOId = await graphQLClient.request<IOSOId>(queryId);
+      const items: IOSOStats = await graphQLClient.request<IOSOStats>(
+        queryStats,
+        queryVars
+      );
 
-      if (!Array.isArray(idData.projects_v1)) {
-        setStats(emptyReturn);
-        return;
+      if (!Array.isArray(items.oso_codeMetricsByProjectV1)) {
```
veganbeef, line 60: You could combine this if statement with the else condition on line 70 with something like:
```
if (!items.oso_codeMetricsByProjectV1?.length) {
  throw new Error("no stats returned");
}
```
and then the `const parsedItems` block can be pulled out of that if condition and you don't need to duplicate the error.

That more closely follows the "return early" pattern, which makes things easier to read IMO.

veganbeef (approved):
LGTM, just one style suggestion

packages/grant-explorer/src/features/api/oso.ts
MasterHW: cheers, will incorporate!

gnomadic (approved):

PR #3675, opened 2024-10-07
WIP: Karma GAP SDK integration

veganbeef:
Fixes: [PAR-434](https://linear.app/gitcoin/issue/PAR-434/spike-out-hacky-integration-with-karma-gap)

## Description

This PR adds a new package to host an expanded version of the `useGap` hook from the `grants-explorer` package, with the intention of using the `karma-gap-sdk` package to integrate directly with Karma GAP.

_Note: There is a dependency mismatch between `karma-gap-sdk`, which requires `ethers` v6, and the rest of the grants-stack packages, which require `ethers` v5. It might be best to prove out this functionality in a standalone app._

## Checklist

This PR:


thelostone-mc:
Closing this PR as the POC is complete. Will keep branch alive

PR #3689, opened 2024-10-15
fix farcaster links

veganbeef:
Fixes: #issue

## Description

This PR updates the farcaster share URLs to properly support embedded links and mentions per the [warpcast docs](https://docs.farcaster.xyz/reference/warpcast/cast-composer-intents).

## Checklist

This PR:


nijoe1 (approved):
LGTM!

PR #3690, opened 2024-10-16
hide attestation

thelostone-mc:
- hide attest behind debug flag.
- this should be reverted for releasing attestation feature to public


hussedev (approved):
LGTM!

PR #3706, opened 2024-10-24
fix price fetching bug

0xKurt:
fixes PAR-566

We have made the following updates:

1. **Checkout**: 
   - If price fetching fails for any reason, users will still be able to proceed with the checkout process.

_2. **CUSD Token Update**:
   - The Redstone token ID for CUSD has been updated in the **chain data package**.

3. **USDGLO Replacement**:
   - Since USDGLO is not available on Redstone and accurate pricing is not critical in this context, we have replaced the Redstone token ID for USDGLO with CUSD. This ensures a similar approximation in price._

**Action Item**:
- Discuss the handling of tokens that are not available on Redstone. Begin by identifying where Redstone is currently used and then determine the importance of accurate prices for these services.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -19,7 +19,7 @@
     "@allo-team/allo-v2-sdk": "^1.0.79",
     "@ethersproject/abstract-signer": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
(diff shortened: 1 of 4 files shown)


hussedev (approved):
LGTM

PR #3728, opened 2024-11-21
Update DefaultLayout.tsx

JacobHomanics:
## Description

Removes the overlay gradient div and applies the gradiant to to the layout's children.

## Checklist

This PR:

- [X] If adding/updating a feature, it adds/updates its test script on Notion.


MasterHW:
like the new (better!) gradient, but i noticed when magnification is set to <100% on certain screens there are side boxes that don't exist on the live site. hopefully quick fix?
image

JacobHomanics:
> like the new (better!) gradient, but i noticed when magnification is set to <100% on certain screens there are side boxes that don't exist on the live site. hopefully quick fix? image

Should be fixed now! Let me know if it still doesn't fit the requirement.

MasterHW (approved):
lgmt :) thanks for tweaking the box behavior

Issue #3717, opened 2024-10-29
Edge case - grant approved then unapproved - display some info to the visitor? 


marsrobertson:
This can be seen as feature / bug / works as expected / no further action / won't fix. Reporting for visibility, especially that I'm _(most likely pretty sure)_ one of the first people to notice.

Steps to reproduce:
1. grant application, approved
2. me posting links on [my Twitter](https://x.com/marsXRobertson/status/1849596109368262961), direct link to the grant: https://explorer.gitcoin.co/#/round/42220/16/74
3. unapproved

<img width="1018" alt="image" src="https://github.com/user-attachments/assets/26504383-ecbd-4490-8f2b-b3e9de114bdd">

You can try any other application ID, such as 42069 https://explorer.gitcoin.co/#/round/42220/16/42069

### Option 1 _(minimal info)_

Brief generic message.

> The grant does not exist, you can visit Gitcoin homepage / grants / round / explorer

It is better than pulsating placeholder that gives an impression of loading.

### Option 2 _(more descriptive)_

Display the grant in the original form with some label: unapproved / rejected.

----

I think option 1 is simpler and faster, therefore better, on top of that it covers other use cases, such as someone iterating through grant IDs.

PR #3731, opened 2024-12-04
[Bugfix]  Better handling of unapproved grants in Grants Explorer

JacobHomanics:
Fixes: #3717 

## Description

The previous implementation was stuck in a loading state when a grant was unapproved leading to confusion. Now, it will properly display a message to the user if the grant is in an unapproved state.

## Checklist

This PR:

- [X] If adding/updating a feature, it adds/updates its test script on Notion.


MasterHW (approved):
small but real QoL enhancement - lgtm!

PR #3773, opened 2025-02-26
feat: indexer v2 query migrations

0xkenj1:
Update queries for Indexer V2 compatibility

## Changes
- Modified existing queries to align with Indexer V2 schema and endpoints
- Updated query syntax to ensure compatibility with new indexing structure

## Testing
- Verified queries return expected results with Indexer V2 (manual)

### DIFF
--- a/packages/data-layer/src/queries.ts
+++ b/packages/data-layer/src/queries.ts
@@ -1,5 +1,7 @@
 import { gql } from "graphql-request";
 
+//Note: All the addresses used in query filters should be checksummed before being used
+
 /**
  * Manager: Get all the programs that a user is a part of
  * @param $address - The address of the user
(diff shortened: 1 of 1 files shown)


packages/data-layer/src/queries.ts
```diff
@@ -582,19 +590,20 @@ export const getProjectsAndRolesByAddress = gql`
...
-      applications(first: 1000) {
+      applications(limit: 100) {
         id
         metadata
       }
     }
   }
 `;
 
+//NOT SUPPORTED ON THE CURRENT INDEXER
```
thelostone-mc, line 601: ^ what's the best way to do this instead ?

packages/data-layer/src/queries.ts
```diff
@@ -604,13 +613,14 @@ export const getBlockNumberQuery = gql`
   }
 `;
 
+//TODO: migrate code in which this query is called
```
thelostone-mc, line 616: Ack

packages/data-layer/src/queries.ts
```diff
@@ -136,6 +143,7 @@ export const getProjectAnchorByIdAndChainId = gql`
  *
  * @returns The project
  */
+//FIXME: Deprecated on indexer v2 ?
```
thelostone-mc, line 146: We'd have to fetch all projects and do the filter on the frontend right ?

packages/data-layer/src/queries.ts
0xKurt: can we make use of the events_registry here?

0xkenj1: i will add it to expose it on the hasura api

PR #3796, opened 2025-03-26
add recipient address copy to manager

MasterHW:
Fixes: #issue

## Checklist

This PR:


packages/grant-explorer/tsconfig.json
```diff
@@ -19,5 +19,5 @@
     "types": ["vitest/globals", "@testing-library/jest-dom"],
     "incremental": true
   },
-  "include": ["src"]
+  "include": ["src", "../common/src/components/CopyToClipboard.tsx", "../common/src/components/truncate.ts"]
```
gnomadic, line 22: this doesn't look right?  shouldn't we include the common package instead of the source files?

gnomadic (approved):

PR #3800, opened 2025-03-27
chore: added staking countdown functionality

hussedev:
- renamed useIsDonationPeriod to useDonationPeriod, now the hook also returns the remaining time to start/end donations
- StakingBanner now has a prop children to receive the button or the countdown label


nijoe1 (approved):
LGTM!

PR #3820, opened 2025-04-07
feat: project activities

amaurymagalhaes:
## Description

Added new section to Impact Measurement tab on project page of explorer. 
It have brings the updates/activities along with deliverables block and indicators/metrics with datapoints.

## Examples
As some projects doesn't have it, there is a few projects that contains project activities.
`<localUrl>/#/round/42161/29/87`
`<localUrl>/#/round/42161/29/2`

## Checklist

This PR:


MasterHW (approved):
lgtm!

